Issue #22, opened 2017-09-07
Service type LoadBalancer lacks balancer IP


odacremolbap:
Users at kubernetes will need the external loadbalancer host for any `Service` type `LoadBalancer`
The cloud provider doesn't populate the IP since the load balancer hasn't  been provisioned yet

We should probably wait for the balancer to be active before writing the service's IP

odacremolbap:
Fixed by #23

PR #23, opened 2017-09-07
Wait for load balancer to be active to retrieve its IP

odacremolbap:
Waits for the balancer to be `active` at Digital Ocean.
Times out if the balancer is not active within a minute

**Please**, review the timeout and timespan values (60, 5 respectively), since they are hardcoded consts.

Related to https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/22

### DIFF
--- a/do/loadbalancers.go
+++ b/do/loadbalancers.go
@@ -17,10 +17,12 @@ limitations under the License.
 package do
 
 import (
+	goctx "context"
 	"errors"
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	"k8s.io/kubernetes/pkg/api/v1"
 	"k8s.io/kubernetes/pkg/cloudprovider"
@@ -54,19 +56,34 @@ const (
 	// should use. Options are round_robin and least_connections. Defaults
 	// to round_robin
 	annDOAlgorithm = "service.beta.kubernetes.io/do-loadbalancer-algorithm"
+
+	// defaultActiveTimeout is the number of seconds to wait for a load balancer to
+	// reach the active state
+	defaultActiveTimeout = 90
+
+	// defaultActiveCheckTick is the number of seconds between load balancer
+	// status checks when waiting for activation
+	defaultActiveCheckTick = 5
+
+	// statuses for Digital Ocean load balancer
+	lbStatusNew     = "new"
+	lbStatusActive  = "active"
+	lbStatusErrored = "errored"
 )
 
 var lbNotFound = errors.New("loadbalancer not found")
 
 // loadbalancers implements cloudprovider.Loadbalancer
 type loadbalancers struct {
-	client *godo.Client
-	region string
+	client            *godo.Client
+	region            string
+	lbActiveTimeout   int
+	lbActiveCheckTick int
 }
 
 // newLoadbalancers returns a type loadbalancer, implementing cloudprovider.Loadbalancer
 func newLoadbalancers(client *godo.Client, region string) cloudprovider.LoadBalancer {
-	return &loadbalancers{client, region}
+	return &loadbalancers{client, region, defaultActiveTimeout, defaultActiveCheckTick}
 }
 
 // GetLoadBalancer specifies whether the loadbalancer exists based on the provided service
@@ -81,6 +98,13 @@ func (l *loadbalancers) GetLoadBalancer(clusterName string, service *v1.Service)
 		}
 
 		return nil, false, err
+	}
+
+	if lb.Status != lbStatusActive {
+		lb, err = l.waitActive(lb.ID)
+		if err != nil {
+			return nil, true, fmt.Errorf("error waiting for load balancer to be active %v", err)
+		}
 	}
 
 	return &v1.LoadBalancerStatus{
@@ -108,6 +132,11 @@ func (l *loadbalancers) EnsureLoadBalancer(clusterName string, service *v1.Servi
 		}
 
 		lb, _, err := l.client.LoadBalancers.Create(context.TODO(), lbRequest)
+		if err != nil {
+			return nil, err
+		}
+
+		lb, err = l.waitActive(lb.ID)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


odacremolbap:
@andrewsykim ^^

andrewsykim:
Thanks for the PR! I'll review this some time today or tomorrow :)

do/loadbalancers.go
```diff
@@ -54,6 +55,19 @@ const (
 	// should use. Options are round_robin and least_connections. Defaults
 	// to round_robin
 	annDOAlgorithm = "service.beta.kubernetes.io/do-loadbalancer-algorithm"
+
+	// lbActiveTimeout is the number of seconds to wait for a load balancer to
+	// reach the active state
+	lbActiveTimeout = 60
```
andrewsykim: 60 seconds seems very reasonable, but since these are hard coded constants I don't think it hurts to extend to 90.

do/loadbalancers.go
```diff
@@ -112,6 +133,11 @@ func (l *loadbalancers) EnsureLoadBalancer(clusterName string, service *v1.Servi
 			return nil, err
 		}
 
+		_, err = l.waitActive(lb.ID, lbActiveTimeout, lbActiveCheckTimeSpan)
```
andrewsykim: should this be `lb, err`?

do/loadbalancers.go
```diff
@@ -251,6 +277,30 @@ func (l *loadbalancers) buildLoadBalancerRequest(service *v1.Service, nodes []*v
...
+			return nil, err
+		}
+
+		if lb.Status == lbStatusActive {
+			return lb, nil
+		}
+		if lb.Status == lbStatusErrored {
+			return nil, fmt.Errorf("error creating balancer %q at Digital Ocean", lbID)
+		}
+		if end < time.Since(start) {
```
andrewsykim: Can we change this code to use ticker/contexts, I find its easier to understand. Something like:

```go
ctx, _ := context.WithTimeout(context.TODO(), time.Second*time.Duration(timeout))
for {
    select {
        case <-ctx.Done():
              return errors.New(...)
        case <- ticker.C:
              // check status here
    }
}
```

What do you think?

do/loadbalancers.go
```diff
@@ -251,6 +277,30 @@ func (l *loadbalancers) buildLoadBalancerRequest(service *v1.Service, nodes []*v
 	}, nil
 }
 
+func (l *loadbalancers) waitActive(lbID string, timeout int, timespan int) (*godo.LoadBalancer, error) {
```
andrewsykim: If timeout and timespan are consts, no need to pass them in as func parameters. We can just call `lbActiveTimeout` from the function itself

do/loadbalancers.go
```diff
@@ -54,6 +55,19 @@ const (
...
 	// to round_robin
 	annDOAlgorithm = "service.beta.kubernetes.io/do-loadbalancer-algorithm"
+
+	// lbActiveTimeout is the number of seconds to wait for a load balancer to
+	// reach the active state
+	lbActiveTimeout = 60
+
+	// lbActiveCheckTimeSpan is the number of seconds between load balancer
+	// status checks when waiting for activation
+	lbActiveCheckTimeSpan = 5
```
andrewsykim: I think `lbActiveCheckTick` is a better name, what do you think?

do/loadbalancers.go
```diff
@@ -83,6 +97,13 @@ func (l *loadbalancers) GetLoadBalancer(clusterName string, service *v1.Service)
 		return nil, false, err
 	}
 
+	if lb.Status != lbStatusActive {
+		lb, err = l.waitActive(lb.ID, lbActiveTimeout, lbActiveCheckTimeSpan)
+		if err != nil {
+			return nil, true, err
```
andrewsykim: We should `fmt.Errorf("error waiting for LB to be active: %v", err)` this

do/loadbalancers.go
andrewsykim: If the params are there for testing, feel free to add it as internal fields to `loadbalancers` struct

do/loadbalancers.go
odacremolbap: doh!

do/loadbalancers.go
```diff
@@ -251,6 +280,31 @@ func (l *loadbalancers) buildLoadBalancerRequest(service *v1.Service, nodes []*v
...
+			lb, _, err := l.client.LoadBalancers.Get(context.TODO(), lbID)
+			if err != nil {
+				return nil, err
+			}
+
+			if lb.Status == lbStatusActive {
+				return lb, nil
+			}
+			if lb.Status == lbStatusErrored {
+				return nil, fmt.Errorf("error creating balancer %q at Digital Ocean", lbID)
```
andrewsykim: nit: `error creating DigitalOcean loadbalancer: %q`

andrewsykim:
Left a minor comment, squash and this should be good to go!

do/loadbalancers.go
```diff
@@ -17,10 +17,12 @@ limitations under the License.
 package do
 
 import (
+	goctx "context"
```
fatih: Is this a common abbreviation in kube land? Unless it's not, I suggest to use `context` and not to rename it.

do/loadbalancers.go
```diff
@@ -251,6 +280,31 @@ func (l *loadbalancers) buildLoadBalancerRequest(service *v1.Service, nodes []*v
 	}, nil
 }
 
+func (l *loadbalancers) waitActive(lbID string) (*godo.LoadBalancer, error) {
+
+	ctx, cancel := goctx.WithTimeout(goctx.TODO(), time.Second*time.Duration(l.lbActiveTimeout))
+	defer cancel()
+	ticker := time.NewTicker(time.Second * time.Duration(l.lbActiveCheckTick))
```
fatih: You should call `defer ticker.Stop()` otherwise it'll leak goroutines.

do/loadbalancers.go
```diff
@@ -251,6 +280,31 @@ func (l *loadbalancers) buildLoadBalancerRequest(service *v1.Service, nodes []*v
...
+	for {
+		select {
+		case <-ticker.C:
+			lb, _, err := l.client.LoadBalancers.Get(context.TODO(), lbID)
```
fatih: Can you pass the context from above?

do/loadbalancers.go
odacremolbap: That alias is there to distinguish go context from godo context.

https://github.com/StackPointCloud/digitalocean-cloud-controller-manager/blob/f56f1ba771fea00550b1653680dcf77f10398d5d/do/loadbalancers.go#L20-L31

(Yes, probably the alias is unfortunate)

fatih: Maybe we should rename that name to `doctx` (because that's how named it internally as well). Thoughts @andrewsykim ?

odacremolbap: sounds good. Waiting for Andrew to remove `goctx` alias and set `doctx` for `github.com/digitalocean/godo/context`

andrewsykim: yeah, doctx seems to make more sense as opposed to aliasing stdlib, @odacremolbap feel free to leave `goctx` for now and I'll address that in another PR :)

do/loadbalancers.go
```diff
@@ -285,10 +285,11 @@ func (l *loadbalancers) waitActive(lbID string) (*godo.LoadBalancer, error) {
 	ctx, cancel := goctx.WithTimeout(goctx.TODO(), time.Second*time.Duration(l.lbActiveTimeout))
 	defer cancel()
 	ticker := time.NewTicker(time.Second * time.Duration(l.lbActiveCheckTick))
+	defer ticker.Stop()
 	for {
 		select {
 		case <-ticker.C:
-			lb, _, err := l.client.LoadBalancers.Get(context.TODO(), lbID)
+			lb, _, err := l.client.LoadBalancers.Get(goctx.TODO(), lbID)
```
fatih: This is still wrong. We need to pass the `ctx` value from above so it's get propagated. The `TODO()` will do nothing in this case.

do/loadbalancers.go
odacremolbap: doh!

fatih:
lgtm. If you can squash all the commits that would be great.

do/loadbalancers.go
odacremolbap: ok

andrewsykim:
Thanks for the PR @odacremolbap, much appreicated!

PR #28, opened 2017-09-26
update vendor for kubernetes v1.8

andrewsykim:
Using release candidate tag for now. https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/27

PR #30, opened 2017-09-27
WIP - implement the k8s 1.8 Zones interface

### LINKED ISSUE #18: Implement GetZoneByProviderID and GetZoneByNodeName
As per https://github.com/kubernetes/kubernetes/issues/50926, DO CCM should implement those methods so we accurately represent the region of droplets instead of assuming droplets are in the same region as CCM.

bhcleek:
I still need to test this against a 1.8 cluster and rebase on top of #28, but would appreciate any early feedback.

Fixes #18

### DIFF
--- a/do/zones.go
+++ b/do/zones.go
@@ -17,30 +17,47 @@ limitations under the License.
 package do
 
 import (
+	"github.com/digitalocean/godo"
+	"github.com/digitalocean/godo/context"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kubernetes/pkg/cloudprovider"
 )
 
 type zones struct {
+	client *godo.Client
 	region string
 }
 
-func newZones(region string) cloudprovider.Zones {
-	return zones{region}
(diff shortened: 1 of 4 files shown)


do/zones.go
```diff
@@ -17,19 +17,47 @@ limitations under the License.
...
+// GetZone returns a cloudprovider.Zone. GetZone will only fill the Region
+// field of cloudprovider.Zone for DO.
 func (z zones) GetZone() (cloudprovider.Zone, error) {
 	return cloudprovider.Zone{Region: z.region}, nil
 }
+
+// GetZoneByProviderID returns the Zone containing the current zone and
+// locality region of the node specified by providerId. GetZoneByProviderID
+// will only fill the Region field of cloudprovider.Zone for DO.
+func (z zones) GetZoneByProviderID(providerID string) (cloudprovider.Zone, error) {
```
andrewsykim: This is actually going to depend on some of the changes in https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/36

Issue #32, opened 2017-09-27
Cloud controller manager should be critical pod and have resource requests


andrewsykim:
CCM is pretty important to cluster operations, we should run it as a critical pod and be given resource requirements. 

cc @odacremolbap

PR #33, opened 2017-09-27
CCM should be critical pod and have resource requests

andrewsykim:
Fixes https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/32. Also adds toleration for master node taint.

@odacremolbap as you mentioned, 100m CPU and 50Mi memory seems pretty reasonable for now. It's also a resource request and not a limit for now so it can still burst if needed. 

In general, I think we should avoid making changes to the files in releases as they should be locked down once a release goes in. But this change seems pretty harmless since we're not changing any of the CCM flags that might change its behavior.

### DIFF
--- a/releases/v0.1.0.yml
+++ b/releases/v0.1.0.yml
@@ -11,6 +11,8 @@ spec:
     metadata:
       labels:
         app: digitalocean-cloud-controller-manager
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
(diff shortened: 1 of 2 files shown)


andrewsykim:
Also in general, I think we should avoid making changes to the files in `releases` as they should be locked down once a release goes in. But this change seems pretty harmless (helpful if anything) and we are still in alpha ;)

odacremolbap:
LGTM

Issue #15, opened 2017-08-16
Use correct format for Node.Spec.ProviderID


andrewsykim:
Kubernetes uses the format `providerName://node-id` for `Node.Spec.ProviderID`.  Functions like [NodeAddressesByProviderID](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/do/droplets.go#L55) should expect the providerID to be in this format. This isn't an issue for the time being since cloud controller manager doesn't even set `Node.Spec.ProviderID` but since it will in the future we should plan accordingly (see issue below). 

See this for the expected format: https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/cloud.go#L73-L84

Related: 
https://github.com/kubernetes/kubernetes/issues/49836

PR #36, opened 2017-09-27
Handle correct providerID format

andrewsykim:
Addresses https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/15. 

In v1.7.x this wasn't important since CCM never set providerID in the first place. In v1.8, it will so we should handle the expected format. The expected format for node's providerID is `digitalocean://droplet-id`, example:

```yaml
apiVersion: v1
kind: Node
metadata:
  annotations:
    node.alpha.kubernetes.io/ttl: "0"
    volumes.kubernetes.io/controller-managed-attach-detach: "true"
  creationTimestamp: 2017-09-27T14:30:44Z
  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/instance-type: 512mb
    beta.kubernetes.io/os: linux
    kubernetes.io/hostname: k8s-worker-01
    kubernetes.io/role: node
  name: k8s-worker-01
  resourceVersion: "7123"
  selfLink: /api/v1/nodes/k8s-worker-01
  uid: 720492a5-a390-11e7-8009-6ea0aaddb09e
spec:
  externalID: k8s-worker-01
  providerID: digitalocean://64011171
```

### DIFF
--- a/do/droplets.go
+++ b/do/droplets.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -54,8 +55,12 @@ func (i *instances) NodeAddresses(nodeName types.NodeName) ([]v1.NodeAddress, er
 // NodeAddressesByProviderID returns all the valid addresses of the specified
 // node by providerId. For DO this is the public/private ipv4 addresses for now.
 func (i *instances) NodeAddressesByProviderID(providerId string) ([]v1.NodeAddress, error) {
-	// we can technically get all the required data from metadata service
-	droplet, err := dropletByID(context.TODO(), i.client, providerId)
+	id, err := dropletIDFromProviderID(providerId)
+	if err != nil {
+		return nil, err
+	}
+
+	droplet, err := dropletByID(context.TODO(), i.client, id)
 	if err != nil {
 		return nil, err
 	}
@@ -111,7 +116,12 @@ func (i *instances) InstanceType(name types.NodeName) (string, error) {
 
 // InstanceTypeByProviderID returns the type of the specified instance.
 func (i *instances) InstanceTypeByProviderID(providerId string) (string, error) {
-	droplet, err := dropletByID(context.TODO(), i.client, providerId)
+	id, err := dropletIDFromProviderID(providerId)
+	if err != nil {
+		return "", err
+	}
+
+	droplet, err := dropletByID(context.TODO(), i.client, id)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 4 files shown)


do/droplets.go
```diff
@@ -174,3 +184,24 @@ func (i *instances) dropletByName(ctx context.Context, nodeName types.NodeName)
...
+
+// dropletIDFromProviderID returns a droplet's ID extracted from the node's
+// providerID spec. The providerID spec should be retrievable from the Kubernetes
+// node object. The expected format is: digitalocean://droplet-id
+func dropletIDFromProviderID(providerID string) (string, error) {
+	if providerID == "" {
+		return "", errors.New("providerID cannot be empty string")
+	}
+
+	split := strings.Split(providerID, "/")
```
bhcleek: `strings.SplitN(providerID, "/", 3)`?

do/droplets.go
```diff
@@ -174,3 +184,24 @@ func (i *instances) dropletByName(ctx context.Context, nodeName types.NodeName)
...
+	if len(split) != 3 {
+		return "", fmt.Errorf("unexpected providerID format: %s, format should be: digitalocean://12345", providerID)
+	}
+
+	// since split[0] is actually "digitalocean:"
+	if strings.TrimSuffix(split[0], ":") != providerName {
```
bhcleek: Maybe just `if split[0] != providerName + ":" {`?

do/droplets.go
```diff
@@ -174,3 +184,24 @@ func (i *instances) dropletByName(ctx context.Context, nodeName types.NodeName)
...
+		return "", fmt.Errorf("provider name from providerID should be digitalocean: %s", providerID)
+	}
```
bhcleek: `return "", fmt.Errorf("provider name from providerID should be %s://%s", providerName, providerID)`?

do/droplets.go
```diff
@@ -174,3 +184,24 @@ func (i *instances) dropletByName(ctx context.Context, nodeName types.NodeName)
 
 	return nil, cloudprovider.InstanceNotFound
 }
+
+// dropletIDFromProviderID returns a droplet's ID extracted from the node's
+// providerID spec. The providerID spec should be retrievable from the Kubernetes
+// node object. The expected format is: digitalocean://droplet-id
+func dropletIDFromProviderID(providerID string) (string, error) {
```
bhcleek: There are some comments below that are just nitpicks. Take them or not.

Also consider: is there any reason to split on the slashes at all? Why not just 
```
func dropletIDFromProviderID(providerID string) (string, error) {
    const scheme := providerID + "://"

    if !strings.HasPrefix(providerID, scheme) != 0 {
        return "", fmt.Errorf("unexpected providerID format: %s, format should be: digitalocean://12345", providerID)
    }
    return strings.TrimPrefix(providerID, scheme), nil
}
```

andrewsykim: Did you mean 
```
const scheme := providerName + "://"
```
?

do/droplets.go
andrewsykim: `providerID` is already in the format `providerName://id` so it would be `digitalocean://digitalocean://droplet-id`. I think the confusion here is that `providerID` is the droplet ID which it is not, it's a separate concept from Kubernetes land :P

do/droplets.go
bhcleek: yes

do/droplets.go
andrewsykim: same comment as above

do/droplets.go
bhcleek: That's a great reason for using `splitN` or `HasPrefix`, though. Right now on master, `instances.NodeAddressesByProviderID` will look for the `providerID` given. All this PR is about is making sure that `digitalocean://` is a prefix for that. The PR, though, changes behavio where previously `instances.dropletById` would have been given `//droplet-id`.

Unless we're prepared to validate other rules about the `droplet-id` value, not allowing leading `/`s seems out of scope.

bhcleek: Having said all that, this is all a nitpick; I don't intend it to be a blocker.

PR #37, opened 2017-10-02
implement the k8s 1.8 Zones interface

### LINKED ISSUE #18: Implement GetZoneByProviderID and GetZoneByNodeName

bhcleek:
Fixes #18


andrewsykim:
lgtm!!! Thanks @bhcleek!

PR #38, opened 2017-10-02
Check for headers on CI

### LINKED ISSUE #29: Check for Licensing Headers in CI
When we run CI on TravisCI we should check for licensing headers on all go files not in `/vendor`.

xmudrii:
Closes #29 

/cc @andrewsykim @fatih


ci/headers.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# Copyright 2017 DigitalOcean
```
andrewsykim: I think the biggest challenge with the header checks is that the format varies depending on the source file. This check will work for Go, but it will change for bash scripts, Dockerfiles, etc. I wouldn't be opposed to having a list of known formats and just using that. What do you think?

xmudrii: Do you want me to create bash scripts for several known such as Bash and Dockerfiles or you have better idea?

xmudrii:
‎Thanks, I left an comment there :)

ci/headers.sh
andrewsykim: I think several known is probably the best for now.

xmudrii:
‎‎Sounds good to me, I'll update the PR once I found some free time, I don't have computer currently

xmudrii:
@andrewsykim I've updated the PR. I'm not sure will this do the job, but it seems to work.
Maybe, we could make only one file and build headers string depending on the file type, but that could be harder to maintain.

ci/gofmt.sh
```diff
@@ -1,12 +1,12 @@
 #!/bin/bash
 
-# Copyright 2016 The Kubernetes Authors.
+# Copyright 2017 DigitalOcean
```
andrewsykim: doh... LOL guess where I got this from ;)

ci/headers-bash.sh
```diff
@@ -0,0 +1,46 @@
...
+EOF
+
+FILES=$(find . -name "*.sh" -not -path "./vendor/*")
+EXIT=0
+
+for FILE in $FILES; do
+        HEADERS=$(head -n 15 $FILE)
+        if [ "$HEADERS" != "$LICENSE" ]; then
+                echo "license headers not found in $FILE"
+		EXIT=1
```
andrewsykim: Any reason you went with using an `EXIT` var? I think it's simpler if we just `exit 1` here and get rid of the EXIT var cause we would `exit 0` by default when the script ends anyways. What do you think?

andrewsykim:
Left a small nit

ci/headers-bash.sh
xmudrii: @andrewsykim If we would just use `exit 1`, once the file without headers is found, it would exit the script. That means we would get only one name of the file without headers, for each run, even if there are more (e.g. if there are 10 files without headers, you would have to run script 10 times).

Using `EXIT` var, allows script to run through all files, but yet to finish with error code.

andrewsykim: hmm, good point!

andrewsykim:
LGTM, thanks for the PR @xmudrii!!!

PR #39, opened 2017-10-04
use the region slug instead of region name

bhcleek:
Use the region slug instead of the region name to communicate the region
of a given node.


andrewsykim:
LGTM!

Issue #21, opened 2017-08-27
Implement InstanceExistsByProviderID()


andrewsykim:
As per https://github.com/kubernetes/kubernetes/issues/51406, we now have `InstanceExistsByProviderID()` in the Instances interface.

PR #40, opened 2017-10-04
Implement InstanceExistsByProviderID

andrewsykim:
For https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/21.

### DIFF
--- a/do/droplets.go
+++ b/do/droplets.go
@@ -144,6 +144,25 @@ func (i *instances) CurrentNodeName(hostname string) (types.NodeName, error) {
 // InstanceExistsByProviderID returns true if the instance for the given provider id still is running.
 // If false is returned with no error, the instance will be immediately deleted by the cloud controller manager.
 func (i *instances) InstanceExistsByProviderID(providerID string) (bool, error) {
+	id, err := dropletIDFromProviderID(providerID)
+	if err != nil {
+		return false, err
+	}
+
+	_, err = dropletByID(context.TODO(), i.client, id)
+	if err == nil {
+		return true, nil
+	}
+
+	godoErr, ok := err.(*godo.ErrorResponse)
+	if !ok {
+		return false, fmt.Errorf("unexpected error type from godo: %T, msg: %v", err, err)
+	}
+
+	if godoErr.Response.StatusCode != http.StatusNotFound {
+		return false, fmt.Errorf("error checking if instance exists: %v", err)
+	}
+
 	return false, nil
 }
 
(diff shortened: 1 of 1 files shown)


do/droplets.go
```diff
@@ -144,6 +144,25 @@ func (i *instances) CurrentNodeName(hostname string) (types.NodeName, error) {
...
 // If false is returned with no error, the instance will be immediately deleted by the cloud controller manager.
 func (i *instances) InstanceExistsByProviderID(providerID string) (bool, error) {
+	id, err := dropletIDFromProviderID(providerID)
+	if err != nil {
+		return false, err
+	}
+
+	_, err = dropletByID(context.TODO(), i.client, id)
+	if err == nil {
+		return true, nil
```
bhcleek: The early return here is surprising. I get what it's doing, but as a reader, I'd expect to something more like

```
_, err := dropletByID(context.TODO(), i.client, id)
if err != nil {
    godoErr, ok := err.(*godo.ErrorResponse)
    if !ok {
        return false, fmt.Errorf("unexpected error type from godo: %T; %v", err, err)
    }

    if godoErr.Response. StatusCode != http.StatusNotFound {
        return false, fmt.Errorf("error checking if instance exists: %v", err)
    }
    return false, nil
}

return true, nil
```

do/droplets.go
```diff
@@ -144,6 +144,25 @@ func (i *instances) CurrentNodeName(hostname string) (types.NodeName, error) {
...
+	}
+
+	godoErr, ok := err.(*godo.ErrorResponse)
+	if !ok {
+		return false, fmt.Errorf("unexpected error type from godo: %T, %v", godoErr, err)
```
bhcleek: `godoErr` will be nil in this case. Did you mean `err`?

do/droplets.go
```diff
@@ -154,15 +173,11 @@ func dropletByID(ctx context.Context, client *godo.Client, id string) (*godo.Dro
...
-	droplet, resp, err := client.Droplets.Get(ctx, intId)
+	droplet, _, err := client.Droplets.Get(ctx, intId)
 	if err != nil {
 		return nil, err
 	}
 
-	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("DO API returned non-200 status code: %d", resp.StatusCode)
-	}
-
```
bhcleek: Removing this code path may cause unexpected side-effects on other code that was expecting errors when the droplet wasn't found. Have you examined those closely to make sure other callers won't panic when `droplet` is nil?

do/droplets.go
andrewsykim: hmm, interesting. I always tend try to reduce the amount of code thats indented for readability so it made more sense here to return early. Though it can be misleading cause we're so used to seeing `err != nil` 😛

do/droplets.go
andrewsykim: ahh yes, good catch

do/droplets.go
andrewsykim: yeah, godo returns nil for anything within 200 - 299 anyways and non-nil err otherwise, so it _should_ be the same behaviour. 

https://github.com/digitalocean/godo/blob/master/godo.go#L350-L352

andrewsykim: if anything, this might be a bug fix cause response may not be `http.StatusOK` but still 2XX

do/droplets.go
bhcleek: I like reducing indentation :-). And you're right, it is because we're used to seeing `err != nil`. Clearly, this is correct, and I don't feel strongly about changing it, but I did stumble over it when I first read it.. take that for what it's worth.

bhcleek:
LGTM.

PR #46, opened 2017-10-07
Add script to generate Kubernetes Secret for DigitalOcean

andrewsykim:
We create a secret containing a DO access token for cloud-controller-manager to consume. Though it's documented how to create the secret, figured it would also help to create a script to make this easier. 

@xmudrii @bhcleek  r?

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ To run digitalocean-cloud-controller-manager, you need a digitalocean access tok
 #### Script
 You can use the script [scripts/generate-secret.sh](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/scripts/generate-secret.sh) in this repo to create the Kubernetes Secret. Note that this will apply changes using your default `kubectl` context. For example, if your token is `abc123abc123abc123`, run the following to create the Kubernetes Secret.
 ```bash
-export DO_ACCESS_TOKEN=abc123abc123abc123
+export DIGITALOCEAN_ACCESS_TOKEN=abc123abc123abc123
 scripts/generate-secret.sh
 ```
 
--- a/README.md
+++ b/README.md
@@ -51,15 +51,23 @@ In the future, it may implement:
 ## Deployment
 
 ### Token
-To run digitalocean-cloud-controller-manager, you need a digitalocean access token. If you are already logged in, you can create one [here](https://cloud.digitalocean.com/settings/api/tokens). Ensure the token you create has both read and write access. Once you have an access token, you want to create a Kubernetes Secret.
+To run digitalocean-cloud-controller-manager, you need a digitalocean access token. If you are already logged in, you can create one [here](https://cloud.digitalocean.com/settings/api/tokens). Ensure the token you create has both read and write access. Once you have an access token, you want to create a Kubernetes Secret as a way for the cloud controller manager to access your token. You can do this with one of the following methods:
 
+#### Script
+You can use the script [scripts/generate-secret.sh](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/scripts/generate-secret.sh) in this repo to create the Kubernetes Secret. Note that this will apply changes using your default `kubectl` context. For example, if your token is `abc123abc123abc123`, run the following to create the Kubernetes Secret.
+```bash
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -51,15 +51,23 @@ In the future, it may implement:
...
+To run digitalocean-cloud-controller-manager, you need a digitalocean access token. If you are already logged in, you can create one [here](https://cloud.digitalocean.com/settings/api/tokens). Ensure the token you create has both read and write access. Once you have an access token, you want to create a Kubernetes Secret as a way for the cloud controller manager to access your token. You can do this with one of the following methods:
 
+#### Script
+You can use the script [scripts/generate-secret.sh](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/scripts/generate-secret.sh) in this repo to create the Kubernetes Secret. Note that this will apply changes using your default `kubectl` context. For example, if your token is `abc123abc123abc123`, run the following to create the Kubernetes Secret.
+```bash
+export DO_ACCESS_TOKEN=abc123abc123abc123
```
xmudrii: I'm not sure about this, but I think `DIGITALOCEAN_ACCESS_TOKEN` var is the most common for storing API token. For example, doctl, kubicorn and few more projects I know uses it.

scripts/generate-secret.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if [[ -z $DO_ACCESS_TOKEN ]]; then
+  echo "DO_ACCESS_TOKEN is empty, this must be set to your DigitalOcean access token..."
+  exit 1
+fi
+
+ENCODED_ACCESS_TOKEN=$(echo -n "$DO_ACCESS_TOKEN" | base64)
+
+GENERATED_SECRET=./generated-secret.yml
+trap "{ rm $GENERATED_SECRET; }" EXIT
```
xmudrii: Maybe we should leave the file, so user can copy it or choose the another cluster (or maybe even better ask user does it want to keep the file).

README.md
```diff
@@ -51,15 +51,23 @@ In the future, it may implement:
 ## Deployment
 
 ### Token
-To run digitalocean-cloud-controller-manager, you need a digitalocean access token. If you are already logged in, you can create one [here](https://cloud.digitalocean.com/settings/api/tokens). Ensure the token you create has both read and write access. Once you have an access token, you want to create a Kubernetes Secret.
+To run digitalocean-cloud-controller-manager, you need a digitalocean access token. If you are already logged in, you can create one [here](https://cloud.digitalocean.com/settings/api/tokens). Ensure the token you create has both read and write access. Once you have an access token, you want to create a Kubernetes Secret as a way for the cloud controller manager to access your token. You can do this with one of the following methods:
 
+#### Script
+You can use the script [scripts/generate-secret.sh](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/scripts/generate-secret.sh) in this repo to create the Kubernetes Secret. Note that this will apply changes using your default `kubectl` context. For example, if your token is `abc123abc123abc123`, run the following to create the Kubernetes Secret.
```
xmudrii: Maybe change the `Note that` part to something mentioning you have the option to apply it, but it's not required.

xmudrii:
@andrewsomething left some stuff for discussion in the review :)

scripts/generate-secret.sh
andrewsykim: The secret is outputted as part of the script, so they can just copy the stdout from the script. Thoughts?

README.md
andrewsykim: Sounds good to me

scripts/generate-secret.sh
xmudrii: Oh, my bad -- I thought it's returning only the secret, not the whole file. In that case, it's okay and should stay as it is.

README.md
andrewsykim: The above section mentions you can create the secret manually if you want :)

andrewsykim:
@xmudrii thanks for the quick response, addressed your comments

README.md
xmudrii: Sounds good to me. :)

PR #48, opened 2017-10-08
reduce risk of a contributor commiting their token

bhcleek:
* Rename releases/token.yml to releases/secret.yml.tmpl
* Add a .gitignore entry for releases/secret.yml
* Edit README.md instructions to direct users to copy
  releases/secret.yml.tmpl to releases/secret.yml before replacing the
  placeholder.

Edit references to DigitalOcean in README.md to be capitalized
correctly.

Edit the placeholder value; remove the the wrapping angle brackets so
that users don't mistakenly think the angle brackets are necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,9 +61,7 @@ scripts/generate-secret.sh
 ```
 
 #### Manually
-Copy [releases/secret.yml.tmpl](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/releases/secret.yml#L10) to releases/secret.yml and replace the placeholder in the copy with your base64 encoded token.
-
-Copy releases/secret.yml.tmpl to releases/secret.yml:
+Copy [releases/secret.yml.tmpl](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/releases/secret.yml.tmpl) to releases/secret.yml:
 ```bash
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -61,12 +61,19 @@ scripts/generate-secret.sh
 ```
 
 #### Manually
+Copy [releases/secret.yml.tmpl](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/releases/secret.yml#L10) to releases/secret.yml and replace the placeholder in the copy with your base64 encoded token.
```
andrewsykim: `s/secret.yml#L10/secret.yml.tmp#L10/` ?

andrewsykim: I think this part should be after the bit about base64 encoding the token since we refer to the base64 encoded token and it may be confusing to what that actually is for people.

README.md
```diff
@@ -61,12 +61,19 @@ scripts/generate-secret.sh
...
+
+Copy releases/secret.yml.tmpl to releases/secret.yml:
```
andrewsykim: This line is just repeating what's above?

bhcleek:
PTAL

andrewsykim:
LGTM, thanks @bhcleek :)

PR #49, opened 2017-10-08
docs: account for new lines in base64 encoded secrets

xmudrii:
This PR adds `-w 0` flag to the all base64 to prevent confusion as `base64` automatically adds new line after 72 characters.

This also fix the secret generation script, as it returns the following error if new line is present:
```
...
Create the following secret to your default kubectl context? [y/N]: y
Creating Kubernetes Secret for DigitalOcean
error: error converting YAML to JSON: yaml: line 9: could not find expected ':'
```

/cc @andrewsykim

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ cp releases/secret.yml.tmpl releases/secret.yml
 
 If your token is `abc123abc123abc123`, then you want to base64 encode your token like so:
 ```bash
-echo -n "abc123abc123abc123" | base64
+echo -n "abc123abc123abc123" | base64 | tr -d '\r\n'
 ```
 
 Replace the placeholder in the copy with your base64 encoded token. When you're done, the releases/secret.yml should look something like this:
--- a/releases/secret.yml.tmpl
+++ b/releases/secret.yml.tmpl
@@ -6,5 +6,5 @@ metadata:
 data:
   # insert your base64 encoded DO access token here, ensure there's no trailing newline:
   # to base64 encode your token run:
-  #      echo -n "abc123abc123doaccesstoken" | base64
+  #      echo -n "abc123abc123doaccesstoken" | base64 | tr -d '\r\n'
   access-token: "REPLACE WITH BASE64 ENCODED DO ACCESS TOKEN"
--- a/scripts/generate-secret.sh
+++ b/scripts/generate-secret.sh
@@ -19,7 +19,7 @@ if [[ -z $DIGITALOCEAN_ACCESS_TOKEN ]]; then
   exit 1
 fi
 
-ENCODED_ACCESS_TOKEN=$(echo -n "$DIGITALOCEAN_ACCESS_TOKEN" | base64)
+ENCODED_ACCESS_TOKEN=$(echo -n "$DIGITALOCEAN_ACCESS_TOKEN" | base64 | tr -d '\r\n')
 
 GENERATED_SECRET=./generated-secret.yml
 trap "{ rm $GENERATED_SECRET; }" EXIT


andrewsykim:
hmm, I definitely ran into this before, hence why I put `echo -n`. 

> -n    Do not print the trailing newline character.  This may also be achieved by appending `\c' to the end of the string, as is done by iBCS2 compatible systems.  Note that this
           option as well as the effect of `\c' are implementation-defined in IEEE Std 1003.1-2001 (``POSIX.1'') as amended by Cor. 1-2002.  Applications aiming for maximum portabil-
           ity are strongly encouraged to use printf(1) to suppress the newline character.

Annoying how `base64` on macOS is `-b` instead of `-w`. 
```
$ base64 --help
Usage:	base64 [-hvD] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -D, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
```

We could have a check in place for that but rather avoid it if possible. Thoughts?

andrewsykim:
I think I misunderstood you here, it's `base64` that adds the new line so `echo -n` wouldn't make a diff here. 

MacOS:
```
$ echo -n "$DIGITALOCEAN_ACCESS_TOKEN" | base64
a2Fqc2RmbGthajEyajNrajEybDRrajJramZsa2phZmxrc2pmbGthajR0bGthajRrbGpmYWxramE0ZmFsNGY=
$ export DIGITALOCEAN_ACCESS_TOKEN=kajsdflkaj12j3kj12l4kj2kjflkjaflksjflkaj4tlkaj4kljfalkja4fal4f
```

Linux:
```
# export DIGITALOCEAN_ACCESS_TOKEN=kajsdflkaj12j3kj12l4kj2kjflkjaflksjflkaj4tlkaj4kljfalkja4fal4f
# echo -n "$DIGITALOCEAN_ACCESS_TOKEN" | base64
a2Fqc2RmbGthajEyajNrajEybDRrajJramZsa2phZmxrc2pmbGthajR0bGthajRrbGpmYWxramE0
ZmFsNGY=
```

xmudrii:
@andrewsykim `echo -n` doesn't help me on Linux. What do you think about using [`tr`](https://linux.die.net/man/1/tr) instead of `-w 0` or `-b`:
```
echo -n $DIGITALOCEAN_ACCESS_TOKEN | base64 | tr -d '\r\n'
```

That seems to return correct output for me, but not sure would it work on macOS.

andrewsykim:
👍 great idea!

xmudrii:
@andrewsykim updated :)

bhcleek:
It may not be worth worrying whether the base64 encoding produces newlines or not, because most base64 implementations will correctly ignore newlines in the value being decoded. To make sure the secret configuration preserves the newlines, we could use a yaml multiline string (i.e. `|`), instead of a single line string. That would keep the instructions simple and rely on the decoding (which is done correctly in Go) to handle the newlines when they exist.

andrewsykim:
LGTM

Hmm, didn't know that :). I'm gonna try to do some testing this week to verify that. That definitely seems like a better solution. In the meantime time, @xmudrii I think you're good to merge, will try to open a follow up PR later.

bhcleek:
@andrewsykim to help your research, I'll point you to a few resources.

In [Section 3.3](https://tools.ietf.org/html/rfc4648#section-3.3), the specification for Base64, RFC 4648, documents that derivative specifications (e.g. MIME) may state that non-alphabet (i.e. the alphabet used to encode data) characters should be ignored in encoded data.

To be strictly compliant with this, most software would need to tell a generic decoder whether or not non-alphabet characters should be ignored or cause the encoded data to be rejected. In practice, though, I've never seen a base64 decoder that will reject encoded data that contains `\r` or `\n` unless told otherwise, though.

Let's examine what Go does.

First, notice in `encoding/base64`, how the encoding (i.e. the encoding alphabet) isn't allowed to contain `\r` or `\n`: https://golang.org/src/encoding/base64/base64.go?s=1307:1349#L32

Next, notice how the padding characters aren't allowed to be `\r` or `\n`: https://golang.org/src/encoding/base64/base64.go?s=2092:2147#L60

`Decode` explicitly documents that `\r` and `\n` are ignored: https://golang.org/pkg/encoding/base64/#Encoding.Decode

andrewsykim:
That's super helpful, thanks @bhcleek :)

andrewsykim:
@xmudrii can you rebase please?

xmudrii:
@andrewsykim It should be alright now :+1:

andrewsykim:
@xmudrii thanks for the PR!

andrewsykim:
@bhcleek @xmudrii we should probably use stringData to simplify this: https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/51

PR #52, opened 2017-10-30
vendor dependencies with dep and upgrade to kubernetes 1.8.2

bhcleek:
Revendor all dependencies using `dep` instead of `govendor` and simultaneously upgrade to kubernetes 1.8.2 tags.


andrewsykim:
lgtm

PR #54, opened 2017-10-31
Makefile improvements

bhcleek:
* make sure `.PHONY` isn't the default target
* add golint target and fix problems identified by `golint`.
* clean up all packages and remove unnecessary build path
* add code quality checks to the `all` target
* add golint to the `ci` target
* separate `.PHONY` target dependencies

### DIFF
--- a/do/cloud.go
+++ b/do/cloud.go
@@ -30,7 +30,7 @@ import (
 	"k8s.io/kubernetes/pkg/controller"
 )
 
-const DOAccessTokenEnv string = "DO_ACCESS_TOKEN"
+const doAccessTokenEnv string = "DO_ACCESS_TOKEN"
 const providerName string = "digitalocean"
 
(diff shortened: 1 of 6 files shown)


do/cloud.go
```diff
@@ -30,7 +30,7 @@ import (
 	"k8s.io/kubernetes/pkg/controller"
 )
 
-const DOAccessTokenEnv string = "DO_ACCESS_TOKEN"
+const doAccessTokenEnv string = "DO_ACCESS_TOKEN"
```
andrewsykim: Did `golint` complain about this? Acronyms shouldn't be caps?

andrewsykim:
LGTM

do/cloud.go
bhcleek: There was no need to export it, so I changed it to be unexported...

Issue #24, opened 2017-09-20
Adding to running cluster


klausenbusk:
Hello

I tried adding `digitalocean-cloud-controller-manager` last night, but it didn't work out.
The cluster was created with `bootkube` and the manifest is updated to match [`v0.6.2`](https://github.com/kubernetes-incubator/bootkube/releases/tag/v0.6.2).

Here is the steps I did:
1. Add `--kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname` to the `kube-apiserver`.
2. Deploy `digitalocean-cloud-controller-manager`
3. Add `--cloud-provider=external` to `kube-apiserver`, `kube-controller-manager` and the `kubelets`
4. Change `--hostname-override` to `--hostname-override=%H`
5. Restart all the `kubelet`
6. `digitalocean-cloud-controller-manager` isn't adding labels like `region` or any other labels. `node.cloudprovider.kubernetes.io/uninitialized` isn't removed either.

Did I miss something, or isn't it supported updating a running cluster?

The cluster ended up crashing (restarting too many `kubelet` at the same time, and `digitalocean-cloud-controller-manager` then removed the node, so the pod was evicted) . In the end I did a rollback to the latest configuration, but I tried getting `digitalocean-cloud-controller-manager` working for a few hours first.

andrewsykim:
Sorry for not getting to this sooner. Can you show me some of the logs from `digitalocean-cloud-controller-manager`?

klausenbusk:
> Can you show me some of the logs from digitalocean-cloud-controller-manager?

Sorry, I can't. Everything got a bit "chaotic", so saving log wasn't the first priority.

I will try again on a test cluster sometime in the future, as automatic labeling is very useful.

But I can remember that I got some error, regard connecting to `169.254.169.254` (for metadata), but I solved that by changing some firewall rules.

BTW: Is the region logic even working? As I understand [the code](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/5a64f9a0729ece886c65c59dcc2a0ecbe7f3b6eb/do/cloud.go#L68), it is pulled from the metadata service on "start". So the `region` label will be set to the region of the droplet where `digital-cloud-controller-manager` is running from == the region label won't match if you run droplets in multiple regions.

odacremolbap:
that's right @klausenbusk

andrewsykim:
[...]

PR #56, opened 2017-11-18
Support clusters where nodeName is the private or public IP

### LINKED ISSUE #20: Handle case where node name is an IP address
A lot of DO Kubernetes clusters override the kubelet node name using the node's private IP, this guarantees node names are unique and that its name can be used to reach the node. 

DO CCM should consider cases where the node is in the form of an IP address.

klausenbusk:
Fix #20

---

I haven't tested this ([bad experience in the past](https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/24)), but it should work..

### DIFF
--- a/do/droplets.go
+++ b/do/droplets.go
@@ -203,6 +203,12 @@ func dropletByName(ctx context.Context, client *godo.Client, nodeName types.Node
 		if droplet.Name == string(nodeName) {
 			return &droplet, nil
 		}
+		addresses, _ := nodeAddresses(&droplet)
+		for _, address := range addresses {
+			if address.Address == string(nodeName) {
+				return &droplet, nil
+			}
+		}
 	}
 
 	return nil, cloudprovider.InstanceNotFound


do/droplets.go
```diff
@@ -200,8 +200,11 @@ func dropletByName(ctx context.Context, client *godo.Client, nodeName types.Node
 	}
 
 	for _, droplet := range droplets {
-		if droplet.Name == string(nodeName) {
```
andrewsykim: I think this check should still be in place. I see that `nodeAddresses` would still catch this cause `Hostname` address type would still match `nodeName`, but it's  explicit and might be hard to debug later. WDYT?

klausenbusk: If people is debugging they gonna look at `nodeAddresses` anyway, and I think it is obvious enough that it also handle `nodeHostname`, but that maybe just me?

You decide :)

andrewsykim: I think it's very obvious to you and me, but it may not always be obvious that the node hostname = droplet name and that the hostname is an address type. I think we should keep the `droplet.Name == string(nodeName)` check :)

klausenbusk: Fixed..

andrewsykim:
LGTM! Will merge when CI passes, had to restart the job.

klausenbusk:
@andrewsykim I forgot to update the README.

See PR: https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/59 , spelling isn't  what I'm best at, so it could probably be rephrased in a better way.

andrewsykim:
```
$ kubectl get no
NAME            STATUS    ROLES     AGE       VERSION
10.137.192.52   Ready     master    4m        v1.8.0+coreos.0
10.137.72.132   Ready     node      3m        v1.8.0+coreos.0
10.137.72.133   Ready     node      4m        v1.8.0+coreos.0
10.137.72.148   Ready     node      4m        v1.8.0+coreos.0
$ kubectl get no 10.137.72.132 -o yaml
apiVersion: v1
kind: Node
metadata:
  annotations:
    node.alpha.kubernetes.io/ttl: "0"
    volumes.kubernetes.io/controller-managed-attach-detach: "true"
  creationTimestamp: 2017-12-06T15:15:24Z
  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/instance-type: c-2
    beta.kubernetes.io/os: linux
    failure-domain.beta.kubernetes.io/region: tor1
    kubernetes.io/hostname: 10.137.72.132
    kubernetes.io/role: node
  name: 10.137.72.132
  resourceVersion: "1368"
  selfLink: /api/v1/nodes/10.137.72.132
  uid: 48677b9d-da98-11e7-b747-e279993afcc5
spec:
  externalID: 10.137.72.132
  providerID: digitalocean://73950518
status:
  addresses:
  - address: k8s-worker-03
    type: Hostname
  - address: 10.137.72.132
    type: InternalIP
  - address: 138.197.172.197
    type: ExternalIP
```

PR #57, opened 2017-11-18
Switch Docker base image to Alpine from Ubuntu

klausenbusk:
This reduce the image size by 75%.
Before this change the image size was 214MB, now it is only 50.2MB.

---

Note: Not tested..


andrewsykim:
I will give this a test and get back to you :)

andrewsykim:
Thanks for this! Image is much smaller now ;)

PR #58, opened 2017-11-18
Use stringData for Kubernetes Secret containing DO Token

### LINKED ISSUE #51: Use stringData for Kubernetes Secret containing DO Token.
The current examples use `data` in Kubernetes secrets which requires you to base64 encode your secrets. Kuberetes now supports `stringData` which is backwards compatible where the kubelet would end up consuming the data in base64 but we can use string data instead. 

In https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/49, we had a discussion around how we should handle new line characters when we base64 encode the DO token, this would address it. 

More details here https://github.com/kubernetes/kubernetes/issues/19575.

klausenbusk:
Fix #51

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,20 +66,15 @@ Copy [releases/secret.yml.tmpl](https://github.com/digitalocean/digitalocean-clo
 cp releases/secret.yml.tmpl releases/secret.yml
 ```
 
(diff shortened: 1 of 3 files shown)


andrewsykim:
Thanks so much for this!

PR #60, opened 2017-12-13
release v0.1.3

andrewsykim:
cc @klausenbusk

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v0.1.3 (alpha) - December 13th 2017
+* Support clusters where nodeName is the private or public IP (@klausenbusk)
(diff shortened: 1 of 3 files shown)


xmudrii:
I guess this is irrelevant and is very tiny nitpick but I've found one occurrence of `0.1.2` in the README file (example under CCM heading -- https://github.com/digitalocean/digitalocean-cloud-controller-manager#cloud-controller-manager).

andrewsykim:
@xmudrii good catch, that part of the README is just an example but it should be more generic. I'll update that :)

klausenbusk:
I'm wondering if there is a better approach than duplicating the manifest file on every release?

bhcleek:
LGTM

I think we can work out something better with the manifest separately.

andrewsykim:
Agreed, the release file seems unnesscary now since we're only updating image tags, but the idea was that it makes more complicated updates easier in the future.

Issue #44, opened 2017-10-07
Document RBAC requirements


andrewsykim:
For clusters using RBAC, we should add documentation/examples of a ClusterRole that can be used for external cloud providers. 

Related:
https://github.com/kubernetes/kubernetes/pull/53511
https://github.com/kubernetes/kubernetes.github.io/pull/5400#discussion_r140004296

andrewsykim:
Here's a ClusterRole I've used before
```yaml
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRole
metadata:
  name: system:cloud-controller-manager
  labels:
    kubernetes.io/cluster-service: "true"
rules:
- apiGroups:
  - ""
  resources:
  - nodes
  verbs:
  - "*"
- apiGroups:
  - ""
  resources:
  - services
  verbs:
  - list
  - watch
  - patch
- apiGroups:
  - ""
  resources:
  - events
  verbs:
  - create
  - patch
  - update
# For leader election
- apiGroups:
  - ""
  resources:
  - endpoints
  verbs:
  - create
  - get
  - list
  - watch
  - update
- apiGroups:
  - ""
  resources:
  - serviceaccounts
  verbs:
  - create
```

You'll need to create the service account and cluster role binding so you can reference it in the CCM deployment spec. 
```yaml
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: cloud-controller-manager
  namespace: kube-system
---
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1beta1
metadata:
  name: system:cloud-controller-manager
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: system:cloud-controller-manager
subjects:
- kind: ServiceAccount
  name: cloud-controller-manager
  namespace: kube-system
```

andrewsykim:
Can use a similar script as https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/scripts/generate-secret.sh to generate RBAC rules.

tamalsaha:
The list of required roles has changed. Based on https://github.com/kubernetes/website/blob/master/docs/concepts/architecture/cloud-controller.md and https://github.com/kubernetes/website/pull/6111

```
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRole
metadata:
  annotations:
[...]

PR #65, opened 2018-03-18
release v0.1.4 - support sticky sessions and add RBAC

andrewsykim:
cc @xmudrii @bhcleek 

Fixes https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/44

### DIFF
--- a/examples/loadbalancers/http-with-sticky-sessions.yml
+++ b/examples/loadbalancers/http-with-sticky-sessions.yml
@@ -1,0 +1,38 @@
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: http-lb
+  annotations:
+    service.beta.kubernetes.io/do-loadbalancer-protocol: "http"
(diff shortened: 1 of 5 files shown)


examples/loadbalancers/http-with-sticky-sessions.yml
```diff
@@ -0,0 +1,38 @@
+---
```
andrewsykim: screen shot 2018-03-18 at 6 52 02 pm

xmudrii: Is this the behavior?

andrewsykim: Yeah! Just putting it here to show what the sticky session config will look like based on the example

xmudrii: I'm happy to see it's working as expected! :blush:

andrewsykim: Same! Thanks again for the contribution :)

PR #66, opened 2018-03-26
travis: use new Go version

fatih:
Needs quoting because of how YAML interprets the value, see: https://twitter.com/dgryski/status/964626204488695808

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: go
 
 go:
-  - "1.10"
+  - 1.9.x
+  - "1.10.x"
   - tip
 
 dist: trusty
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.8.3
+  - "1.10"
```
andrewsykim: Can we add a 1.9.X version? I think we should keep testing whatever version is used upstream.

fatih:
Added `1.9.x` which will pick up 1.9.4 for now or any new versions in the future (Such as security fixes)

andrewsykim:
lgtm

PR #70, opened 2018-04-01
Handle firewall rule creation for load balancers

ghubcoder:
This adds the ability for digitalocean-cloud-controller-manager to inspect the droplets which are part of a newly created load balancer and check whether they have an attached firewall. 

If a firewall is present, it then checks to see if a firewall rule is required to allow the load balancer to communicate with the droplet on the target port. 

Upon deletion of the load balancer, the reverse happens and the firewall rules are removed. 

I've tested this and it does appear to work. I think perhaps the firewall function mocks need pulling out into their own file and more tests need adding. 

There also needs to be some cleanup of the error handling/logging messages. Would be good to see if this works for anyone else.

### DIFF
--- a/do/firewalls.go
+++ b/do/firewalls.go
@@ -52,6 +52,7 @@ func checkIfPortInRange(port int, fwPortRange string) bool {
 
 }
 
+// Rule only requires deletion if it is not in a range
 func checkIfPortRequiresDeletion(port int, fwPortRange string) bool {
 
 	fwPortList := strings.Split(fwPortRange, "-")
@@ -84,16 +85,15 @@ func createRuleRequest(sourcelb string, targetPort int) *godo.FirewallRulesReque
 	return rr
 }
 
-// Check if a firewall rule already exists for a load balancer
-// forwarding rule ( returns true if covered by ALL, a port range or a single port)
-func lbfwRuleExists(lbID string, lbRule godo.ForwardingRule, firewalls []godo.Firewall) bool {
+// Check if a firewall rule exists for a load balancer forwarding rule
+func lbfwRuleExists(lbID string, lbRule godo.ForwardingRule, firewalls []godo.Firewall, checkFunc func(port int, fwPortRange string) bool) bool {
 	ruleExists := false
 	//check each firewall to see if rule already exists
 	for _, firewall := range firewalls {
 		for _, fwInboundRule := range firewall.InboundRules {
 			for _, fwSourcelbs := range fwInboundRule.Sources.LoadBalancerUIDs {
 				if fwSourcelbs == lbID {
-					res := checkIfPortInRange(lbRule.TargetPort, fwInboundRule.PortRange)
+					res := checkFunc(lbRule.TargetPort, fwInboundRule.PortRange)
 					if res && fwInboundRule.Protocol == "tcp" {
 						ruleExists = true
 					}
@@ -117,28 +117,6 @@ func requestExists(requests []firewallRequest, firewallID string, lbID string, t
 	}
 
 	return false
-}
-
-// Check if a single port rule exists for a load balancer
-// forwarding rule ( will not remove ranges defined by ports or ALL )
-func lbfwRuleRequiresDeletion(lbID string, lbRule godo.ForwardingRule, firewalls []godo.Firewall) bool {
-	requiresDeletion := false
-	//check each firewall to see if rule need deleting
-	for _, firewall := range firewalls {
-		for _, fwInboundRule := range firewall.InboundRules {
-			for _, fwSourcelbs := range fwInboundRule.Sources.LoadBalancerUIDs {
-				if fwSourcelbs == lbID {
-					res := checkIfPortRequiresDeletion(lbRule.TargetPort, fwInboundRule.PortRange)
-					if res && fwInboundRule.Protocol == "tcp" {
-						requiresDeletion = true
-					}
-				}
-			}
-		}
-	}
-
-	return requiresDeletion
-
 }
 
 func addRules(client *godo.Client, rulesToAdd []firewallRequest) error {
@@ -181,7 +159,7 @@ func (l *loadbalancers) EnsureFWRuleExists(lb *godo.LoadBalancer) ([]firewallReq
 		}
 
 		for _, lbRule := range lb.ForwardingRules {
-			if lbfwRuleExists(lb.ID, lbRule, firewalls) == false &&
+			if lbfwRuleExists(lb.ID, lbRule, firewalls, checkIfPortInRange) == false &&
 				requestExists(rulesToAdd, firewalls[0].ID, lb.ID, lbRule.TargetPort) == false {
 
 				rulesToAdd = append(rulesToAdd, firewallRequest{firewalls[0].ID, createRuleRequest(lb.ID, lbRule.TargetPort)})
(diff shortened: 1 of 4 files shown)


cagedmantis:
Hey @ghubcoder apologies for the delay, I've been getting ramped up on the controller code. Taking a look now.

do/loadbalancers.go
```diff
@@ -179,6 +184,126 @@ func (l *loadbalancers) EnsureLoadBalancer(clusterName string, service *v1.Servi
...
+
+// EnsureFirewall ensures that the node firewalls have the correct rules for
+// the load balancer.
+//
+// EnsureFirewall will not modify service or nodes.
+func (l *loadbalancers) EnsureFirewall(lb *godo.LoadBalancer) error {
+	var ruleExists bool
+	var rulesToAdd []firewallRequest
+	var alreadyAdded bool
+	//var completedFirewalls []string
```
cagedmantis: Remove unused variable

do/loadbalancers.go
```diff
@@ -179,6 +184,126 @@ func (l *loadbalancers) EnsureLoadBalancer(clusterName string, service *v1.Servi
...
+type firewallRequest struct {
+	firewallId string
+	rules      *godo.FirewallRulesRequest
+}
+
+// EnsureFirewall ensures that the node firewalls have the correct rules for
+// the load balancer.
+//
+// EnsureFirewall will not modify service or nodes.
+func (l *loadbalancers) EnsureFirewall(lb *godo.LoadBalancer) error {
```
cagedmantis: The function name isn't really describing what this function does. It's ensuring that the firewall rules allow a load balancer. Not that a firewall exists.

do/loadbalancers.go
```diff
@@ -179,6 +184,126 @@ func (l *loadbalancers) EnsureLoadBalancer(clusterName string, service *v1.Servi
...
+	var ruleExists bool
+	var rulesToAdd []firewallRequest
+	var alreadyAdded bool
+	//var completedFirewalls []string
+
+	for _, dropletid := range lb.DropletIDs {
```
cagedmantis: Variable should be `dropletID`

do/loadbalancers.go
```diff
@@ -179,6 +184,126 @@ func (l *loadbalancers) EnsureLoadBalancer(clusterName string, service *v1.Servi
...
+		firewalls, _, err := l.client.Firewalls.ListByDroplet(context.TODO(), dropletid, nil)
+		if err != nil {
+			return fmt.Errorf("Error listing firewalls for droplet %d", dropletid)
+		}
+
+		if len(firewalls) > 0 {
```
cagedmantis: `if len(firewalls) == 0 { continue }` would reduce nesting.

do/loadbalancers.go
```diff
@@ -222,10 +347,73 @@ func (l *loadbalancers) EnsureLoadBalancerDeleted(clusterName string, service *v
...
+	err = l.EnsureFirewallDeleted(lb)
+	if err != nil {
+		return err
+	}
+
 	_, err = l.client.LoadBalancers.Delete(ctx, lb.ID)
 	return err
 }
 
+func (l *loadbalancers) EnsureFirewallDeleted(lb *godo.LoadBalancer) error {
```
cagedmantis: Function name is misleading. Perhaps `FirewallRuleDeleted`.

do/loadbalancers.go
```diff
@@ -179,6 +184,126 @@ func (l *loadbalancers) EnsureLoadBalancer(clusterName string, service *v1.Servi
...
+			return fmt.Errorf("Error listing firewalls for droplet %d", dropletid)
+		}
+
+		if len(firewalls) > 0 {
+			for _, lbRule := range lb.ForwardingRules {
+				ruleExists = false
+				//check each firewall to see if rule already exists
+				for _, firewall := range firewalls {
+					for _, fwInboundRule := range firewall.InboundRules {
+						for _, fwSourcelbs := range fwInboundRule.Sources.LoadBalancerUIDs {
```
cagedmantis: This logic can be broken up into functions. Checking if the rule exists, deleting the rule, creating the rule.

do/loadbalancers.go
```diff
@@ -222,10 +347,73 @@ func (l *loadbalancers) EnsureLoadBalancerDeleted(clusterName string, service *v
...
+
+	for _, dropletid := range lb.DropletIDs {
+
+		firewalls, _, err := l.client.Firewalls.ListByDroplet(context.TODO(), dropletid, nil)
+
+		if err != nil {
+			return fmt.Errorf("Error listing firewalls for droplet %d", dropletid)
+		}
+
+		if len(firewalls) > 0 {
```
cagedmantis: Similar comments to the ones noted above.

cagedmantis:
It would make things cleaner if they were moved to a firewall.go file.

cagedmantis:
Thanks for the submission. I'm excited to use this.

ghubcoder:
@cagedmantis thanks for this, I'll take a look at making the changes.

do/firewalls.go
```diff
@@ -0,0 +1,235 @@
...
+			return true
+		}
+	}
+
+	return false
+}
+
+// Check if a single port rule exists for a load balancer
+// forwarding rule ( will not remove ranges defined by ports or ALL )
+func lbfwRuleRequiresDeletion(lbID string, lbRule godo.ForwardingRule, firewalls []godo.Firewall) bool {
```
ghubcoder: This function is pretty much the same as lbfwRuleExists, only difference is it makes a call to checkIfPortRequiresDeletion, rather than checkIfPortInRange. These are slightly different as we only delete a port if it's not included in a range of ports all defined with "ALL". Whereas we only add a port if it's not contained within a list of ranges. 

These two functions (lbfwRuleExists, lbfwRuleRequiresDeletion) could be replaced by one function, and the checking function required could be passed in as a parameter. Although I'm not sure if that is good practice.

do/loadbalancers.go
```diff
@@ -179,6 +184,126 @@ func (l *loadbalancers) EnsureLoadBalancer(clusterName string, service *v1.Servi
 
 }
 
+func checkIfPortInRange(port int, fwPortRange string) bool {
+
+	fwPortList := strings.Split(fwPortRange, "-")
```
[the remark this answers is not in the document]
ghubcoder: I had a look at this, the only issue is that the first value can sometimes be a string containing "all", rather than just two integers. Need to think about how to clean this up.

ghubcoder:
I've tested it once more, seems to work ok after the changes. Need to think about adding some tests now for the firewalls.go code.

timoreimann:
I wonder what the status of this PR is? @ghubcoder, seems like you added the remaining tests, is that correct?

pritpal-sabharwal:
For what it is worth, I have merged this PR locally and am running it on a K8S cluster. It seems to work. Creating a Service creates a LB and updates the Firewalls for the relevant droplets. Upon deletion of the service the LB is deleted and the Firewall rules are removed too.

Not sure if there is anything else left to do here to get this over the line.

ghubcoder:
@pritpal-sabharwal thanks for giving this a try, very much appreciated !

do/loadbalancers.go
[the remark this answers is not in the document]
Sandradeb5: kind : Service 
apiVersion : v1 
metadata :
   name : bonjour 
  annotations :
     service.beta.kubernetes.io/do-loadbalancer-certificate-id : " 1234-5678-9012-3456 "
     service.beta.kubernetes.io/do-loadbalancer- protocole : " https " 
    # Décommentez une fois que hello.example.com pointe vers l'adresse IP externe de l'équilibreur de charge DO. 
    # service.beta.kubernetes.io/do-loadbalancer-hostname: "hello.example.com" 
spec :
   type : LoadBalancer 
  sélecteur :
    app : hello 
  ports :
    - Nom : https 
      protocole : TCP 
      Port : 443 
      targetPort : 80
---
apiVersion : apps / v1 
kind : Métadonnées de déploiement
 :
   nom : bonjour spec :
   répliques : 3 sélecteur :
     matchLabels :
       app : bonjour modèle :
     metadata :
       labels :
         app : bonjour spec :
       conteneurs :

  
  
    
      - nom : bonjour 
        image : ports snormore / bonjour
         :
        - protocole containerPort : 80
           : TCP

timoreimann:
I'm going to close this PR. Apart from the question of whether there's still interest, we have pending PR #332 to take a different approach to opening up and closing down ports dynamically through a dedicated, managed firewall. While the PR targets NodePorts, theoretically it can be extended for LB ports as well.

PR #73, opened 2018-04-04
csi: implement IdentityServer and add test suite

fatih:
* Add initial `IdentityServer` implementation of the DO driver.
* This also adds the kubernetes-csi test suite which enables us to test all driver interface methods. However because we didn't implemented yet all, it's behind a `-suite` flag that you have to pass explicitly. This is to prevent our CI to fail for all following PR's if we merge it.

Please review: https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/73/commits/617ca279322219d8920651e79714ddb26c84160a


do/csi/driver/server_test.go
```diff
@@ -0,0 +1,53 @@
...
+	go func(drv *Driver) {
+		errCh <- drv.Run()
+	}(drv)
+
+	select {
+	case err := <-errCh:
+		if err != nil {
+			t.Fatal(err)
+		}
+	case <-time.After(time.Second * 2):
```
lxfontes: in the future, will we be able to test this without the sleep?

lxfontes:
lgtm

do/csi/driver/server_test.go
fatih: The test is running against a blocking server. This is just a leeway so we can capture the error. Unfortunately I can't call `t.Fatal` in a goroutine itself so I only have this technique currently.  Happy to change it to something different.

fatih: Ok removed it. The test suite has an internal timeout of `one minute` better than nothing.

PR #74, opened 2018-04-06
csi: implement ControllerServer interface

fatih:
This PR implements the ControllerServer plugin part of the [CSI Spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#container-storage-interface).  It's not tested yet, but it's good to go to be reviewed. 

- [x] Add tests


do/csi/driver/controller.go
```diff
@@ -3,48 +3,296 @@ package driver
...
+		Capabilities: caps,
+	}, nil
+}
+
+// extractStorage extracts the storage size in GB from the given capacity
+// range. If the capacity range is not satisfied it returns the default volume
+// size.
+func extractStorage(capRange *csi.CapacityRange) (int64, error) {
+	if capRange == nil {
+		return defaultVolumeSizeInGB, nil
```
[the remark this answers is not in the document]
fatih: This happens if the CO (in our case Kubernetes) doesn't set the size of the requested volume. In this case we need to define a standard our self. I've set it to 16GB, but this might change to anything else.

do/csi/driver/driver_test.go
```diff
@@ -16,14 +16,23 @@ func TestDriverSuite(t *testing.T) {
...
 		t.Fatalf("failed to remove unix domain socket file %s, error: %s", socket, err)
 	}
 
 	// run the driver
-	go NewDriver(endpoint).Run()
+	driver, err := NewDriver(endpoint, token)
+	if err != nil {
+		t.Fatal(err)
+	}
+	go driver.Run()
```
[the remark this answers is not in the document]
fatih: Good question. I'll explose the Stop() method of the internal gRPC server. That would allow the user stop in various way, such as via `defer driver.Stop()`.

lxfontes:
lgtm

do/csi/driver/controller.go
```diff
@@ -3,48 +3,347 @@ package driver
...
+
+	size, err := extractStorage(req.CapacityRange)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	volumeName := req.Name
+
+	// get volume first, if it's created do no thing
+	volumes, _, err := d.doClient.Storage.ListVolumes(ctx, &godo.ListVolumeParams{
```
andrewsykim: Something we can do later, but if you don't paginate you'll only get the first 25 volumes.

andrewsykim: oh nvm, seems like we're doing a filter on volume name so don't think it matters anyways

do/csi/driver/controller.go
```diff
@@ -3,48 +3,347 @@ package driver
...
+	KB = 1 << (10 * iota)
+	MB
+	GB
+	TB
+)
+
+const (
+	defaultVolumeSizeInGB = 16 * GB
+
+	createdByDO = "Created by DigitalOcean CSI driver"
```
andrewsykim: Do you know if there's any other metadata we can add here that would be passed into the CSI driver (like cluster name, etc)?

do/csi/driver/controller.go
```diff
@@ -3,48 +3,347 @@ package driver
...
+		return &csi.CreateVolumeResponse{
+			Volume: &csi.Volume{
+				Id:            vol.ID,
+				CapacityBytes: vol.SizeGigaBytes * GB,
+			},
+		}, nil
+	}
+
+	volumeReq := &godo.VolumeCreateRequest{
+		Region:        d.nodeId,
```
andrewsykim: should be d.region?

do/csi/driver/controller.go
```diff
@@ -3,48 +3,347 @@ package driver
...
-// accesible by anymore.
-func (d *Driver) DeleteVolume(context.Context, *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
-	return nil, errors.New("not implemented")
+// DeleteVolume deletes the given volume. The function is idempotent.
+func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
+	if req.VolumeId == "" {
+		return nil, status.Error(codes.InvalidArgument, "DeleteVolume Volume ID must be provided")
+	}
+
+	_, err := d.doClient.Storage.DeleteVolume(ctx, req.VolumeId)
```
andrewsykim: Not familiar enough with CSI yet to know what is calling this method but do we need to check volume description here?

do/csi/driver/controller.go
```diff
@@ -3,48 +3,347 @@ package driver
...
+	var volumes []godo.Volume
+	lastPage := 0
+	for {
+		vols, resp, err := d.doClient.Storage.ListVolumes(ctx, listOpts)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, vol := range vols {
+			volumes = append(volumes, vol)
```
andrewsykim: I think this can just be 
```
volumes = append(volumes, vols...)
```

do/csi/driver/controller.go
andrewsykim: Asking cause checking this description doesn't work if there are multiple clusters using this CSI driver.

andrewsykim:
LGTM

do/csi/driver/controller.go
fatih: The description is only checked if the driver is created by the CSI driver or not. We don't use it for anything else. For multiple K8S clusters, you're right. There is a way to pass down metadata inside the request:

```
	// Plugin specific parameters passed in as opaque key-value pairs.
	// This field is OPTIONAL. The Plugin is responsible for parsing and
	// validating these parameters. COs will treat these as opaque.
	Parameters map[string]string `protobuf:"bytes,4,rep,name=parameters" json:"parameters,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
```

However it's not clear to me what the CO (in our case Kubernetes) passes down. I'll be make sure to check it and incorporate any information into the description (such as including the cluster name, if available)

do/csi/driver/controller.go
fatih: oops, fixed it!

do/csi/driver/controller.go
fatih: Here is the full logic step by step (From https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/container-storage-interface.md):

screen shot 2018-04-08 at 12 15 40 am

In this case, because the Volume is bound to the PVC, I'm thinking how we can accidentally deleted it? Because we use the Volume ID that is bound to the PVC.

fatih:
Merging this, fixed the review comments from @andrewsykim as well. I'll look into the "Description" model more once I have a plugin that is usable.

do/csi/driver/controller.go
```diff
@@ -3,48 +3,345 @@ package driver
...
+		return nil, status.Error(codes.InvalidArgument, "ControllerPublishVolume Volume ID must be provided")
+	}
+
+	dropletID, err := strconv.Atoi(req.NodeId)
+	if err != nil {
+		return nil, fmt.Errorf("malformed nodeId %q detected: %s", req.NodeId, err)
+	}
+
+	// TODO(arslan): wait volume to deattach
+	_, resp, err := d.doClient.StorageActions.DetachByDropletID(ctx, req.NodeId, dropletID)
```
renan: I believe this needs a `s/req.NodeId/req.VolumeId`, should be:
`_, resp, err := d.doClient.StorageActions.DetachByDropletID(ctx, req.VolumeId, dropletID)`

PR #75, opened 2018-04-06
 loadbalancers: Support nodes where nodeName is the private or public IP 

### LINKED ISSUE #69: LBs does not account for nodes using IPs as node names
When we add droplets to LBs we only check if the node name matches the droplet names, however, it's common for droplets to be named after their public/private IPs so it should check for that as well. Here's where the check is https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/do/loadbalancers.go#L259.

klausenbusk:
Fix #69

### DIFF
--- a/do/loadbalancers.go
+++ b/do/loadbalancers.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/digitalocean/godo"
 	"github.com/digitalocean/godo/context"
+	"github.com/golang/glog"
 )
 
 const (
@@ -255,10 +256,22 @@ func (l *loadbalancers) nodesToDropletIDs(nodes []*v1.Node) ([]int, error) {
 
 	var dropletIDs []int
 	for _, node := range nodes {
+	Loop:
 		for _, droplet := range droplets {
 			if node.Name == droplet.Name {
 				dropletIDs = append(dropletIDs, droplet.ID)
 				break
+			}
+			addresses, err := nodeAddresses(&droplet)
+			if err != nil {
+				glog.Errorf("error getting node addresses for %s: %v", droplet.Name, err)
+				continue
+			}
+			for _, address := range addresses {
+				if address.Address == string(node.Name) {
+					dropletIDs = append(dropletIDs, droplet.ID)
+					break Loop
+				}
 			}
 		}
 	}
--- a/do/common.go
+++ b/do/common.go
@@ -19,6 +19,8 @@ package do
 import (
 	"fmt"
 
+	"k8s.io/api/core/v1"
+
 	"github.com/digitalocean/godo"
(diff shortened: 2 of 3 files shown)


do/loadbalancers.go
```diff
@@ -255,11 +255,19 @@ func (l *loadbalancers) nodesToDropletIDs(nodes []*v1.Node) ([]int, error) {
 
 	var dropletIDs []int
 	for _, node := range nodes {
+	Loop:
 		for _, droplet := range droplets {
 			if node.Name == droplet.Name {
 				dropletIDs = append(dropletIDs, droplet.ID)
 				break
 			}
+			addresses, _ := nodeAddresses(&droplet)
```
andrewsykim: Could we handle this error here

andrewsykim:
@klausenbusk thanks so much for this PR. Just had one minor comment and it should good to go

do/loadbalancers.go
klausenbusk: I think we should only log it for now or change the `nodeAddresses` function, right now a single node without `private network` would cause a error. Why do we require `private network` anyway? It should be optional.
https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/57618014f2a6248c220d15383bcbdd24a6cb7485/do/droplets.go#L75-L92

We do not handle error in https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/57618014f2a6248c220d15383bcbdd24a6cb7485/do/droplets.go#L206 either.

andrewsykim: That is very reasonable, just logging works for me :)

andrewsykim:
sorry @klausenbusk one more comment

do/loadbalancers.go
```diff
@@ -255,11 +256,22 @@ func (l *loadbalancers) nodesToDropletIDs(nodes []*v1.Node) ([]int, error) {
...
 	for _, node := range nodes {
+	Loop:
 		for _, droplet := range droplets {
 			if node.Name == droplet.Name {
 				dropletIDs = append(dropletIDs, droplet.ID)
 				break
 			}
+			addresses, err := nodeAddresses(&droplet)
+			if err != nil {
+				glog.Errorf("error getting node addresses for %s: %v", droplet.Name, err)
```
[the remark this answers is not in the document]
klausenbusk: Done, are you sure about the `panic`? The [spec says](https://golang.org/ref/spec) `For a nil slice, the number of iterations is 0.`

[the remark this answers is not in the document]
andrewsykim: TIL :)

andrewsykim:
LGTM thanks!

PR #81, opened 2018-04-11
csi: various improvements and fixes

fatih:
* Include mkfs executables to the plugin image so we can format volumes
* Remove the unix socket whenever we start the plugin, to avoid getting an error because of socket usage
* Vendor sirupsen/logrus and add initial logging to the CSI plugin for getting a better overview of what's going on.
* Wait until volume actions are completed
* Do not format the volume again when we mount it from the global directory to the pod directory

Please reviews the individual commits if needed. Each commit is self contained.


romansky:
Closely following the merges :) need this to test ;)
BTW, need to add note about needing the NodeId to equal the DO ID (found this important trivia piece thanks to comments in the code..)

fatih:
> Any reason for not using glog? I don't necessarily think i'ts better than logrus, it would only be for consistency with CCM and other things.

@andrewsykim the plugin will be splitted out eventually in its own repo, so we don't have the requirement of using glog at all. It doesn't depend on Kubernetes (which glog uses so it's a free dependency that comes with it). I wanted to use a better logger, with KV support and more Go idiomatic.

lxfontes:
lgtm
wrt logrus: just to reiterate to clean it up once we split repos ❤️

PR #87, opened 2018-04-19
vendor update github.com/spf13/pflag

andrewsykim:
Updates github.com/spf13/pflag which fixes an issue with noisy logging from glog. More details here  https://github.com/kubernetes/kubernetes/issues/17162#issuecomment-378224717

Ran
```
dep ensure -update github.com/spf13/pflag
```


cagedmantis:
lgtm. Waiting on @bhcleek

bhcleek:
Can you run `dep prune` and amend the commit? All the extra files here are because of that missing step. I'll create a separate PR to make sure it happens automatically in the future.

PR #89, opened 2018-04-19
move apiextensions-apiserver and client-go to overrides

bhcleek:
k8s.io/apiextensions-apiserver and k8s.io/client-go are transitive
dependencies, not direct dependencies, so need to be in the overrides
instead of the constraints.

After moving them and running dep ensure,the apiextensions-apiserver was
properly updated (it must have previously been vendored from the
kubernetes-1.8.1 tag that last time I upgraded k8s.io packages).


andrewsykim:
LGTM

PR #90, opened 2018-04-23
Add the ability to overide the DO API address.

cagedmantis:


do/cloud.go
```diff
@@ -63,11 +72,14 @@ func newCloud(config io.Reader) (cloudprovider.Interface, error) {
 	}
 
 	oauthClient := oauth2.NewClient(oauth2.NoContext, tokenSource)
-	doClient := godo.NewClient(oauthClient)
+	doClient, err := godo.New(oauthClient, opts...)
+	if err != nil {
+		return nil, errors.New("failed to create godo client")
```
andrewsykim: `fmt.Errorf()` to include the error in the return value

cagedmantis: Sure. Would you like me to do the same for the dropletRegion call?

andrewsykim: If you don't mind :)

andrewsykim:
lgtm, thanks for the clean up!

PR #91, opened 2018-04-27
Adding simple logging to load balancer.

cagedmantis:
- Add simple logging to load balancer related actions.
- Remove an unused variable.
- Removing a noisy log call which logs an entry for each non-k8s node you have on your DigitalOcean account.

### DIFF
--- a/do/loadbalancers.go
+++ b/do/loadbalancers.go
@@ -105,6 +105,8 @@ func newLoadbalancers(client *godo.Client, region string) cloudprovider.LoadBala
 //
 // GetLoadBalancer will not modify service.
 func (l *loadbalancers) GetLoadBalancer(clusterName string, service *v1.Service) (*v1.LoadBalancerStatus, bool, error) {
+	glog.Infof("getting load balancer cluster=%s service=%s", clusterName, service.Name)
+
 	lbName := cloudprovider.GetLoadBalancerName(service)
 	lb, err := l.lbByName(context.TODO(), lbName)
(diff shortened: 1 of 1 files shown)


do/loadbalancers.go
```diff
@@ -264,7 +270,6 @@ func (l *loadbalancers) nodesToDropletIDs(nodes []*v1.Node) ([]int, error) {
 			}
 			addresses, err := nodeAddresses(&droplet)
 			if err != nil {
-				glog.Errorf("error getting node addresses for %s: %v", droplet.Name, err)
```
andrewsykim: This could be a legitimate error though?

do/loadbalancers.go
```diff
@@ -136,6 +138,8 @@ func (l *loadbalancers) GetLoadBalancer(clusterName string, service *v1.Service)
 //
 // EnsureLoadBalancer will not modify service or nodes.
 func (l *loadbalancers) EnsureLoadBalancer(clusterName string, service *v1.Service, nodes []*v1.Node) (*v1.LoadBalancerStatus, error) {
+	glog.Infof("ensuring load balancer cluster=%s service=%s", clusterName, service.Name)
```
andrewsykim: These logs could get pretty noisy since `EnsureLoadbalancer` can run during a resync. There's also some logging that happens in the controller https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/service/service_controller.go#L284-L310.

PR #93, opened 2018-05-09
Upgrade all Kubernetes dependencies to v1.10.2

andrewsykim:
Notes:
* in v1.10 all methods in the cloud provider interface now expect context.Context as the first argument. This requires us to import both go context and godo context :( 
* Cloud Controller Manager now uses component config so had to update main.go accordingly. 
* add override for `--allow-untagged-cloud=true` which allows CCM to run even though cluster ID is not set
* updated godo to v1.2.0 which means we can use the context passed by the controller instead of using godo/context


fatih:
lgtm 👍

andrewsykim:
Forgot to run `dep prune`, much better now 😆

bhcleek:
`github.com/digitalocean/godo/context` doesn't exist anymore. Instead of refactoring to use a package that's been removed from godo, why not just use the stdlib `context` package for all those functions to which you added context?

andrewsykim:
Ah, had no idea it was removed, I'll update it to use stdlib context, I much prefer that!

bhcleek:
lgtm

PR #98, opened 2018-05-10
[load_balancer] add annotation for redirect_http_to_https option

peterver:
Add the option to specify redirect_http_to_https on the DO load balancer resource using the annotation

```
```

As per the specification at https://developers.digitalocean.com/documentation/v2/#load-balancers this defaults to false.

### DIFF
--- a/do/loadbalancers.go
+++ b/do/loadbalancers.go
@@ -72,9 +72,9 @@ const (
 	// annDOStickySessionType is set to cookies.
 	annDOStickySessionsCookieTTL = "service.beta.kubernetes.io/do-loadbalancer-sticky-sessions-cookie-ttl"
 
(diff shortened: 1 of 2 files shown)


andrewsykim:
Hey @peterver changes look great! 

Can we add a test case for the redirect annotation [here](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/do/loadbalancers_test.go#L1083). Otherwise I think we're good to :shipit:.

peterver:
@andrewsykim will do ! :)

peterver:
@andrewsykim I've added a test case for the getRedirectHttpToHttps function, any other test case you would like me to look at ?

peterver:
@andrewsykim I added the buildLoadBalancerRequest test for http_to_https redirect :)

andrewsykim:
LGTM! Thanks!

PR #105, opened 2018-06-06
conform to kubernetes cloud provider repo requirements

andrewsykim:
As per https://github.com/kubernetes/community/blob/master/keps/sig-cloud-provider/0002-cloud-controller-manager.md#repository-requirements


Makefile
```diff
@@ -23,11 +23,11 @@ clean:
 
 .PHONY: compile
 compile:
-	CGO_ENABLED=0 GOOS=linux go build
+	CGO_ENABLED=0 GOOS=linux go build github.com/digitalocean/digitalocean-cloud-controller-manager/cloud-controller-manager/cmd/digitalocean-cloud-controller-manager
```
andrewsykim: This doesn't make much sense now, but it will when we move this into `github.com/kubernetes/cloud-provider-digitalocean`

bhcleek: Why reference it by fully qualified package path? Wouldn't `./cloud-controller-manager/cmd/digitalocean-cloud-controller-manager` be better for preparing for the time when this is moved to `github.com/kubernetes/cloud-provider-digitalocean`?

andrewsykim:
Good catch @bhcleek, was ignored by `.gitignore`, PTAL

bhcleek:
lgtm

Issue #94, opened 2018-05-09
Implement InstanceShutdownByProviderID


andrewsykim:
As added in https://github.com/kubernetes/kubernetes/pull/60009

PR #109, opened 2018-07-03
Implement InstanceShutdownByProviderID

andrewsykim:
fixes https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/94 

Adds taints to node if it is shutdown

other refs: https://github.com/kubernetes/kubernetes/issues/58635

### DIFF
--- a/do/droplets.go
+++ b/do/droplets.go
@@ -31,6 +31,10 @@ import (
 	"github.com/digitalocean/godo"
 )
 
+const (
+	dropletShutdownStatus = "off"
+)
+
 type instances struct {
 	client *godo.Client
 	region string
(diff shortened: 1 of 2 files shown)


yastij:
@andrewsykim - this is cool, the taint should be active for 1.12, Thanks !

andrewsykim:
@yastij I had thought it was in v1.11 https://github.com/kubernetes/kubernetes/blob/release-1.11/pkg/controller/cloud/node_controller.go#L254-L270

yastij:
@andrewsykim - applying the taint is done since 1.11 but the detach logic will land in 1.12 see kubernetes/kubernetes#66181

andrewsykim:
@yastij ahh I see, thanks for bringing that to my attention. Is that for in-tree volume attach/detach only?

yastij:
it’ll be implemented for in-tree and out-tree.

PR #112, opened 2018-07-21
set prune options

bhcleek:
Set prune options and update vendor directory afterwards by running `dep
ensure`.

I was originally concerned about the differences between the result of using the prune options and `dep prune`. In https://github.com/golang/dep/issues/1820, Sam Boyer says they're expected.

With this change, `dep ensure` will prune automatically, and there's no need to run `dep prune` explicitly.


andrewsykim:
lgtm

PR #114, opened 2018-07-31
Add ldflags to go build so --version prints proper git version

andrewsykim:


Makefile
```diff
@@ -23,7 +23,7 @@ clean:
 
 .PHONY: compile
 compile:
-	CGO_ENABLED=0 GOOS=linux go build ./cloud-controller-manager/cmd/digitalocean-cloud-controller-manager
+	CGO_ENABLED=0 GOOS=linux go build -ldflags '-X github.com/digitalocean/digitalocean-cloud-controller-manager/vendor/k8s.io/kubernetes/pkg/version.gitVersion=$(VERSION)' ./cloud-controller-manager/cmd/digitalocean-cloud-controller-manager
```
andrewsykim: https://github.com/kubernetes/kubernetes/blob/master/pkg/version/version.go#L26-L42

bhcleek:
Instead of setting a vendored packages unexported variable at linking time, we should probably just expose our own in `cloud-controller-manager/cmd/digitalocean-cloud-controller-manager/main.go` and set _it_ at linking time.

andrewsykim:
Can you expand on that a bit more? The consumer of the `version` package is also vendored (see [this](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/vendor/k8s.io/kubernetes/cmd/cloud-controller-manager/app/controllermanager.go#L67)) and the gitVersion variable used by apimachinery is unexported (see [this](https://github.com/kubernetes/kubernetes/blob/master/pkg/version/base.go#L58)). Maybe I'm missing something?

bhcleek:
Ah, makes sense to me.

Lgtm

Issue #117, opened 2018-08-03
Healthcheck Path


scottrobertson:
Hey

Is there a way to define the healthcheck path for the load balancers?

My problem is that the root is protected by Basic Auth, so it always fails the health check. There is a specific `/health` endpoint i need to call to see if the app is alive or not.

andrewsykim:
@scottrobertson please take a look at https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/117

scottrobertson:
@andrewsykim awesome, thank you. Looks good!

scottrobertson:
Although, i actually switched to using Voyager as i could not work out a way to use 1 load balancer as an ingress for multiple services (using different domains) while using this CCM directly.

andrewsykim:
@scottrobertson yeah, unfortunately, LoadBalancer correlates to Services so there's no simple way to do this (yet). I think a common pattern is to put a LB infront of the ingress so that at least you can do health checking on our ingresses behind a single IP.

scottrobertson:
@andrewsykim i used the following before i switched to DO's kubernetes (i used DO via Stackpoint before), and it worked perfectly. However, for some reason when using DO directly, it does not work.

```
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: ingress
spec:
  rules:
    - host: example.com
      http:
        paths:
        - path:
          backend:
            serviceName: example
            servicePort: 80
```

The ingress does not get an address.

It's probably due to it missing something that Stackpoint installed. Maybe haproxy, or nginx or something.

andrewsykim:
From what I know, you're supposed to have a Kubernetes Service pointing to the haproxy/nginx controller running the ingress. So DO LoadBalancer -> droplet node port -> ingress (haproxy/nginx/etc) -> pods

andrewsykim:
^ You may want to set `externalTrafficPolicy: Local` for the Service representing the ingress

andrewsykim:
Hey Scott, I have a PR open for this here https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/118 but it fell off my radar, will try to get to it later this week, sorry!

scottrobertson:
Thank you 👍

andrewsykim:
Verified that https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/118 fixed this, I'll make a release with this fix soon :)

[...]

PR #118, opened 2018-08-06
add support for loadbalancer health check paths via service annotations

andrewsykim:
Adds support for custom health check paths (defaults to "/") using annotation `service.beta.kubernetes.io/do-loadbalancer-healthcheck-path` on a Service. 

Fixes https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/117


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -476,6 +482,17 @@ func getProtocol(service *v1.Service) (string, error) {
 	return protocol, nil
 }
 
+// getHealthCheckPath returns the desired path for health checking
+// health check path should default to / if not specified
+func getHealthCheckPath(service *v1.Service) string {
+	path, ok := service.Annotations[annDOHealthCheckPath]
+	if !ok {
+		return ""
```
bhcleek: Shouldn't this `return "/"`?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -476,6 +482,17 @@ func getProtocol(service *v1.Service) (string, error) {
 	return protocol, nil
 }
 
+// getHealthCheckPath returns the desired path for health checking
+// health check path should default to / if not specified
+func getHealthCheckPath(service *v1.Service) string {
```
bhcleek: (nit): There's no need for `get` in the name in Go.

cloud-controller-manager/do/loadbalancers.go
andrewsykim: I think godo passes in ""  but have to double check.

andrewsykim: godo passes in `""` for the request, meaning "let server side decide default" which I think is the correct behavior here.

andrewsykim:
@bhcleek address comments, PTAL

cloud-controller-manager/do/loadbalancers.go
bhcleek: The docs for the annotation says the default is `/`, though. Is the godoc there wrong, or this code path wrong?

PR #122, opened 2018-09-04
Add release process

fatih:
This PR simplifies making new releases and also adds instructions on how to make a new release. The goal is to easily ship new releases. The following changes have been made:

* The `PHONY` targets are grouped to make it easier to see all PHONY targets in one place
* Changed how we read the version string. This is now placed into `VERSION` flag. Every time we release a new version, this VERSION file will be also updated.
* Updated the `LDFLAGS` to include `COMMIT` SHA and the Git tree state
* Add echo logs to targets and hide the command itself via `@` to make it easier to read on the terminal
* Added instructions on how to develop and release new stuff
* Changed default `all` to just call `test` as it shouldn't publish a release in my opinion for safety. With that, I also added a guard against pushing to a stable version if it's **not** `master` branch.

We're using the same flow for https://github.com/digitalocean/csi-digitalocean now for a couple of releases and it makes everything much easier.


nanzhong:
lgtm

Makefile
```diff
@@ -12,46 +12,94 @@
...
+	$(eval VERSION = dev)
+	@CGO_ENABLED=0 GOOS=linux go build -ldflags "$(LDFLAGS)" ${PKG}
 
-.PHONY: build
 build:
-	docker build -t $(REGISTRY)/digitalocean-cloud-controller-manager:$(VERSION) -f cloud-controller-manager/cmd/digitalocean-cloud-controller-manager/Dockerfile .
+	@echo "==> Building the docker image"
+	@docker build -t $(REGISTRY)/digitalocean-cloud-controller-manager:$(VERSION) -f cloud-controller-manager/cmd/digitalocean-cloud-controller-manager/Dockerfile .
+
+build-dev:
```
andrewsykim: (not blocking): I see the convenience in this, but not sure we need a `build-dev` and `push-dev` target. The same can be accomplished with `VERSION=dev make push`.

andrewsykim:
lgtm

Makefile
fatih: Agree. Removed the `-dev` targets and changed the README description to `VERSION=dev make push`.

PR #123, opened 2018-09-04
Improving provider docs

andrewsykim:
as per https://github.com/kubernetes/community/blob/master/keps/sig-cloud-provider/0019-cloud-provider-documentation.md#goal-2


docs/controllers/services/annotations.md
```diff
@@ -0,0 +1,38 @@
...
+
+The default protocol for DO load balancers. Ports specified in the annotation `service.beta.kubernetes.io/do-loadbalancer-tls-ports` will be overwritten to https. Options are `tcp`, `http` and `https`. Defaults to `tcp`.
+
+### service.beta.kubernetes.io/do-loadbalancer-tls-ports 
+
+Specify which ports of the loadbalancer should use the https protocol. This is a comma separated list of ports (e.g. 443,6443,7443).
+
+### service.beta.kubernetes.io/do-loadbalancer-tls-passthrough
+
+Specify whether the DigitalOcean loadbalancer should pass encrypted data to backend droplets. This is optional. Options are `true` or `false`. Defaults to `false`.
```
fatih: Nit: we're talking sometimes about `DO load balancers`, here it is `DigitalOcean loadbalancer`. It seems like (according to the https://www.digitalocean.com/products/load-balancer/ page) it should be called:

```
* DigitalOcean Load Balancer
* DigitalOcean Load Balancers
```

So I suggest we always use this name whenever we want to refer to the load balancer product.

docs/controllers/services/annotations.md
```diff
@@ -0,0 +1,38 @@
...
+
+Specifies which stick session type the loadbalancer should use. Options are `none` or `cookies`.
+
+### service.beta.kubernetes.io/do-loadbalancer-sticky-sessions-cookie-ttl 
+
+Specifies the TTL of cookies used for loadbalancer sticky sessions. This annotation is required if `service.beta.kubernetes.io/do-loadbalancer-sticky-sessions-type` is set. 
+
+### service.beta.kubernetes.io/do-loadbalancer-redirect-http-to-https
+
+Indiciates whether or not http traffic should be redirected to https. Options are `true` or `false`. Defaults to false.
```
fatih: false should be quoted ( i.e: `false`), just like others

docs/example-manifests/README.md
```diff
@@ -0,0 +1,15 @@
+# Example Manifests
+
+The following directory provides examples of static pod, daemonset and systemd manifests
+that were used in a working Kubernetes cluster on DigitalOcean. Use the following examples as a guide
+to setup your cluster correctly using the DigitalOcean Cloud Controller Manager.
+Note that the examples are not meant to be copy and pasted as it contains configuration
+for a specific cluster.
```
fatih: I think this is so they can set up a Kubernetes cluster? If that is the case, do we really need to take care of that aspect as well? Otherwise, I think we only should provide the `cloud-controller-manager.yml` file to the user.

andrewsykim: In general, I agree that it's not needed. But for now just following requirements set by https://github.com/kubernetes/community/blob/master/keps/sig-cloud-provider/0019-cloud-provider-documentation.md#requirement-1-1 which assumes some people may choose to manage control plane themselves and want examples of how to configure them for DigitalOcean.

fatih: Gotcha

fatih:
lgtm

PR #127, opened 2018-09-15
Fix incorrect links to examples

nanzhong:
Looks like the links to these docs were broken when the docs were shuffled around.

cc @andrewsykim


andrewsykim:
lgtm, thanks!

PR #132, opened 2018-10-12
Update golint import path to please the import path checker.

timoreimann:
golint [introduced an import path check](https://github.com/golang/lint/commit/9a272034dedb2a3ed05231d5604ce17fb40f0e58) recently.

Updating to the new path re-enables golint.


bhcleek:
Thanks

PR #134, opened 2018-10-13
Add Go 1.11 to Travis CI.

timoreimann:


andrewsykim:
LGTM, thanks @timoreimann! 

It'll be a while til we compile against go 1.11 (or not?) but I think this is a good test case to catch any issues that might come up :)

Issue #135, opened 2018-10-16
Set droplet tags as kubernetes labels


gianarb:
I think it can be very useful to set up the cloud-controller-manager to set the tags attached to the droplet as labels for the node.

This feature will give me the automation I need to make a more complex cluster topology

andrewsykim:
Starting work on this

timoreimann:
I am closing this one in favor of #136 which has accumulated more context by now. Please see [my comment here](https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/136#issuecomment-466504830) outlining the next steps forward.

Issue #136, opened 2018-10-19
Add label according to node's pool


wnagele:
Currently node's do not have a label that shows which pool they are part of. It would be helpful to have this information in order to schedule resources better.

andrewsykim:
Hey @wnagele, we don't add labels per node pool, but we do have lables for droplet _size_ (see [here](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/docs/controllers/node/labels.md#betakubernetesioinstance-type)). Would that still accomplish what you need?

wnagele:
I envision a case of scaling a cluster up and down to cater for base VS peak load. In this case it seems like a good idea to use different node pools to do so. It would allow to pin resources that have to run to ensure base functionality to a specific node pool. A generic transpose of tags as suggested in #135 into labels would also work as we could tag the droplets accordingly.

mattkelly:
@wnagele thanks for opening this! We're interested in this exact use case for integrating the Cerebral autoscaler - see this related issue: https://github.com/containership/cerebral/issues/52. This is the only thing blocking us from having a fully functional autoscaling integration on DO Kubernetes.

@andrewsykim I did notice that #142 is ready (?) but sitting around going stale. Are you able to provide any insight on an ETA for this going in? Thanks!

andrewsykim:
Hi @mattkelly! We're currently discussing what the best label format should be as it will be difficult to change once it's deployed. Please let me know if you have any opinions on this, right now it's `digitalocean.com/droplet-tag/<tag>`. cc @joonas

normanjoyner:
So this is a bit tricky in my opinion. Labels in Kubernetes have more strict, and generally different, validation than do tags in DigitalOcean. For instance, by default DigitalOcean Kubernetes node pools are tagged with `k8s:${cluster_id}`. This tag would be invalid in Kubernetes since colons are not allowed in the neither the label key, nor the value. Ignoring validation, in my opinion it would be nice if instead of this creating a `digitalocean.com/droplet-tag/k8s:${cluster_id}=""` label, it created `digitalocean.com/droplet-tag/k8s=${cluster_id}` (ignoring the fact that label key should really be more explicitly named such as `digitalocean.com/droplet-tag/cluster-id=${cluster_id}`). Not being able to achieve setting a value on labels feels like we're nerfing the power of Kubernetes.

[...]

PR #142, opened 2018-10-24
Add controller to apply droplet tags as node labels

andrewsykim:
Fixes https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/135 and https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/136


cloud-controller-manager/do/cloud.go
```diff
@@ -96,6 +97,16 @@ func init() {
...
+	clientset := clientBuilder.ClientOrDie("do-shared-informers")
+	sharedInformer := informers.NewSharedInformerFactory(clientset, 0)
+
+	tags := NewTagsController(sharedInformer.Core().V1().Nodes(), clientset, c.client)
+
+	// TODO: pass in stopCh once supported upstream
+	// see https://github.com/kubernetes/kubernetes/pull/70038 for more details
+	sharedInformer.Start(nil)
+	sharedInformer.WaitForCacheSync(nil)
+	go tags.Run()
```
fatih: This is not something I liked. Maybe we should pass a `ctx` and cancell it ourself? Do we have to wait for `stopCh` support ?

cloud-controller-manager/do/tags.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Run starts the tags controller. Tags controllers runs on a loop rather than watching
+// for changes because there is no way to watch for droplet events (yet) and watching
+// node events can be expensive as each node update event (mostly node health checks)
+// would result in an API call to fetch droplets. Caching droplets in memory
+// may be a better solution but this should work for now.
+// TODO: pass in stopch once it's supported upstream
+func (t *tagsController) Run() error {
+	tick := time.Tick(controllerSyncPeriod)
```
fatih: Can you use `time.NewTicker()` and call `ticker.Stop()` in a defer call. This is open door for goroutine leaks.

cloud-controller-manager/do/tags.go
```diff
@@ -0,0 +1,212 @@
...
+			err := t.sync()
+			if err != nil {
+				glog.Errorf("failed to sync droplet tags: %s", err)
+			}
+		}
+	}
+}
+
+func (t *tagsController) sync() error {
+	droplets, err := allDropletList(context.TODO(), t.gclient)
```
fatih: we should pass down `ctx` from top to here, if we implement it and should remove `context.TODO()`.

cloud-controller-manager/do/cloud.go
andrewsykim: `stopCh` support PR is already merged so it will definitely be supported in 1.13. Do you think it's still worth passing a ctx and cancelling it ourselves?

fatih: @andrewsykim it's ok then. Could you create an additional issue so we can track this down later when 1.13 is released.

fatih:
lgtm

andrewsykim:
Will leave it to @fatih @nanzhong!

Despawned:
Hi guys, just pinging again wondering if there's any change we'll get this merged with proper support during the setup process to name/label node pools. Something that's been quite a pressing issue for us, and it seems like this has been ready for a while. 

Thanks!

timoreimann:
First of all, apologies for the delay. We realize how the issues this PR addresses are relevant to a number of users.

We are currently reviewing to what degree tags and labels should be coupled across the different entities they are applied on (droplets, nodes, clusters). The PR partially touches on this subject, which is why we'd first like to gain clarity on the design front before moving forward.

I am hopeful we can come up with the greater picture soon that will allow us to get this PR into a mergeable state. Thanks for your patience, I'll keep you posted on the progress.

Despawned:
Hi,

I'm fairly certain the largest use case for this (and what v0.01 should be) is simply to assign a Kubernetes node label to all nodes in a node group. That's our issue, and a pressing one for that matter, given I just received an email from DO saying that nodes will be recycled due to a security issue. When the nodes are recycled, because we can't assign a Kubernetes label to all nodes in a cluster's node group, we're going to have to type "kubectl label" for every single new node that comes up, which is very unlike a managed service.

timoreimann:
Resolved the merge conflict as a first step.

We hear that label management is a major concern to many DOKS users. We have plans to finish this PR by next week.

normanjoyner:
thanks for the update @timoreimann. I outlined my thoughts on this over on 
https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/136#issuecomment-449193345. Do you have any insight on how this will be approached?

timoreimann:
Hey there @normanjoyner. What you describe in your comment pretty much matches the considerations we have made internally as well. In particular, the part about whether we should consider the (final) colon as a key/value separator or not is somewhat tricky as DO's key-only model may or may not expand in the future. We must be certain to have a path forward should DO tags support real values at some point.

Apologies for the delay; I'm still optimistic we can wrap this up soonish.

normanjoyner:
No need to apologize @timoreimann, I fully understand the dilemma and subsequently the implications of different approaches; it definitely feels like a sticky situation. I’m curious if there are near term are plans to support the node pool ID as a value to a label key (such as `digitalocean.com/node-pool-id for example) either as part of this PR, or separately. In my mind there may be a base set of labels DigitalOcean is interested in applying regardless of user added tags, this potentially being one of them. My specific use case is outlined in [this comment](
https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/136#issuecomment-449088790).

timoreimann:
Hey @normanjoyner and others, I have [left a proposal on #136](https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/136#issuecomment-466504830). Please have a look at let us know what you think.

timoreimann:
Closing this PR in favor of the new course laid out in my comment above.

Issue #145, opened 2018-11-21
Allow setting different protocol for health checks


daimoniac:
We need to be able to set a different protocol for health checks than for the balanced service.

http is being balanced (with ssl termination), but the health check should be a simple tcp one.
but the protocol seems to be hardcoded to the balanced service ones here:
https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/1938ec185300d91f49ed305ecdc65c83acb356c2/cloud-controller-manager/do/loadbalancers.go#L361

![image](https://user-images.githubusercontent.com/7752824/48835337-2c527b80-ed80-11e8-97dc-416aea3264c8.png)

andrewsykim:
Hi @daimoniac! We can add a new service annotation to specify health check protocol. Would that work?

daimoniac:
Perfect!

Am Mi., 21. Nov. 2018, 17:58 hat Andrew Sy Kim <notifications@github.com>
geschrieben:

PR #147, opened 2018-11-21
support overriding load balancer health check protocol

andrewsykim:
Fixes https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/145


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -482,6 +498,20 @@ func getProtocol(service *v1.Service) (string, error) {
 	return protocol, nil
 }
 
+// healthCheckProtocol returns the health check protocol as specifiec in the service
```
```suggestion
// healthCheckProtocol returns the health check protocol as specified in the service
```

:)

Issue #101, opened 2018-05-11
Verify version compatibility with Kubernetes


andrewsykim:
The latest version of DO CCM should support the latest stable Kubernetes and 2 major version before it. At the moment we only advertise that the latest version of CCM supports the latest stable version of Kubernetes because those are the only versions we tested. We should verify that DO CCM is compatible with the latest stable and 2 major versions behind.

timoreimann:
I wondered if I could take a stab on this one.

Presumably, we want to have some kind of automated test? What I am thinking of is to fire up a set of droplets, install Kubernetes (through kubeadm, another tool, or some helper scripts such as https://github.com/caglar10ur/dok8s) with DO's CCM, and finally run some end-to-end tests through it.

Happy to hear your thoughts on my proposal or any other approach.

andrewsykim:
@timoreimann go for it! I had written a simliar thing before to run E2E tests for kube-router https://github.com/cloudnativelabs/kube-router/pull/382 in case it helps.

timoreimann:
@andrewsykim great, I'll see if I can put something together. That kube-router PR looks pretty helpful, thanks for head start!

timoreimann:
Quick update: Been working on this one, will hopefully be able to submit a WIP PR soon. Stay tuned. :)

Issue #45, opened 2017-10-07
E2E Testing 


andrewsykim:
We should start to consider adding an E2E test suite. In terms of tooling I personally don't have a preference but using a combination fo kubeadm/terraform would make the most sense to me.

xmudrii:
On the unrelated note, does anybody know how to use cloud-controller with terraform provider?

I believe I need to insert `cloud-provider` flag for `kubelet`, but adding it to `ExecStart` [here](https://github.com/kubernetes-digitalocean-terraform/kubernetes-digitalocean-terraform/blob/master/01-master.yaml#L23
) doesn't seem to work.

Am I missing something?

andrewsykim:
I did some initial work in https://github.com/kubernetes-digitalocean-terraform/kubernetes-digitalocean-terraform/pull/39 for cloud controller integration but it needs to be updated to work with v1.8 which is pending review for https://github.com/kubernetes-digitalocean-terraform/kubernetes-digitalocean-terraform/pull/47.

bhcleek:
@xmudrii Here's a patch to https://github.com/kubernetes-digitalocean-terraform/kubernetes-digitalocean-terraform that works for me. It includes https://github.com/kubernetes-digitalocean-terraform/kubernetes-digitalocean-terraform#39 and https://github.com/kubernetes-digitalocean-terraform/kubernetes-digitalocean-terraform#47 plus one other change (the release file to use).

bhcleek:
@xmudrii Are you planning to work on this issue? I was planning to jump in it next, but don't want to duplicate work if you're already working on it.

xmudrii:
@bhcleek Thanks for the answer. No not for now, feel free to take it. :)

bhcleek:
I forgot to paste the patch. Here you go:
```
diff --git a/01-master.yaml b/01-master.yaml
index 97cf2f5..2bbce16 100644
--- a/01-master.yaml
+++ b/01-master.yaml
@@ -36,12 +36,14 @@ write_files:
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
           --anonymous-auth=false \
           --client-ca-file=/etc/kubernetes/ssl/ca.pem \
-          --api-servers=http://127.0.0.1:8080 \
+          --kubeconfig=/etc/kubernetes/kubelet-kubeconfig.yaml \
+          --require-kubeconfig \
           --network-plugin-dir=/etc/kubernetes/cni/net.d \
[...]

PR #148, opened 2018-11-26
Add end-to-end test verifying Kubernetes compatibility.

timoreimann:
This change adds an end-to-end test verifying that the CCM core functionalities (namely, node and service controllers) work as intended.

In a nutshell, it basically does the following for each Kubernetes version supported:

- dynamically create a cluster via kops
- wait for the cluster to become ready
- wait for the nodes to be fully configured
- deploy an example pod
- expose the pod through a DO-managed LB
- send a sample request over the LB
- clean up all objects that were created along the way (LB, cluster, state store)

The tests should work both when invoked from the local machine as well as on Travis CI. The latter requires some extra tuning to the (Travis CI) configuration which I have done on [this development branch of mine](https://github.com/timoreimann/digitalocean-cloud-controller-manager/tree/verify-version-compat) resulting in [this build output](https://travis-ci.org/timoreimann/digitalocean-cloud-controller-manager/builds/459761520). I tried to explain all the needed environment variables in a separate README file.

The PR grew bigger than I initially thought it'd need to; a fair chunk of that resulted from the fact that cleaning up state manually during repeated test runs became tedious at some point, so I automated things as much as I could. The extra dependency on minio is a consequence of that so that I could talk to the Spaces/S3 API accordingly. I'm open to suggestions to reduce the overall code size.

There are a few minor tasks left to do if we wanted to integrate the test properly:

1. pick some DigitalOcean account to set up a token in and push an SSH key to
1. possibly make the CCM version configurable (and maybe even build it right before the test is invoked)
1. (if we want to allow running end-to-end tests from multiple builds concurrently:) include a (sufficiently) unique identifier in the kops state name, e.g., the PR ID or branch name

I didn't want to go too far (which I might have done already ;) ) though prior to getting some initial agreement on the direction taken. I'd be more than happy to close the final code gap in a follow-up PR if desirable.

Some final notes:

- For the longest time, I tested everything in the AMS3 region, which turned out to be fairly unreliable (machines not getting set up, load balancers not finishing, cluster cleanup failing, etc). I got a [hint](https://kubernetes.slack.com/archives/C3QUFP0QM/p1542892880619600?thread_ts=1542874467.611500&cid=C3QUFP0QM) on the `kops-users` Kubernetes Slack channel to switch to SFO2 which immediately made all my builds go green consistently. The person I talked to mentioned that the NYC regions seemed flaky to; I'm not sure if there's an issue with the other regions or if I (we) have been using them wrong, but I wanted to mention it.
- I also have a variant of the implementation ready that parallelizes testing against the supported Kubernetes, which speeds up overall test execution quite significantly. The downsides are that it increases code complexity, complicates following test output, and aggravates debugging, which is why I did not include it in this PR. We are safely within the Travis execution deadline of 50 minutes with the sequential approach, so I'd only add the parallelized code if requested.

The PR references / fixes possibly a combination of #101 and #45.

I'm very keen on hearing about what you think -- happy to discuss and adapt as needed!


andrewsykim:
This is awesome! Thanks for putting this together, will take a look soon!

e2e/e2e_test.go
```diff
@@ -0,0 +1,301 @@
...
+				t.Fatalf("failed to get working directory: %s", err)
+			}
+			kubeConfFile := path.Join(wd, "kubeconfig-e2e."+dnsName)
+
+			// Delete old kubeconfig
+			if err := os.Remove(kubeConfFile); err != nil && !os.IsNotExist(err) {
+				t.Fatalf("failed to delete kubeconfig %q: %s", kubeConfFile, err)
+			}
+
+			// Create space.
```
andrewsykim: I would prefer to keep all the cluster setup steps to be done via bash or at least in a separate go function to separate where we do setup vs testing. what do you think?

andrewsykim: Also, might be easier to just assume a bucket is created and always use that bucket, kops should handle subpath creation from there.

e2e/e2e_test.go
```diff
@@ -0,0 +1,301 @@
...
+					}
+				}
+				return false, nil
+			}); err != nil {
+				t.Fatalf("failed to observe ready example pod %q in time: %s\npod: %v", appName, err, appPod)
+			}
+			l.Printf("Took %v for pod to become ready\n", time.Since(start))
+
+			// Create service object.
+			svcName := "svc"
```
andrewsykim: Probably good to append these names with some random hash to avoid conflicts in future runs (in case they're not cleaned up in previous runs).

e2e/e2e_test.go
```diff
@@ -0,0 +1,301 @@
...
+				return numReadyNodes == numWantNodes, nil
+			}); err != nil {
+				t.Fatalf("got %d ready node(s), want %d: %s\nnnodes: %v", numReadyNodes, numWantNodes, err, spew.Sdump(gotNodes))
+			}
+			l.Printf("Took %v for nodes to become ready\n", time.Since(start))
+
+			// Check that load balancer is working.
+
+			// Install example pod to load-balance to.
+			appName := "app"
```
andrewsykim: Probably good to append these names with some random hash to avoid conflicts in future runs (in case they're not cleaned up in previous runs).

e2e/e2e_test.go
```diff
@@ -0,0 +1,301 @@
...
+			// Wait for service IP address to be assigned.
+			l.Println("Polling for service load balancer IP address assignment")
+			start = time.Now()
+			var lbAddr string
+			if err := wait.Poll(5*time.Second, 10*time.Minute, func() (bool, error) {
+				svc, err := cs.CoreV1().Services(corev1.NamespaceDefault).Get(svcName, metav1.GetOptions{})
+				if err != nil {
+					return false, err
+				}
+				for _, ing := range svc.Status.LoadBalancer.Ingress {
```
andrewsykim: I think this for loop here is a bit confusing cause we return on first index. Should just check for length and use `svc.Status.LoadBalancer.Ingress[0].IP`.

e2e/e2e_test.go
andrewsykim: ah, seems like we did this here so we can set the version on the test table?

e2e/e2e_test.go
```diff
@@ -0,0 +1,301 @@
...
+								numReadyNodes++
+							}
+						}
+					}
+				}
+
+				l.Printf("Found %d/%d ready node(s)", numReadyNodes, numWantNodes)
+				return numReadyNodes == numWantNodes, nil
+			}); err != nil {
+				t.Fatalf("got %d ready node(s), want %d: %s\nnnodes: %v", numReadyNodes, numWantNodes, err, spew.Sdump(gotNodes))
```
andrewsykim: t.Error I think

e2e/e2e_test.go
```diff
@@ -0,0 +1,301 @@
...
+				}
+				appPod = pod
+				for _, cond := range appPod.Status.Conditions {
+					if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
+						return true, nil
+					}
+				}
+				return false, nil
+			}); err != nil {
+				t.Fatalf("failed to observe ready example pod %q in time: %s\npod: %v", appName, err, appPod)
```
andrewsykim: Prefer t.Errorf since it's a test case failure not a setup failure.

e2e/e2e_test.go
```diff
@@ -0,0 +1,301 @@
...
+				nl, err := cs.Core().Nodes().List(metav1.ListOptions{LabelSelector: "kubernetes.io/role=node"})
+				if err != nil {
+					return false, err
+				}
+
+				gotNodes = nl.Items
+				numReadyNodes = 0
+				for _, node := range gotNodes {
+					for _, cond := range node.Status.Conditions {
+						if cond.Type == corev1.NodeReady && cond.Status == corev1.ConditionTrue {
```
andrewsykim: I think it's really important here that we also verify that there are no taints on the node to test registration, particularly the cloud taint `node.cloudprovider.kubernetes.io/uninitialized` should not exist.

e2e/e2e_test.go
andrewsykim: Prefer t.Errorf since it's a test case failure not a setup failure.

e2e/e2e_test.go
```diff
@@ -0,0 +1,301 @@
...
+					return false, err
+				}
+				for _, ing := range svc.Status.LoadBalancer.Ingress {
+					lbAddr = ing.IP
+					return true, nil
+				}
+
+				return false, nil
+			}); err != nil {
+				t.Fatalf("failed to observe load balancer IP address assignment: %s", err)
```
andrewsykim: Prefer t.Errorf since it's a test case failure not a setup failure.

e2e/e2e_test.go
```diff
@@ -0,0 +1,301 @@
...
+				defer resp.Body.Close()
+
+				lastStatusCode = resp.StatusCode
+				if resp.StatusCode != http.StatusOK {
+					return false, nil
+				}
+
+				return true, nil
+			}); err != nil {
+				t.Fatalf("failed to send request over LB to example application: %s (last status code: %d / attempts: %d)", err, lastStatusCode, attempts)
```
andrewsykim: Prefer t.Errorf since it's a test case failure not a setup failure.

e2e/e2e_test.go
timoreimann: >ah, seems like we did this here so we can set the version on the test table?

Yes, exactly: we couldn't do this in bash alone as we need to parameterize kops for at least the Kubernetes version and the state store URL which are currently driven by the Go test.
One alternative approach might be to let (a smarter version of) the bash script be in charge and call specific (sub-)tests with the right parameters. Personally though, I'd rather have more logic in Go than in bash as it's easier to understand and modify that way.
Does that address your comments?

I'll follow your advice for sure to group the test setup and test execution into separate functions for better readability. 👍

e2e/e2e_test.go
timoreimann: I pushed a commit that makes sure the taint you mentioned isn't set. Let me know if you want me to  broaden it further to "no taints at all must be set".

e2e/e2e_test.go
timoreimann: Right now, we remove the entire cluster after each Kubernetes version-specific sub-test run and create a brand new cluster for each sub-test, so it shouldn't be possible to run into a conflict situation. Or did I miss something?

e2e/e2e_test.go
timoreimann: With sub-tests, `t.Fatalf` just causes the corresponding sub-test to terminate, not the entire (parent) test. So `t.Fatalf` seems appropriate to me given that we don't want to move forward with submitting the HTTP request if the sample pod fails to turn ready. Does that make sense?

e2e/e2e_test.go
timoreimann: Similar to pods, we should never collide as each test cluster is brought up for a single test case only.

e2e/e2e_test.go
timoreimann: Sure, makes sense -- done!

e2e/e2e_test.go
timoreimann: See my response above -- `t.Fatalf` within sub-tests should work as expected.

e2e/e2e_test.go
timoreimann: Same as above. :)

e2e/e2e_test.go
timoreimann: I should also mention that I started off my PR with a static bucket but moved away from it for two reasons:

1. I ran into at least one scenario where not having a clean state to start from caused trouble: I once tried to change the number of worker nodes, which from kops' perspective seems to be a change to an existing cluster. Unfortunately, kops does not support downsizing an k8s cluster in DO, so my test failed even though it was entirely unrelated to the system-under-test. Nobody will likely do this for the test once it's solid, but it showed me that relying on in-place adjustments is a different mode of operation, so I decided to get rid of it altogether by having separate clusters per test (and cleaning up accordingly).
1. Dynamic buckets that can be removed after test completion means no further cost for the user that runs an end-to-end test and forgets to clean up manually.

If we want a static bucket for simplicity reasons, I'm okay to make the change. Just wanted to make sure the trade-offs are clear before making a decision.

e2e/e2e_test.go
andrewsykim: No I think it's fine then :)

e2e/e2e_test.go
andrewsykim: Yes! Good point!

e2e/e2e_test.go
```diff
@@ -0,0 +1,311 @@
...
+				if err != nil {
+					return false, err
+				}
+
+				gotNodes = nl.Items
+				numReadyNodes = 0
+				for _, node := range gotNodes {
+					// Make sure the "uninitialized" node taint is missing.
+					for _, taint := range node.Spec.Taints {
+						if taint.Key == unitializedNodeTaint {
```
andrewsykim: where is `unitializedNodeTaint` set? I don't see it anywhere 🤔

andrewsykim: there's a variable for this upstream btw, might be best to use that.

e2e/e2e_test.go
timoreimann: Just pushed a few minor improvements / fixes along with a change that moves the setup part into a dedicated function.

One extra complication with the latter is that we now need the caller to (defer-)invoke the cleanup functions defined during the setup (e.g., store state removal). It takes a bit of iterating and error aggregation to make it happen but I personally think it's worth it, and the pattern is quite common.

Happy to hear your thoughts, and ready to make any further adjustments we may find relevant. :)

andrewsykim:
@timoreimann this looks great and I think it's ready to merge. Can you squash the commits into two, one for vendor and one for everything else. Thanks!

timoreimann:
@andrewsykim squashing done. :)

timoreimann:
@andrewsykim Once merged, it will be possible to execute the test manually. Any thoughts on whether we should hook it up to the CI pipeline in general? I noted down a few possible follow-up tasks in the PR description (_"There are a few minor tasks left to do [...]"_). If you'd like some of them (or some other, not mentioned) to be implemented, I'd be happy to follow up.

Some means of integration might be good IMHO to guarantee that the test doesn't rot. If we find calling the test on every PR to be too heavy, we could perhaps invoke it automatically before a release?

Let me know in case you'd rather continue the discussion in a separate issue. :)

timoreimann:
Just found [/docs/testing.md](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/5ef072a20baa1ec3ceed1ea18be6452928eebfee/docs/testing.md). Do we need to update that too somehow?

andrewsykim:
@timoreimann yes please update testing.md, but feel free to update in a separate PR. I need to put more thought into where in CI we should run these tests. Separate issue sounds great!

Issue #153, opened 2018-12-06
Better state handling of LBs


andrewsykim:
We should add smarter logic for ensuring load balancers. Right now we check for the "active" state when we initially create the LB, but on follow up reconciliation of LBs, we only check for existence. This makes it difficult to handle cases where an LB may be created properly but is in a bad state.

timoreimann:
Closing this one as I think we haven't run into a case of LBs going into an error state for a reasonable time period.

PR #154, opened 2018-12-07
Delete load balancers in error state and when IP is not assigned

andrewsykim:
Delete a load balancer if after a certain timeout period it was not successfully assigned a load balancer IP. It's important to not delete LBs with an already assigned IP because trying to recreate it would mean losing the LB IP. For LBs in errored state but with an IP assigned, we ignore it and the user should manually recreate the service. 

Fixes https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/153

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -80,9 +80,11 @@ const (
 	// should be redirected to Https. Defaults to false
 	annDORedirectHTTPToHTTPS = "service.beta.kubernetes.io/do-loadbalancer-redirect-http-to-https"
 
-	// defaultActiveTimeout is the number of seconds to wait for a load balancer to
-	// reach the active state.
-	defaultActiveTimeout = 90
+	// defaultActiveTimeout is the number of seconds to wait for a load balancer to reach active state.
+	// if a load balancer is not assigned an IP within this timeout, it will be deleted and recreated.
+	// A default timeout of 120s seems reasonable given load balancers are typically provisioned in less than
+	// half the amount of time.
+	defaultActiveTimeout = 120
 
 	// defaultActiveCheckTick is the number of seconds between load balancer
 	// status checks when waiting for activation.
@@ -95,6 +97,15 @@ const (
 )
 
 var errLBNotFound = errors.New("loadbalancer not found")
+
+type errLBTimeout struct {
+	msg  string
+	lbID string
+}
+
+func (e *errLBTimeout) Error() string {
+	return fmt.Sprintf("error waiting for load balancer %q to be active: %v", e.lbID, e.msg)
+}
 
 type loadbalancers struct {
 	client            *godo.Client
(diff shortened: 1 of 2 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -82,7 +82,7 @@ const (
 
 	// defaultActiveTimeout is the number of seconds to wait for a load balancer to
 	// reach the active state.
-	defaultActiveTimeout = 90
+	defaultActiveTimeout = 120
```
andrewsykim: To add a bit of a buffer between when we'll choose to delete an LB.

timoreimann: It might make sense to include the buffer rationale into the GoDoc so that it's visible beyond this PR.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -96,6 +96,15 @@ const (
 
 var errLBNotFound = errors.New("loadbalancer not found")
 
+type errLBTimeout struct {
+	err  string
```
timoreimann: `err` is conventionally used for `error` types. Might be a good idea to rename it to `message` or change the type to `error`.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -144,10 +157,36 @@ func (l *loadbalancers) GetLoadBalancer(ctx context.Context, clusterName string,
...
 		return nil, err
 	}
 
+	if _, ok := err.(*errLBTimeout); ok && exists {
+		// we should attempt to delete load balancers where state is not active after the provided timeout
+		// and if it has not been given an IP yet. Load balancers in an error state that already have an IP address
+		// should be manually recreated.
+
+		// we should only delete this load balancer, if a previous IP was not assigned to it.
+		// this ensures load balancerse with an already assigned IP aren't accidentally deleted.
```
```suggestion
		// this ensures load balancers with an already assigned IP aren't accidentally deleted.
```

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -144,10 +157,36 @@ func (l *loadbalancers) GetLoadBalancer(ctx context.Context, clusterName string,
...
+		if len(service.Status.LoadBalancer.Ingress) > 0 {
+			return nil, err
```
timoreimann: How about annotating the error to indicate to the user that manual intervention is required? Something like

```
return nil, fmt.Errorf("cannot reconcile load-balancer in error state having IP address assigned already, needs to be recreated manually: %s", err)
```

andrewsykim: good idea

andrewsykim:
@timoreimann always welcome reviews from you!

andrewsykim:
@timoreimann updated, PTAL when you have a chance :)

timoreimann:
I'm closing this PR for now -- turns out we may want to address the affected problem in a different way involving internal DO components only. Will reopen the PR should we decide otherwise.

Thanks for the PR, Andrew. 👍

PR #155, opened 2018-12-07
add @timoreimann as a maintainer

andrewsykim:
cc @timoreimann

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Requirements:
 
 * Go: min `v1.10.x`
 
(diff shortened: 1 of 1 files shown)


fatih:
I think we should remove that section, Github already has a contributor tab and the people maintaining might change in the future again. Not saying @timoreimann shouldn't be on the list of course, just my 2 cents :)

andrewsykim:
Noted! I'm gonna merge this for now but I'll follow up :)

PR #170, opened 2019-02-13
Tag created load balancers with existing cluster ID.

timoreimann:
This enables a consistent view on all DO resources managed by the cluster.

As a drive-by change, some struct initializations were changed to use named fields in order to be more robust towards (optional) field additions in the future.

digitalocean/godo had to be updated in order to access load balancer tags.

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -24,7 +24,7 @@ import (
 	"strings"
 	"time"
 
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/core/v1"
 	"k8s.io/kubernetes/pkg/cloudprovider"
 
 	"github.com/digitalocean/godo"
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -24,7 +24,7 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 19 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -24,7 +24,7 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
nathanjsweet: unnecessary label

cloud-controller-manager/do/loadbalancers_test.go
```diff
@@ -25,7 +25,7 @@ import (
 
 	"github.com/digitalocean/godo"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
nathanjsweet: unnecessary package label

timoreimann:
@nathanjsweet grrr, vscode started to add those sometimes for reasons unknown to me. Gotta fix this some time.

Pushed another commit to remove the aliases. PTAL again.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -319,13 +326,19 @@ func (l *loadBalancers) buildLoadBalancerRequest(service *v1.Service, nodes []*v
 
 	redirectHTTPToHTTPS := getRedirectHTTPToHTTPS(service)
 
+	var tags []string
+	if l.clusterID != "" {
+		tags = []string{l.clusterID}
```
bouk: The tag should have the k8s: prefix

PR #180, opened 2019-02-25
Makefile: Fix check-headers target and header violations.

timoreimann:
The target would previously not work (because the wildcard in the command did not apply).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,5 +102,7 @@ test:
 
 .PHONY: check-headers
 check-headers:
-	@./ci/headers-*.sh
+	@./ci/headers-bash.sh
+	@./ci/headers-docker.sh
+	@./ci/headers-go.sh
 
--- a/cloud-controller-manager/cmd/digitalocean-cloud-controller-manager/Dockerfile
+++ b/cloud-controller-manager/cmd/digitalocean-cloud-controller-manager/Dockerfile
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#    http://www.apache.org/licenses/LICENSE-2.0
+#     http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
--- a/cloud-controller-manager/do/zones_test.go
+++ b/cloud-controller-manager/do/zones_test.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2017 DigitalOcean
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -101,6 +101,9 @@ test:
 	@go test $(shell go list ./... | grep -v vendor)
 
 .PHONY: check-headers
-check-headers:
-	@./ci/headers-*.sh
+check-headers: ci/headers-*.sh
+
+.PHONY: ci/headers-*.sh
+ci/headers-*.sh:
+	@./$@
```
bouk: I think just changing it to:

```
check-headers:
  ./ci/headers-bash.sh
  ./ci/headers-docker.sh
  ./ci/headers-go.sh
```

Would be best. Removing the `@` also makes it show which files fail in CI.

timoreimann: I had it like that initially but then changed it into the more generic approach so that adding a new headers check would require no modifications to the Makefile. I find that a tad more convenient, WDYT?

I think the `@` is only suppressing the output of the line being executed, not the command's STDOUT itself. For instance, when I run the rule with a Dockerfile missing its header, I get the following:

```
$ make check-headers
license headers not found in ./cloud-controller-manager/cmd/digitalocean-cloud-controller-manager/Dockerfile
make: *** [ci/headers-docker.sh] Error 1
```

bouk: I think clarity trumps convenience in this case, it would be better to clearly show what’s going on. Trying to be clever with convenience is what broke this in the first case :)

TIL what @ actually does in Make...

bouk: I think clarity trumps convenience in this case, it would be better to clearly show what’s going on. Trying to be clever with convenience is what broke this in the first case :)

TIL what @ actually does in Make...

timoreimann: @bouk convinced and updated. Extra benefit that the Makefile now doesn't stop on the first header type violation but executes them all. 👍 

PTAL again.

Makefile
```diff
@@ -102,5 +102,7 @@ test:
 
 .PHONY: check-headers
 check-headers:
-	@./ci/headers-*.sh
+	@./ci/headers-bash.sh; \
+	./ci/headers-docker.sh; \
+	./ci/headers-go.sh; \
```
bouk: Huh, why did you do it like this with the semicolon and backslash instead of just newlines? Won’t the semicolon cause the exit codes to be ignored?

nanzhong: I thinks it's for 
https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/180#discussion_r259825592?

nanzhong: Though, does that mean only the exit code of the last `headers-go.sh` check is going to be respected? 🤔 Wow, I really need to learn more about make...

timoreimann: @nanzhong good catch, you're right. I played it safe now and went with @bouk's initial suggestion.

PR #181, opened 2019-02-25
Reconcile cluster ID tags on DO load-balancer resources

timoreimann:
This ensures that cluster IDs exist on load-balancers which either missed them because previous versions of CCM did not add them, or because users removed them through a (yet to be implemented) deletion method of the DO API.


cloud-controller-manager/do/resources.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if len(res) == 0 {
+		return nil
+	}
+
+	// Tag collected resources with the cluster ID. If the tag does not exist
+	// (for reasons outlined below), we will create it and retry tagging again.
+	for {
+		ctx, cancel = context.WithTimeout(context.Background(), requestTimeout)
+		defer cancel()
```
nanzhong: Not sure about this myself, but should this `context.WithTimeout` and `defer cancel()` be outside of this loop?

I'm unsure what the desired semantics here are. Do we want to have this loop in `sync()` run until success or timeout, or do we want each individual invocation of `sync()` to have a timeout?

timoreimann: My primary intention was to bound the time spent with the requests _somehow_ so that a single sync run cannot starve subsequent ones.

Doing the timeout / cancel before vs. inside the loop defines whether we want to bound sync-overall vs. per-loop. Practically, I don't think it makes too much of a difference since the loop executes at most three requests over its lifetime span of up to two iterations (1st loop: 404 on the first tag request followed by tag creation; 2nd loop: following successful tag request). So each sync run will take ~30 seconds max (3 * 10 sec timeout plus neglectable CPU time).
Of course, that's only if the dependent APIs behave as expected. I guess we could argue that a sync-overall timeout would provide protection even for pathological cases. Not sure how realistic that might be, however.

Happy to discuss.

nanzhong: Yup, just wanted to clarify the intention here. I'm inclined to suggest me move the cancel outside of the loop and rely on sync overall timeout because it means that if new LBs are created between syncs the new ones would be picked up in the next sync call.

cloud-controller-manager/do/cloud.go
```diff
@@ -99,6 +103,21 @@ func init() {
 }
 
 func (c *cloud) Initialize(clientBuilder controller.ControllerClientBuilder) {
+	if c.clusterID == "" {
+		glog.Info("No cluster ID configured -- skipping resource controller initialization.")
```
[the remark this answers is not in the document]
timoreimann: @nanzhong I pushed another commit, PTAL.

cloud-controller-manager/do/resources.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// Collect tag resources for known load balancer names (i.e., services
+	// with type=LoadBalancer.)
+	svcs, err := r.svcLister.List(labels.Everything())
+	if err != err {
+		return fmt.Errorf("failed to list services: %s", err)
+	}
+	var res []godo.Resource
+	for _, lb := range lbs {
+		for _, svc := range svcs {
```
bouk: This loop has some nice quadratic performance, you should use a map here to make it O(N) instead of O(MN)

cloud-controller-manager/do/resources.go
```diff
@@ -0,0 +1,142 @@
...
+			if svc.Spec.Type == corev1.ServiceTypeLoadBalancer && cloudprovider.GetLoadBalancerName(svc) == lb.Name {
+				res = append(res, godo.Resource{
+					ID:   lb.ID,
+					Type: godo.ResourceType(resourceTypeLoadBalancer),
```
bouk: You should make a PR for godo to add Load Balancer resource type

cloud-controller-manager/do/resources.go
bouk: Yeah this loop is a bit funky, can be structured in a nicer way I think.

timoreimann: Tried a different approach in https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/184, PTAL.

cloud-controller-manager/do/resources.go
timoreimann: I think it'd still be O(MN) even if we kept track of visited services as we go through the elements. (Practically, it will be faster because the inner loop needs to run through all services only once per outer loop.)

In order to make this truly O(N), I believe we'd have to manage a map that is modified as service objects join and leave in the cluster. However, that'd require extra synchronization and more testing.

I'm not sure if it's worth the effort though given that practically speaking, quadratic time probably won't bother us before a few other bottlenecks. WDYT?

cloud-controller-manager/do/resources.go
timoreimann: Yep, that's on the TODO list. 👍

cloud-controller-manager/do/resources.go
bouk: https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/185

cloud-controller-manager/do/resources.go
timoreimann: https://github.com/digitalocean/godo/pull/206

PR #184, opened 2019-02-26
Unwind resource controller tag loop.

timoreimann:
Improves readability.

### DIFF
--- a/cloud-controller-manager/do/resources.go
+++ b/cloud-controller-manager/do/resources.go
@@ -148,13 +148,10 @@ func (r *ResourcesController) tagResources(res []godo.Resource) error {
 	resp, err := r.gclient.Tags.TagResources(ctx, tag, &godo.TagResourcesRequest{
 		Resources: res,
 	})
-	if err == nil {
-		return nil
-	}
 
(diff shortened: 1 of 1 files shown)


cloud-controller-manager/do/resources.go
```diff
@@ -136,7 +128,33 @@ func (r *ResourcesController) sync() error {
...
 			return fmt.Errorf("failed to create tag %q: %s", r.clusterIDTag, err)
 		}
+
+		// Try tagging again, which should not fail anymore due to a missing
+		// tag.
+		err = r.tagResources(res)
+	}
+
+	if err != nil {
+		return fmt.Errorf("failed to tag LB resource(s) %v with tag %q: %s", res, r.clusterIDTag, err)
```
bouk: Won't this cause the error to be `failed to tag LB resource(s) %v with tag %q: %s: failed to tag LB resource(s) %v with tag %q: %s` if tagResources fails?

timoreimann: Oh yes it would. I pushed another commit to address that.

PR #186, opened 2019-02-26
loadbalancers: don't use pointer to loop variable in load balancers map

bouk:
`loadBalancers[lb.Name] = &lb` was getting a pointer to the loop variable, not to the value inside the slice (tricky Go escape semantics...). This meant that every value in the map would point to the same value, which would be the load balancer returned last from the API. So if there were multiple load balancers only the last one would get the cluster tag applied.

I changed it to have the IDs in the map instead of the load balancer pointers. I also added a test that actually verifies that the correct API requests are made.

### DIFF
--- a/cloud-controller-manager/do/resources_test.go
+++ b/cloud-controller-manager/do/resources_test.go
@@ -152,11 +152,11 @@ func TestTagsSync(t *testing.T) {
 					Resources: []godo.Resource{
 						{
 							ID:   "1",
-							Type: "load_balancer",
+							Type: resourceTypeLoadBalancer,
 						},
 						{
 							ID:   "2",
-							Type: "load_balancer",
+							Type: resourceTypeLoadBalancer,
 						},
 					},
 				},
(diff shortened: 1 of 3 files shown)


cloud-controller-manager/do/resources_test.go
```diff
@@ -125,6 +128,40 @@ func TestTagsSync(t *testing.T) {
...
+					},
+				}, newFakeOKResponse(), nil
+			},
+			tagSvc: newFakeTagsService(clusterIDTag),
+			tagRequests: []*godo.TagResourcesRequest{
+				{
+					Resources: []godo.Resource{
+						{
+							ID:   "1",
+							Type: "load_balancer",
```
timoreimann: We can use the `resourceTypeLoadBalancer` variable here...

cloud-controller-manager/do/resources_test.go
```diff
@@ -125,6 +128,40 @@ func TestTagsSync(t *testing.T) {
...
+						},
+						{
+							ID:   "2",
+							Type: "load_balancer",
```
timoreimann: ...and here.

PR #188, opened 2019-02-26
loadbalancers: deflake test for multiple tags.

timoreimann:
The ordering of the collected tag request resources depends on the ordering of the services which are retrieved from the informer cache. However, the informer's List() is not deterministic (under the hood, it iterates over a map).

Thus, we sort the resources prior to asserting.


timoreimann:
Verified by running `go test -count 150 -run TestTagsSync ./cloud-controller-manager/do/...` successfully.

PR #190, opened 2019-02-27
Travis: replace Go tip by 1.12.

timoreimann:
Tip [keeps failing](https://travis-ci.org/digitalocean/digitalocean-cloud-controller-manager/jobs/499524349) for reasons we should not care about at this stage while Go 1.12 has been released very recently.


timoreimann:
@bouk WDYT?

PR #200, opened 2019-03-23
Cache cloud resources to address node update race conditions

nanzhong:
Currently droplet information is fetched in an unoptimized fashion where we loop over all the droplets in the account (via api pagination) to find the single droplet ware interested in. This process can take a while for users who have a large number of droplets in their account.

This delay causes us to run into 2 race conditions.

1. The forced sync delay in the shared informer we use currently uses the default delay for forced syncs. This delay is not long enough for accounts that have O(100s) droplets to be processed. This causes contexts to get cancelled before any progress updating nodes can even happen.
2. Even if that delay is bumped up, the amount of time between when kubernetes' node controller fetches the node and when it tries to update the node based on CCM provided information is long enough cause update conflicts that trigger and endless cycle of attempts.

This PR introduces a cloud resource manager with an cached api implementation that fetches all the droplet and LB information periodically as needed so that when node updates need to happen all the lookups can happen in memory. 

I've verified that this works on accounts with ~500 droplets.


cloud-controller-manager/do/common.go
```diff
@@ -20,12 +20,12 @@ import (
 	"context"
 	"fmt"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
nanzhong: 🤔 not sure why my editor keeps doing this...

cloud-controller-manager/do/common.go
```diff
@@ -20,12 +20,12 @@ import (
...
 
 	"github.com/digitalocean/godo"
 )
 
-const apiPerPage = 100
+const apiPerPage = 200
```
nanzhong: 200 is the max page size for our api, might as well reduce # of round trips.

cloud-controller-manager/do/cloud.go
```diff
@@ -151,3 +158,105 @@ func (c *cloud) ScrubDNS(nameservers, searches []string) (nsOut, srchOut []strin
...
+	Reload(ctx context.Context) error
+}
+
+type memResources struct {
+	dropletIDMap   map[int]*godo.Droplet
+	dropletNameMap map[string]*godo.Droplet
+	mutex          sync.RWMutex
+}
+
+func (c *memResources) Reload(ctx context.Context) error {
```
jcodybaker: nit: `c *memResource` should probably be `m *memResource`

cloud-controller-manager/do/cloud.go
```diff
@@ -151,3 +158,105 @@ func (c *cloud) ScrubDNS(nameservers, searches []string) (nsOut, srchOut []strin
...
+	return &cachedAPI{
+		memResources: &memResources{},
+
+		client: client,
+		ttl:    ttl,
+		now:    time.Now,
+	}
+}
+
+func (c *cachedAPI) Reload(ctx context.Context) error {
```
timoreimann: I see a few concerns with the way `Reload` needs to be integrated and how it is implemented:

- due to the lock hold, `Reload` will block concurrent callers until the (potentially long) operation completes. This could lead to delays and maybe even timeouts on the caller end.
- The cost of reloading are amortized across all calls, which means that some consumers will pay more than others. This may lead to quality degradation scenarios that are harder to identify.
- A failed reload causes the entire operation to fail even in the case where we may have cached (albeit somewhat stale) droplet data that we could return.
- We have to build in logic that determines whether it's time to reload and calculate the future reload point in time.
- We have to make sure to invoke the `Reload` side-effect on every method that wants to look up droplets, which is something that could be forgotten if we ever add more methods or refactor the code.

Some of these shortcomings can likely be avoided by improving the current implementation. However, I want to propose an alternative design that may be better suited systematically: let's outsource droplet management into a dedicated goroutine responsible for reloading and serving lookups. The key benefits would be that it could run at a static periodicity, return droplet data via a (inherently synchronized) channel, require an (internal) lock only when swapping out old data via new one, decouple reload from lookup cost, and keep returning stale data until the reload completes.

WDYT?

nanzhong: Agree with many of your concerns. I steered away from doing a periodic reload on a separate goroutine for 2 reasons:

1. This could just be a conceptual difference in how we are viewing a sync loop in a separate goroutine. Both approaches achieve the goal of invalidating/refreshing a cache, but conceptually the sync loop to me feels like we are trying to sync state between the DigitalOcean api and ccm. I feel like the cloud provider interface lends itself better to on demand querying a cloud's non-kubernetes api (the caching in the current approach only serving as a short-circuit if fresh enough). Again, I feel like this is just a conceptual difference, not really a material one.

2. There is rate limit at play on the public api token that ccm uses, so minimizing the number of requests made would be ideal. A sync loop that continually runs would use up a portion of the alloted requests. Not a big deal, just something that came to mind.

Like you said, there are quite a few things we can do to improve the current implementation to address some of your concerns, but I'm happy to swtich over to a sync loop for managing the cache as well. Let me know if you would still prefer that.

timoreimann: With regards to 1: The cloud provider interface includes a method `Initialize` for the dedicated purpose of initializing asynchronous activities. [From the in-code comment](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/68f5e06565659dcd45c3c608dd05d542ad111e66/vendor/k8s.io/kubernetes/pkg/cloudprovider/cloud.go#L33-L35):

>	// Initialize provides the cloud with a kubernetes client builder and may spawn goroutines
	// to perform housekeeping activities within the cloud provider.

So I believe it'd be okay to implement a cloud/CCM synchronization for performance reasons from a conceptual point of view.

Fair point about the rate limiting aspect, I missed to take that into account. Looks like the DigitalOcean API currently [permits 5,000 requests per hour and token](https://developers.digitalocean.com/documentation/v2/#rate-limit). With a static update frequency of once per minute (as currently configured), that's 60 extra hourly requests or 0.83% of the overall quota. I can see us possibly doubling the frequency to twice per minute, so overall that'd be a good precent overhead.

Whether that's acceptable or not likely depends on a user's API usage pattern. What I'd think is that clusters which require many updates would also frequently reload with the current implementation anyway while users with infrequent reloads may not be bothered much by the few extra reloads. Users sharing tokens across many (high-volume) clusters are likely better off using dedicated API tokens per cluster for reasons not only related to CCM. Admittedly, that's a fair amount of assumptions and "you should do this and that" advice.

So while I'd lean towards adapting the goroutine approach allowing us to address the concerns I outlined above now and dealing with potential quota issues if/when they arise, I'm also okay with sticking to the PR approach. The only bits I'd then like to ask for is to use cached droplet data if the reload fails (and [emit an event](https://github.com/kubernetes/sample-controller/blob/0e5b089d858f35daf461613074207039f23cac77/controller.go#L291) to inform the user [point no. 3 from my list above]) and maybe also _try_ to address the concurrency / blocking user issue. (I say try because I can see this may not be trivial to implement with the current approach, so it's not a must-have) [point no. 1].

I'll let you make the call. Once decided, we can proceed with the code review.

Thanks!

nanzhong:
@timoreimann would you mind taking another look at this PR and see if it align with what you had in mind?

I've moved to an sync loop and merged it together into the resources controller. There are still a few changes that I'd like to make (eg. make the `cloudprovider.LoadBalancer` implementation also use the same set of resources), but wanted to get this wip up so you can take a look.

cloud-controller-manager/do/resources.go
```diff
@@ -34,62 +35,211 @@ import (
 )
 
 const (
-	controllerSyncPeriod = 1 * time.Minute
-	requestTimeout       = 10 * time.Second
+	controllerSyncTagsPeriod      = 1 * time.Minute
+	controllerSyncResourcesPeriod = 1 * time.Minute
+	requestTimeout                = 1 * time.Minute
```
timoreimann: Just to confirm that you increased the request timeout to account for customers with many objects?

timoreimann: Also, can we rename this one to `syncTagsTimeout` to be consistent with the naming for the periods, and to tell more easily what this timeout is for compared to `syncResourcesTimeout` below?

cloud-controller-manager/do/resources.go
```diff
@@ -34,62 +35,211 @@ import (
...
+	var droplets []*godo.Droplet
+	for _, droplet := range c.dropletIDMap {
+		droplet := droplet
+		droplets = append(droplets, droplet)
+	}
+
+	return droplets
+}
+
+func (c *resources) LoadBalancerByID(id string) (droplet *godo.LoadBalancer, found bool, err error) {
```
timoreimann: No need for the `error` return type.

cloud-controller-manager/do/resources.go
```diff
@@ -34,62 +35,211 @@ import (
...
+	newNameMap := make(map[string]*godo.Droplet)
+
+	for _, droplet := range droplets {
+		droplet := droplet
+		newIDMap[droplet.ID] = &droplet
+		newNameMap[droplet.Name] = &droplet
+	}
+
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
```
timoreimann: Nit: no need to `defer` if we Unlock right after updating the two maps.

cloud-controller-manager/do/resources.go
```diff
@@ -34,62 +35,211 @@ import (
...
+	requestTimeout                = 1 * time.Minute
+	syncResourcesTimeout          = 3 * time.Minute
 )
 
 type tagMissingError struct {
 	error
 }
 
+type resources struct {
+	dropletIDMap      map[int]*godo.Droplet
```
timoreimann: Personally, I like to name maps by their lookup key, so in this case `dropletsByID`. This avoids encoding the type into the name (which could change) and makes it clear what is being mapped from and where to.

cloud-controller-manager/do/resources.go
```diff
@@ -34,62 +35,211 @@ import (
...
 // making sure the right tags are set.
 func (r *ResourcesController) Run(stopCh <-chan struct{}) {
-	ticker := time.NewTicker(controllerSyncPeriod)
-	defer ticker.Stop()
-	// Do not wait for initial tick to pass; run immediately for reduced sync
-	// latency.
-	for tickerC := ticker.C; ; {
-		if err := r.sync(); err != nil {
-			glog.Errorf("failed to sync load-balancer tags: %s", err)
+	syncResourcesTicker := time.NewTicker(controllerSyncResourcesPeriod)
```
timoreimann: The invocation of `Run()` is still [coupled to a non-empty cluster ID being provided](https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/200/files#diff-ad7d9529fb69c881ff9a4bb2a7d06ccdR111) (you may need to expand lines in Github as the line referenced by the link was unchanged) since the resource controller used to handle tags only. We need to move that check from `Initialize()` into this method.

cloud-controller-manager/do/resources.go
```diff
@@ -34,62 +35,211 @@ import (
 )
 
 const (
-	controllerSyncPeriod = 1 * time.Minute
-	requestTimeout       = 10 * time.Second
+	controllerSyncTagsPeriod      = 1 * time.Minute
+	controllerSyncResourcesPeriod = 1 * time.Minute
+	requestTimeout                = 1 * time.Minute
+	syncResourcesTimeout          = 3 * time.Minute
```
timoreimann: The timeout is currently covering all resources. I wonder if we should define it as timeout per resource type (and reduce accordingly) so that we can add potentially more resource types in the future without having to remember to bump / revisit the timeout?

cloud-controller-manager/do/resources.go
```diff
@@ -34,62 +35,211 @@ import (
...
-	ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
+func (r *ResourcesController) syncResources() error {
+	ctx, cancel := context.WithTimeout(context.Background(), syncResourcesTimeout)
 	defer cancel()
-	lbs, _, err := r.gclient.LoadBalancers.List(ctx, &godo.ListOptions{})
+
+	glog.V(2).Info("syncing droplet resources.")
+	droplets, err := allDropletList(ctx, r.gclient)
+	if err != nil {
+		return err
```
timoreimann: Can we change the logic so that failure to sync the droplets does not shortcut the method by returning an error?

IMHO, the method should not return an error but emit events when some of the resources cannot be sycned so that other resources still stand a chance to get synched.

cloud-controller-manager/do/resources.go
```diff
@@ -122,6 +272,7 @@ func (r *ResourcesController) sync() error {
 		return nil
 	}
 
+	tag := buildK8sTag(r.clusterID)
```
timoreimann: The `ResourcesController` used to store the cluster ID as tagged version so that we wouldn't need to remember attaching the tag everywhere we'd use it. Was there a specific reason to change that?

cloud-controller-manager/do/resources.go
```diff
@@ -34,62 +35,211 @@ import (
...
+
+	glog.V(2).Info("syncing droplet resources.")
+	droplets, err := allDropletList(ctx, r.gclient)
+	if err != nil {
+		return err
+	}
+	r.resources.UpdateDroplets(droplets)
+	glog.V(2).Info("synced droplet resources.")
+
+	glog.V(2).Info("syncing lb resources.")
```
timoreimann: Nit: we tend to refer to an LB as "load-balancer" in the logs consistently.

cloud-controller-manager/do/resources.go
nanzhong: Yup, exactly.

cloud-controller-manager/do/resources.go
nanzhong: Yeah, switched to just logging for now. I can update to emit events in another PR when we want/need to act on these failures.

cloud-controller-manager/do/resources.go
nanzhong: You're right, but I'm not sure how much tweaking this value will need. 3 minutes basically means 3 full-length api requests (I believe timeout is configured at 1 min on DO's api). It's hard to tell as well because depending on the number of resources/pages needed, this could vary wildly. 🤔 

Let's stick with 1 value that large value for now and once we have a better idea (have this running on more clusters/see issues), let's tweak this to be better? Pretty sure we'll need to revisit these timeouts + periods to be based on # of resources in the account.

timoreimann: 👌

cloud-controller-manager/do/resources.go
nanzhong: This change snuck in here when I thought we were returning the droplet ids for nodes in our `GET /v2/kubernetes/cluster/<uuid>` endpoint. I thought we could use that instead of listing all droplets in the account if the cluster id was present. I forgot that I removed droplet id in that api a while back :P. 

One thing that might help is if we built something like a `GET /v2/kubernetes/cluster/<uuid>/resources` endpoint or something that listed all the resources associated with a cluster that we could use here. 🤔 

Should I switch this back for now since it's not used or keep it here?

timoreimann: I think it's okay this way too, not a biggie. Was more curious about the reasoning behind it.

Yeah a dedicated API to list resources associated with a cluster would be nice.

timoreimann:
CI seems to be complaining about a test build failure:

```
# github.com/digitalocean/digitalocean-cloud-controller-manager/cloud-controller-manager/do [github.com/digitalocean/digitalocean-cloud-controller-manager/cloud-controller-manager/do.test]
cloud-controller-manager/do/resources_test.go:227:28: res.syncResources() used as value
```

cloud-controller-manager/do/resources_test.go
```diff
@@ -37,7 +37,207 @@ import (
...
+			expectedResources: &resources{
+				dropletIDMap:      map[int]*godo.Droplet{2: {ID: 2, Name: "two"}},
+				dropletNameMap:    map[string]*godo.Droplet{"two": {ID: 2, Name: "two"}},
+				loadBalancerIDMap: map[string]*godo.LoadBalancer{"one": {ID: "one"}},
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
```
timoreimann: This is missing a `test := test` variable capture (like in line 354) before the call to `t.Run()` (which is important when doing `t.Parallel()`).

cloud-controller-manager/do/resources_test.go
```diff
@@ -37,7 +37,207 @@ import (
...
+		t.Errorf("incorrect found\nwant: %#v\n got: %#v", want, got)
+	}
+	if want, got := droplet, foundDroplet; !reflect.DeepEqual(want, got) {
+		t.Errorf("incorrect droplet\nwant: %#v\n got: %#v", want, got)
+	}
+
+	_, found = resources.DropletByID(1000)
+	if want, got := false, found; want != got {
+		t.Errorf("incorrect found\nwant: %#v\n got: %#v", want, got)
+	}
```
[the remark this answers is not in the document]
nanzhong: Sure, updating. I think I'm used to github.com/stretchr/testify/assert that has expected before actual.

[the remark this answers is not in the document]
timoreimann: Yes, testify got it right/wrong.

cloud-controller-manager/do/resources.go
```diff
@@ -34,63 +35,228 @@ import (
...
+				return
+			}
+		}
+	}()
 }
 
-func (r *ResourcesController) sync() error {
-	// Fetch all load-balancers.
-	ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
+func (r *ResourcesController) syncResources() {
```
fatih: Can you add a doc comment that this function is responsible of updating the cache with up to date information?

timoreimann: Done.

fatih:
lgtm

cloud-controller-manager/do/resources_test.go
```diff
@@ -37,7 +39,352 @@ import (
...
+	defer cancel()
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				stop <- struct{}{}
+			default:
+				syncer.mutex.Lock()
+				if syncer.synced["resources syncer"] > 0 && syncer.synced["tags syncer"] > 0 {
```
timoreimann: This is going to busy-spin for up to 3 seconds while locking/unlocking continuously. I think a more elegant solution could be the following:

- make `newRecordingSyncer` accept the stop channel (and add a field to `recordingSyncer` for storing it)
- let `recorderSyncer`'s `recordingFn` close the stop signal as soon as it has been called once for each name
- reduce the `go func()` in the test to a simple `select` between the stop signal firing and a custom timeout defining an upper timeout for the test (like the current 3 seconds)
- call `t.Errorf` straight when the timeout has been hit and include the number of invocations per name

I believe this would remove the busy spinning, the locking, and simplify the test overall.

WDYT?

cloud-controller-manager/do/resources_test.go
```diff
@@ -37,7 +39,352 @@ import (
...
+	}
+
+	res := NewResourcesController(clusterID, fakeResources, inf.Core().V1().Services(), kclient, gclient)
+	syncer := newRecordingSyncer()
+	res.syncer = syncer
+
+	stop := make(chan struct{})
+	res.Run(stop)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
```
timoreimann: This can be simplified to a `timer.Timer`.

PR #203, opened 2019-03-29
Assume HTTPS protocol on default TLS port when TLS option is given.

### LINKED ISSUE #192: Handle situation where HTTP-to-HTTPS redirection is enabled but HTTPS service port is missing
Given a service spec like

```yaml
apiVersion: v1
kind: Service
metadata:
  name: app
  annotations:
spec:
  type: LoadBalancer
  selector:
    app: app
  ports:
    - name: http
      protocol: TCP
      port: 80
      targetPort: 80
```

HTTP-to-HTTPS won't work because the spec is missing an HTTPS service port like: 

```yaml
    - name: https
      protocol: TCP
      port: 443
      targetPort: 443
```

We should try to detect the problem and surface an error message / event.

timoreimann:
Previously, TLS options (such as the certificate ID or pass-through) would be ignored silently if either an HTTP* protocol or TLS port annotation was missing. With this change, we automatically enable the
HTTPS protocol on services exposing the default TLS port 443.

Effectively, this means that users may now omit the TLS port and protocol annotations for common use cases.

Fixes #192

@nanzhong @bouk @jcodybaker PTAL

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -432,6 +432,10 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
 	certificateID := getCertificateID(service)
 	tlsPassThrough := getTLSPassThrough(service)
 
+	if (certificateID != "" || tlsPassThrough) && len(tlsPorts) == 0 {
+		tlsPorts = append(tlsPorts, 443)
+	}
+
 	if len(tlsPorts) > 0 {
 		if certificateID == "" && !tlsPassThrough {
 			return nil, errors.New("must set certificate id or enable tls pass through")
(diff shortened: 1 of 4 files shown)


docs/controllers/services/annotations.md
```diff
@@ -35,14 +35,17 @@ The number of times a health check must pass for a backend Droplet to be marked
...
 
 ## service.beta.kubernetes.io/do-loadbalancer-tls-passthrough
 
 Specify whether the DigitalOcean Load Balancer should pass encrypted data to backend droplets. This is optional. Options are `true` or `false`. Defaults to `false`.
+If enabled, ports specified in `service.beta.kubernetes.io/do-loadbalancer-tls-ports` will use https, or 443 if none are given.
 
 ## service.beta.kubernetes.io/do-loadbalancer-certificate-id
 
-Specifies the certificate ID used for https. This annotation is required if `service.beta.kubernetes.io/do-loadbalancer-tls-ports` is used. To list available certificates and their IDs, use `doctl compute certificate list` or find it in the [control panel](https://cloud.digitalocean.com/account/security).
+Specifies the certificate ID used for https. To list available certificates and their IDs, use `doctl compute certificate list` or find it in the [control panel](https://cloud.digitalocean.com/account/security).
```
snormore: > or find it in the [control panel](https://cloud.digitalocean.com/account/security)

I'm pretty sure the ID is only available via `doctl`, I wasn't able to find it in the control panel, just the name and the fingerprint.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -432,6 +432,10 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
 	certificateID := getCertificateID(service)
 	tlsPassThrough := getTLSPassThrough(service)
 
+	if (certificateID != "" || tlsPassThrough) && len(tlsPorts) == 0 {
+		tlsPorts = append(tlsPorts, 443)
```
jcodybaker: I feel like should only do this if port 443 is listed among the entryPorts.  Otherwise, this is moot.  It seems likely if the user forgot to include to include 443 in their `do-loadbalancer-tls-ports`, they also forgot it in their entryPorts, especially if they're using a certificateID.  

Alternatively, if they do provide entry port 80, `do-loadbalancer-certificate-id` and `do-loadbalancer-redirect-http-to-https` we could safely assume the port 80 should entry point target port should be really be used for port 443 and create it implicitly.

docs/controllers/services/annotations.md
timoreimann: Great catch @snormore, I never tried out the web UI myself. Pushed a commit to remov the reference to the cloud panel.

PR #206, opened 2019-04-03
workaround: cluster->lb traffic w/ TLS/ProxyProto

jcodybaker:
This PR excludes traffic with Layer7 rules from a kube-proxy optimization.
The kube-proxy optimization intercepts and redirects cluster egress traffic
destined for the LB to the backing services, avoiding the DigitalOcean LB
entirely.  If the DO LB is providing L7 services like TLS termination, Proxy
Protocol, or StickySessions this breaks access to LB-hosted services
from within the cluster. This PR opts out of the optimization by returning
the IP address in the Hostname field rather than IP.  The IP will be reported
to the user in the same way via tools like kubectl, but is interpreted differently
by kube-proxy.

https://github.com/kubernetes/kubernetes/issues/66607

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -168,14 +168,7 @@ func (l *loadBalancers) GetLoadBalancer(ctx context.Context, clusterName string,
 	if lb.Status != lbStatusActive {
 		return nil, true, fmt.Errorf("load-balancer not active, currently %s", lb.Status)
 	}
-
-	return &v1.LoadBalancerStatus{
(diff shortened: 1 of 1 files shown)


jcodybaker:
This doesn't work because there's validation to explicitly prevent this:
```service_controller.go:353] Failed to persist updated LoadBalancerStatus to service 'default/server' after creating its load balancer: Service "server" is invalid: status.loadBalancer.ingress[0].hostname: Invalid value: "178.128.134.186": must be a DNS name, not an IP address```
https://github.com/kubernetes/kubernetes/blob/v1.13.5/pkg/apis/core/validation/validation.go#L5316-L5334

PR #165, opened 2019-01-31
vendor: update k8s.io packages to v1.13.2

MorrisLaw:
Update Kubernetes version.

PR #208, opened 2019-04-11
Migrate from go dep to Go modules.

timoreimann:
As a side effect, we hope to simplify upgrading to a more recent version of the Kubernetes dependencies (which has been someone difficult with go dep, see also #165).

- Run `go mod init`.
- Remove Go versions not supporting Go modules from Travis.
- Use Go 1.12.3 for building.
- Vendor golint through tools.go approach.
- Remove go dep's Gopkg.toml.
- Update vendor folder.


timoreimann:
@nanzhong @jcodybaker @fatih PTAL

fatih:
One last question. Did you run `go mod tidy` before opening this? I know it's part of the new `vendor` task in the makefile, but wasn't sure if you have called it or not.

lgtm

timoreimann:
@fatih just ran `make vendor` again to be sure: all vendored dependencies are tidied up and in order.

We probably want another script in the future that validates all vendored files have been properly added through gomod and run that during CI. That's for a future PR.

Issue #102, opened 2018-05-17
Feature Request : Allow name prefix on created resources


peterver:
At the moment when for example a load balancer is provisioned through CCM it gets assigned a name that is bound to be unique.

This approach works beautifully for a purely machine-based operator. But since there is no way to prefix a resource name it leads to confusion when looking at the digitalocean dashboard. As such it would/could be beneficial to allow for a prefix to be set ( a prefix, not an override ).

For example : let's say a load balancer in the current scheme would get a uid of qwerty123
with a prefix it could become something like : staging-api-qwerty123

Which is both unique for automation reasons and allows for readability to human operators

This behavior can be added through an annotation, or could take the name of the k8s service managing it as the prefix.

In regards to verification it would need to take into account any limitations digitalocean poses on the length of the name that a resource such as a loadbalancer can have

andrewsykim:
Prefixing will be tricky because we actually rely on the service controller from upstream Kubernetes to handle this part for us. We could add a prefix to whatever name is generated by Kubernetes but that could lead to unexpected behaviour from the service controller.  Might be better to add that logic to the [Kubernetes service controller](https://github.com/kubernetes/kubernetes/tree/master/pkg/controller/service) instead of trying to patch it here. What do you think?

andrewsykim:
Yeah, unfortunately `cloudprovider.GetLoadBalancerName` is a shared method and not something that can be overwritten by each cloud provider (yet). As mentioned in the issue, changing something like the loadbalancer name requires very careful consideration since you don't want to break code paths that need to know about the previous LB names.  

Having said that, it's not impossible to override LB names that are passed in from the service controller, but it may lead to unexpected behaviours in the future since service controller may have assumptions about the LB name. I think annotations to override this would be okay as long as users are aware of the implications 😬.

lxfontes:
feels like what we really want here is LB tags ;)

andrewsykim:
+1 for tags, upstream recently merged https://github.com/kubernetes/kubernetes/pull/66589 as well so starting in v1.12 we have to implement our own GetLoadBalancerName method (optional fallback to current naming format). Only tricky bit here is accounting for existing LBs with the old name.

peterver:
+1 :) would love to see lb tags :D, for me the whole point is to be able to organize resources. Whether that is through tags or naming doesn't matter. As long as it's possible :) 

Thanks for replying @lxfontes @andrewsykim

[...]

PR #209, opened 2019-04-15
Update Kubernetes dependencies to 1.14.1 and adjust code

timoreimann:
Besides being more up-to-date in general and supporting Kubernetes 1.14, the update will also allow us to customize LB names in the future (and thus address #102).

Changes in detail:

- Update package imports.
- Implement new interface function around load balancer naming (without modifying the naming schema).
- Use cobra command-based bootstrapping method to ensure all flags are set properly. (This will require enabling the allow-untagged-cloud flag on the CLI.)
- Update end-to-end tests.
- Follow upstream in switching from glog to klog.

**Note:** Due to the way cloud providers are to be initialized now (basically by calling a factory function that returns a Cobra command), we cannot in-code-enable the flag allowing untagged clouds easily anymore. As such, the flag should move to the CLI in form of the `--allow-untagged-cloud` parameter.


timoreimann:
This supersedes #165.

timoreimann:
I needed to pin the dependency for apimachinery (`replace` in Go modules lingo) to 1.14.1 for some reason. Haven't digged into why it's needed, but it certainly is.

timoreimann:
Submitted an upstream fix for the workaround we had to do around the GCE-only flag problem (see [main.go](https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/209/commits/c88a079b748d4c1b7648181b2e3d27d2dc48f3f3#diff-a29863390f1b096d59524d3e76ad507dR40) in this PR).

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager/main.go
```diff
@@ -17,51 +17,40 @@ limitations under the License.
...
+	// utilflag.InitFlags() (by removing its pflag.Parse() call). For now, we have to set the
+	// normalize func and add the go flag set by hand.
+	// utilflag.InitFlags()
 	logs.InitLogs()
 	defer logs.FlushLogs()
 
-	verflag.PrintAndExitIfRequested()
-
-	// digitalocean overrides
-	s.KubeCloudShared.AllowUntaggedCloud = true
```
jcodybaker: @timoreimann and I had an offline discussion about this.  Because of the upstream changes were no longer able to do this, but it sounds like we need to change the parameters required to invoke the CCM.  We should also update the documentation in this PR so they ride together.

timoreimann: Good catch -- I pushed a commit that updates the docs. PTAL again.

timoreimann:
Also confirming that the new version was tested on Kubernetes 1.11 through 1.13. The latest release 1.14 will be tested as soon as images are readily available for us.

PR #212, opened 2019-04-22
Fix stale cache node deletion

nanzhong:
There is a race condition that can happen at the moment when a stale resource cache for droplets can cause nodes to be deleted. The race happens when the cache for droplets is missing an actual droplet that was created since the last sync and the node controller tries to check for the existence of that node.

This PR moves the logic for update droplets and lbs into the resources struct, and introduces a `SyncDroplet` method that ensure that a droplet, given by id, is up to date in the cache. When the existence check is run, we first ensure that the droplet is up to date in the cache, before looking for its existence.

There is still a window here where a race is possible (sync run started before `SyncDroplet` is called, but doesn't finish until after `SyncDroplet` is complete and before `GetDropletByID` is called, but this window should be very unlikely).

### DIFF
--- a/cloud-controller-manager/do/cloud.go
+++ b/cloud-controller-manager/do/cloud.go
@@ -87,7 +87,7 @@ func newCloud() (cloudprovider.Interface, error) {
 
 	clusterID := os.Getenv(doClusterIDEnv)
 	clusterVPCID := os.Getenv(doClusterVPCIDEnv)
-	resources := newResources(clusterID, clusterVPCID)
+	resources := newResources(clusterID, clusterVPCID, doClient)
 
 	return &cloud{
 		client:        doClient,
--- a/cloud-controller-manager/do/droplets.go
+++ b/cloud-controller-manager/do/droplets.go
@@ -86,10 +86,19 @@ func (i *instances) ExternalID(ctx context.Context, nodeName types.NodeName) (st
 }
 
 // InstanceID returns the cloud provider ID of the droplet identified by nodeName.
-func (i *instances) InstanceID(_ context.Context, nodeName types.NodeName) (string, error) {
+func (i *instances) InstanceID(ctx context.Context, nodeName types.NodeName) (string, error) {
 	droplet, found := i.resources.DropletByName(string(nodeName))
 	if !found {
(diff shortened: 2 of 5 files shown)


nanzhong:
> 1. The cloud provider interface prescribes [that `InstanceID` must return `("", cloudprovider.InstanceNotFound)` if no instance is found](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/4d82a0c66dc2931ad6d2f163ca3333d0a238fccd/vendor/k8s.io/cloud-provider/cloud.go#L151-L154). Doesn't that imply that we should do a call to `SyncDroplet` in our implementation of that method as well? (Upstream, both `InstanceID` and `InstanceExistsByProviderID` are [called from the same function](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/4d82a0c66dc2931ad6d2f163ca3333d0a238fccd/vendor/k8s.io/kubernetes/pkg/controller/cloud/node_controller.go#L347).)

Good catch. I'm not sure how to address this to be honest... we don't have an API to find a droplet by name... we have to iterate across all droplets to translate droplet name -> droplet id. Doing that for every call to `InstanceID` basically defeats a large part of the cache. On the other hand, not doing that means we're not solving the problem... I need to thing about this a little.

> I'm not sure I understand why this would be problematic, could you explain please? AFAIU, you are talking about the following flow:
> 
> 1. resource sync starts
> 2. SyncDroplet starts
> 3. syncDroplet returns
> 4. resource sync returns
> 5. DropletByID starts

Depending on how many pages of droplets there are and what page the synced droplet lies on it's possible that the resource sync could have missed this new droplet overwriting the cache when it returns.

nanzhong:
@timoreimann could you take another look?

I don't have a good solution to the issue with `InstanceID` you pointed out. There's no real solution here until we either:

1. have an api that lets us lookup a droplet by name
2. can deterministically derive droplet id from name

The solution I added for now just forces a resync when it can't find a droplet by name. This is not ideal, but I'm not sure I have a better solution now that doesn't add a lot of complexity (eg. keeping track of not found droplets and confirming existence via second sync, etc.).

PR #215, opened 2019-05-14
Set a custom user agent for the godo client.

andrewsomething:
Currently, requests made by digitalocean-cloud-controller-manager send `godo/$godoVersion` as their user agent. This change updates the user agent to use the format `digitalocean-cloud-controller-manager/$ccmVersion godo/$godoVersion` in order to help us better understand `godo` usage.

### DIFF
--- a/cloud-controller-manager/do/cloud.go
+++ b/cloud-controller-manager/do/cloud.go
@@ -37,6 +37,8 @@ const (
 	providerName        string = "digitalocean"
 )
 
+var version string
+
 type tokenSource struct {
 	AccessToken string
 }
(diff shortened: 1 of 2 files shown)


cloud-controller-manager/do/cloud.go
```diff
@@ -66,6 +68,12 @@ func newCloud() (cloudprovider.Interface, error) {
 		opts = append(opts, godo.SetBaseURL(overrideURL))
 	}
 
+	if version != "" {
+		opts = append(opts, godo.SetUserAgent("digitalocean-cloud-controller-manager"+"/"+version))
+	} else {
+		opts = append(opts, godo.SetUserAgent("digitalocean-cloud-controller-manager"+"/dev"))
```
timoreimann: I think the new section could be slightly simplified by setting `version` appropriately and doing a single append call only, i.e.:

```go
if version == "" {
  version = "dev"
}
opts = append(opts, godo.SetUserAgent("digitalocean-cloud-controller-manager"+"/"+version))
```

andrewsomething: :ok_hand: I like it. Updated.

andrewsomething:
@timoreimann Yes. I've successfully seen requests logged with the user agent `digitalocean-cloud-controller-manager/v0.1.14 godo/1.11.0`

PR #216, opened 2019-05-15
Fix RBAC

tahsinrahman:


timoreimann:
@tahsinrahman thanks for your PR. I suppose you are suggesting the change for reasons similar to #217? If so, I'd like to keep the PR open until we have concluded whether this is something we want to address in terms of an RBAC extension. (My initial guess is that the permission being asked for should not be required in the first place.)

mattkelly:
@timoreimann I'm in favor of merging this for now with an eye towards removing it once the upstream stuff is tidied up. I also think we might have to add it to previous release manifests (not 100% positive - see below).

It seems that the code path that hits this is only taken sometimes. I'm not sure what the path is but it's probably pretty easy to figure out.

We recently started seeing some e2e test failures for clusters which run the DO CCM, and I _think_ that this is the cause. Those clusters were still running v0.1.7 but on Kubernetes 1.12+. I know that the Kubernetes dependencies for this CCM weren't updated until v0.1.14, so it is a little curious depending on the actual code path, but this has likely [been an issue since Kubernetes 1.12](https://github.com/kubernetes/kubernetes/issues/68986). Unfortunately, those clusters are all gone now, so I can't be 100% sure that it was this bug - but it seems highly likely because the way it manifested was exactly the same as the issue on v0.1.14 that I did debug.

I hit this issue every time when trying to provision a fresh Kubernetes 1.14.3 cluster with v0.1.14 of the CCM. I've verified that adding this `configmaps` rule fixes the issue. (The DO CCM is a plugin in Containership land, and I'm adding support for v0.1.14 currently.)

The problem is that it's very difficult to debug. The way it manifests for me is that the CCM crashloops with the following not-very-useful log:

```
Error from server: no preferred addresses found; known addresses: []
```

I think this happens because of the weird chicken-and-egg thing where the CCM, if it exists, is needed to set the node's IP addresses. But if that can't be done successfully, then you end up in this weird state where the CCM won't come up and also doesn't have the useful log re: missing the RBAC rule.

If I add the `configmaps` rule and let it come up once so the node addresses are initialized properly, and then remove the rule and restart the CCM, I get the useful error message:

```
W0621 16:34:19.459188       1 authentication.go:262] Unable to get configmap/extension-apiserver-authentication in kube-system.  Usually fixed by 'kubectl create rolebinding -n kube-system ROLEBINDING_NAME --role=extension-apiserver-authentication-reader --serviceaccount=YOUR_NS:YOUR_SA'
configmaps "extension-apiserver-authentication" is forbidden: User "system:serviceaccount:kube-system:cloud-controller-manager" cannot get resource "configmaps" in API group "" in the namespace "kube-system"
```

Sorry for the wall of text, but I hope this helps explain why it's necessary to merge a workaround for this.

timoreimann:
Hey @mattkelly,

appreciate the detailed description, thanks!

I went on a hunt to try to better understand the root cause of this issue. Following is what I was able to gather.

At least with the Kubernetes v1.14 dependencies that got rolled into the DO-CCM v0.1.14 release, it seems to be that the error message reported in #217 and by you (`Unable to get configmap/extension-apiserver-authentication in kube-system`) is either invoked right after CCM starts up (that is, before it gets even close to bootstrapping nodes) or never at all. It's being invoked when CCM (more accurately, the upstream part of it in the cloud provider framework) determines that it doesn't have all the configuration pieces together to authenticate clients (such as CA certificates).

One way to deal with this is to extend the RBAC rules. Another one is to skip authentication lookups via the `--authentication-skip-lookup=true` CLI switch. The difference between the two is that the latter will leave the client authentication configuration empty/non-working. IMHO, we should prefer skipping auth configuration because client authentication and token access review creation should not be needed with our CCM (unlike the API server, which may be why the upstream issue was solved by extending the RBAC ruleset instead). Please let me know though if you can think of a use case why we should have client authentication in DO-CCM.

I merged #226 earlier which hard-codes the skip switch in the code. (FWIW, the flag can still be overridden on the CLI.) Matt, do you have a chance to test bootstrapping on your end with the latest DO-CCM master or the v0.1.15 release I am planning to cut this week? I'm hoping and expecting that the skip approach should fix your problem as well.

(FWIW, I think the `Error from server: no preferred addresses found; known addresses: []` indicates that no node address of a type specified by one of the `--kubelet-preferred-address-types` API server flag values could be found. Can you double-check that your nodes are reachable by one of the given types?)

mattkelly:
Thanks for the great explanation, @timoreimann! I'm more than happy to test v0.1.15 and report back. If I have time before that's released, I may also try just testing master as you suggested (though no promises there).

Good to know about the `--kubelet-preferred-address-types` thing. I will dig in more this week.

timoreimann:
[v0.1.15 has been released](https://github.com/digitalocean/digitalocean-cloud-controller-manager/releases/tag/v0.1.15). 🎉

I'll be closing the PR, but everyone feel free to create a new issue if it turns out that the related issue is still not fixed.

Thanks!

Issue #218, opened 2019-05-23
Public/private IP addresses cannot be used as node names anymore


yamamoto-febc:
(from: https://github.com/yamamoto-febc/terraform-provider-rke/pull/88#discussion_r286410875)

I tried ccm by specifying private IP as the node name with following options: 

- kube-apiserver: `--kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname`
- kubelet: ` --cloud-provider=external`

However, ccm reported following error.

```sh
NodeAddress: Error fetching by providerID: providerID cannot be empty string Error fetching by NodeName: instance not found
```

I found in #56 that private or public IP was supported, but the current code seems to support only the droplet name.

- func `NodeAddress` calls to `resource.DropletByName`

https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/cloud-controller-manager/do/droplets.go#L52-L58

- `resource.DropletByName` finds droplet from `dropletNameMap`

https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/cloud-controller-manager/do/resources.go#L83-L89

- `dropletNameMap` is set from the following two places, but public/private IP is not used

https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/cloud-controller-manager/do/resources.go#L147
https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/cloud-controller-manager/do/resources.go#L210

### versions

- kubernetes: v1.13
- ccm: v0.1.13

timoreimann:
Hey @yamamoto-febc,

thanks for filing the issue. Indeed, I believe we introduced a regression when we added the caching layer along #200. :(

One way to deal with this is to expand on the keys we support within [`SyncDroplet`](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/151e75b4f4ce04a025787a2e5d64d82c2e75a5b4/cloud-controller-manager/do/resources.go#L180) (and likely also [`UpdateDroplets`](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/151e75b4f4ce04a025787a2e5d64d82c2e75a5b4/cloud-controller-manager/do/resources.go#L133:21)) by including addresses. I also happen to be experimenting with an idea currently that may allow us to remove part of the caching layer. To my understanding, it should return the ability to use private and public IP addresses as a convenient side effect.

[...]

PR #223, opened 2019-06-13
Drop droplet cache usage in Instances implementation

timoreimann:
This change reverts part of the caching layer introduced along #200 in favor of the older synchronous implementation. It is done on the assumption that the provider ID is passed on by each node's kubelet instance through the `--provider-id digitalocean://<droplet id>` parameter, with the droplet ID being available via the metadata service. (See the amended documentation for details.) Nodes having the provider ID preset in this way allow the cloud provider framework to do O(1) DO API lookups of droplet data, thereby rendering cache usage obsolete.

For DigitalOcean accounts with low to medium droplet counts, the described kubelet change can likely be skipped without any detrimental effects to performance; if anything, node bootstrapping latency should presumably improve due to cache synchronization delays not being at play anymore.

We continue to use the droplet cache for load-balancers where associations to droplets can still be costly when done synchronously (due to the linear complexity involved).

I have successfully verified the change with node pools as big as 50 nodes and 500 non-cluster droplets sitting in the same DO account. I also made sure LB integration would continue to work.


docs/getting-started.md
```diff
@@ -34,6 +34,16 @@ SIG Cloud Provider requires all cloud providers to specify a cluster ID in order
...
+
+For DigitalOcean accounts with small to medium numbers of droplets, iteration should not be much of an issue. However, accounts having a large number of droplets (at the order of hundreds) would likely experience observable performance degradations.
+
+A solution to the problem is to pass the provider ID (aka droplet ID) via the kubelet on bootstrap through the `--provider-id=digitalocean://<droplet ID>` parameter. The droplet ID can be retrieved from the metadata service available on each droplet at the `http://169.254.169.254/metadata/v1/id` address. Passing the provider ID like this causes the node annotation to be available right from the start, thereby enabling fast API lookups for droplets at all times.
+
+DigitalOcean's managed Kubernetes offering DOKS sets the parameter ID on each worker node kubelet instance.
```
nanzhong: `s/parameter ID/provider ID/`

timoreimann: Good catch, addressed.

timoreimann:
This one should also fix #218.

PR #225, opened 2019-06-21
Add tests for droplet matching

timoreimann:
Add tests for droplet matching unrelated to the droplet name as provided by the `dropletByName()` function.
For test resilience reasons, we run the tests through a public method.

Refs #218.


fatih:
lgtm 👍

PR #230, opened 2019-07-03
Check for unused and missing packages in go.mod + vendor

snormore:
To avoid missing go.mod updates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,13 @@ all: test
 
 publish: clean ci compile build push
 
-ci: check-headers gofmt govet golint test
+ci: check-headers check-unused gofmt govet golint test
+
+.PHONY: check-unused
+check-unused:
+	@GO111MODULE=on go mod tidy
+	@GO111MODULE=on go mod vendor
+	@git diff --exit-code -- go.sum go.mod vendor/ || ( echo "there are uncommitted changes to the Go modules and/or vendor files -- please run 'make vendor' and commit the changes first"; exit 1 )
 
 .PHONY: e2e
 e2e:
@@ -83,7 +89,6 @@ else
 	@docker push $(REGISTRY)/digitalocean-cloud-controller-manager:$(VERSION)
 	@echo "==> Your image is now available at $(REGISTRY)/digitalocean-cloud-controller-manager:$(VERSION)"
 endif
-
 
 .PHONY: govet
 govet:


Makefile
```diff
@@ -102,7 +100,7 @@ gofmt: # run in script cause gofmt will exit 0 even if files need formatting
 .PHONY: test
 test:
 	@echo "==> Testing all packages"
-	@go test -race $(shell go list ./... | grep -v vendor)
+	@GO111MODULE=on go test -mod=readonly -race $(shell go list ./... | grep -v vendor)
```
timoreimann: We have a bit of a conflict here because the usage of Go modules is required to use the `-mod=readonly` flag. For people like myself who still work exclusively off of the vendor folder, This would be somewhat of a workflow change. Moreover, we would also stop using the vendor folder during tests in the CI. I think we need a different approach.

Sadly, https://github.com/golang/go/issues/31372 is not a thing yet, which would allow us to run `go mod tidy -mod=readonly`. What we could do instead is follow [the suggestion](https://github.com/golang/go/issues/26850) to run `go list -mod=readonly all`. My understanding is though that this does not check other architectures (for which we have no specific code right now but could potentially have in the future) and seemingly only applies to `go.mod`, not `go.sum` (though I'm not absolutely sure on the last point).

Another idea I came up with is to use `go mod tidy` and check for git diff'erences to the `go.*` files. To cover unexpected or missing changes to the vendor folder too, we can apply the same logic by running `go mod vendor`. Turns out [the Prometheus Makefile](https://github.com/prometheus/prometheus/blob/d7f38dfddee25b8836eb22080977dc59daf3a93c/Makefile.common#L187-L196) does just that, so my recommendation would be to copy that. This approach doesn't use `-mod=readonly` anymore but should still work reliably.

WDYT?

snormore: I like the latter option too, will update :+1:

snormore: WDYT? https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/230/commits/16c43be4dc517f206f30d674394bfe0d8c9a329a

Makefile
```diff
@@ -34,7 +34,15 @@ all: test
...
-ci: check-headers gofmt govet golint test
+ci: check-headers check-unused gofmt govet golint test
+
+.PHONY: check-unused
+check-unused:
+	@echo "==> Running check for unused/missing packages in go.mod"
+	@GO111MODULE=on go mod tidy
+	@echo "==> Running check for unused packages in vendor/"
+	@GO111MODULE=on go mod vendor
+	@git diff --exit-code -- go.sum go.mod vendor/
```
timoreimann: I think it'd helpful to output a message to the user indicating how a failure of this check can be addressed. That is, I'd suggest to expand the line like (warning, untested)

```
@git diff --exit-code -- go.sum go.mod vendor/ || { echo "there are uncommitted changes to the Go modules and/or vendor files -- please run 'make vendor' and commit the changes first"; exit 1 }
```

Makefile
```diff
@@ -34,7 +34,15 @@ all: test
 
 publish: clean ci compile build push
 
-ci: check-headers gofmt govet golint test
+ci: check-headers check-unused gofmt govet golint test
+
+.PHONY: check-unused
+check-unused:
+	@echo "==> Running check for unused/missing packages in go.mod"
```
timoreimann: uber-nit pick: I think Go modules distinguishes between packages and modules:

```suggestion
	@echo "==> Running check for unused/missing modules in go.mod"
```

timoreimann:
@snormore if you haven't done so yet, could you please run a test to make sure the change does what it should both for `go.*` and as well as `vendor/`?

snormore:
@timoreimann Updated and tested ✅

PR #233, opened 2019-07-05
Add sticky session annotation note about externalTrafficPolicy Local and max one pod per node

snormore:
Add notes indicating:
 - Sticky sessions will route consistently to the same nodes, not pods, so you should avoid having more than one pod per node serving requests.
 - Sticky sessions require your Service to configure `externalTrafficPolicy: Local` to avoid NAT confusion on the way in.

### DIFF
--- a/docs/controllers/services/examples/http-with-sticky-sessions.yml
+++ b/docs/controllers/services/examples/http-with-sticky-sessions.yml
@@ -40,11 +40,11 @@ spec:
         - containerPort: 80
           protocol: TCP
       # This is necessary for sticky-sessions because it can
-      # only consistently route to to the same nodes, not pods. 
+      # only consistently route to the same nodes, not pods.
       affinity:
         podAntiAffinity:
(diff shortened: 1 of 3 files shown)


docs/controllers/services/examples/http-with-sticky-sessions.yml
```diff
@@ -36,3 +39,12 @@ spec:
...
         - containerPort: 80
           protocol: TCP
+      # This is necessary for sticky-sessions because it can
+      # only consistently route to the same nodes, not pods.
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchLabels:
+                app: hello
```
timoreimann: Wouldn't this need to say `app: nginx-example` because that's what the label is like in line 34?

snormore: Nice catch, was using the `hello` app to test. Updated :+1:

PR #234, opened 2019-07-05
Default to HTTP for sticky sessions if no protocol is defined

snormore:
This PR updates `buildForwardingRules` to default to protocol `http` if sticky sessions enabled and `tcp` (or no protocol) is set. Otherwise this can be pretty confusing for the user if they don't explicitly define the protocol via annotation.

### DIFF
--- a/cloud-controller-manager/do/loadbalancers_test.go
+++ b/cloud-controller-manager/do/loadbalancers_test.go
@@ -864,6 +864,129 @@ func Test_buildForwardingRules(t *testing.T) {
 			nil,
 		},
 		{
+			"default forwarding rules with sticky sessions",
+			&v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+					UID:  "abc123",
+					Annotations: map[string]string{
+						annDOStickySessionsType:       "cookies",
+						annDOStickySessionsCookieName: "DO-CCM",
+						annDOStickySessionsCookieTTL:  "300",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Ports: []v1.ServicePort{
+						{
+							Name:     "test",
+							Protocol: "TCP",
+							Port:     int32(80),
+							NodePort: int32(30000),
+						},
+					},
+				},
+			},
+			[]godo.ForwardingRule{
+				{
+					EntryProtocol:  "http",
+					EntryPort:      80,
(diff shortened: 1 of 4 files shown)


cloud-controller-manager/do/loadbalancers_test.go
```diff
@@ -863,6 +863,129 @@ func Test_buildForwardingRules(t *testing.T) {
 			},
 			nil,
 		},
+		{
+			"default forwarding rules",
```
timoreimann: The test name [seems to be used already](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/749a63e79dc08d4f1ed7f66d7fa69a7c748d1b04/cloud-controller-manager/do/loadbalancers_test.go#L612)? Copy-pasta?

cloud-controller-manager/do/loadbalancers_test.go
```diff
@@ -863,6 +863,129 @@ func Test_buildForwardingRules(t *testing.T) {
...
+					TargetProtocol: "http",
+					TargetPort:     30000,
+					CertificateID:  "",
+					TlsPassthrough: false,
+				},
+			},
+			nil,
+		},
+		{
+			"http forwarding rules",
```
timoreimann: [Also used already](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/749a63e79dc08d4f1ed7f66d7fa69a7c748d1b04/cloud-controller-manager/do/loadbalancers_test.go#L642)?

cloud-controller-manager/do/loadbalancers_test.go
```diff
@@ -863,6 +863,129 @@ func Test_buildForwardingRules(t *testing.T) {
...
+					TargetProtocol: "http",
+					TargetPort:     30000,
+					CertificateID:  "",
+					TlsPassthrough: false,
+				},
+			},
+			nil,
+		},
+		{
+			"https forwarding rules with sticky sessions",
```
timoreimann: Is the base of this test similar enough to ["tls forwarding rules"](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/749a63e79dc08d4f1ed7f66d7fa69a7c748d1b04/cloud-controller-manager/do/loadbalancers_test.go#L675)that we may want to rename it to "tls forwarding rules with sticky sessions"? WDYT?

PR #231, opened 2019-07-03
Add HTTP/2 support for LBs

snormore:
Currently pushed to `snormore/digitalocean-cloud-controller-manager:dev` if you want to try it out.

~Rebased on top of https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/236 and includes changes from https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/235 and https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/234 until they're merged.~

~Rebased on top of https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/235.~

### Example

```
kind: Service
apiVersion: v1
metadata:
  name: hello
  annotations:
    service.beta.kubernetes.io/do-loadbalancer-certificate-id: "my-cert-uuid"
    service.beta.kubernetes.io/do-loadbalancer-protocol: "http2"
spec:
  type: LoadBalancer
  selector:
    app: hello
  ports:
    - name: http
      protocol: TCP
      port: 443
      targetPort: 80
---
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: hello
spec:
  replicas: 3
  template:
    metadata:
      labels:
        app: hello
    spec:
      containers:
      - name: hello
        image: snormore/hello
        ports:
        - containerPort: 80
          protocol: TCP
```

```
$ curl -v https://blue.snormore.dev
...
* Using HTTP2, server supports multi-use
...
Hello, you've requested: /
```

PR #235, opened 2019-07-05
Default health check protocol to TCP if not specified

snormore:
Default health check protocol to TCP if not specified, since DO LBs support only TCP or HTTP for these anyway. https://developers.digitalocean.com/documentation/v2/#load-balancers

First of 2 PRs to be extracted from https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/231

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -389,17 +389,6 @@ func buildHealthCheck(service *v1.Service) (*godo.HealthCheck, error) {
 		return nil, err
 	}
 
-	// if no health check protocol is specified, use the protocol used for
-	// load balancer traffic.
-	if healthCheckProtocol == "" {
-		protocol, err := getProtocol(service)
-		if err != nil {
-			return nil, err
-		}
-
-		healthCheckProtocol = protocol
-	}
-
 	checkIntervalSecs, err := healthCheckIntervalSeconds(service)
 	if err != nil {
 		return nil, err
@@ -567,11 +556,13 @@ func getProtocol(service *v1.Service) (string, error) {
 	return protocol, nil
 }
 
-// healthCheckProtocol returns the health check protocol as specified in the service
+// healthCheckProtocol returns the health check protocol as specified in the service,
+// falling back to TCP if not specified.
 func healthCheckProtocol(service *v1.Service) (string, error) {
-	protocol, ok := service.Annotations[annDOHealthCheckProtocol]
-	if !ok {
-		return "", nil
+	protocol := service.Annotations[annDOHealthCheckProtocol]
+
+	if protocol == "" {
+		return protocolTCP, nil
 	}
 
 	if protocol != protocolTCP && protocol != protocolHTTP {
--- a/cloud-controller-manager/do/loadbalancers_test.go
+++ b/cloud-controller-manager/do/loadbalancers_test.go
@@ -1237,7 +1237,7 @@ func Test_buildHealthCheck(t *testing.T) {
 			healthcheck: defaultHealthCheck("tcp", 30000, ""),
 		},
 		{
-			name: "http health check",
+			name: "default health check with http service protocol",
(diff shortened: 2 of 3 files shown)


snormore:
@timoreimann comments addressed, thanks :pray:

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -388,16 +388,26 @@ func buildHealthCheck(service *v1.Service) (*godo.HealthCheck, error) {
...
 		protocol, err := getProtocol(service)
 		if err != nil {
 			return nil, err
 		}
 
-		healthCheckProtocol = protocol
+		tlsPassThrough := getTLSPassThrough(service)
+
+		if tlsPassThrough || (protocol != "https") {
+			healthCheckProtocol = protocol
```
timoreimann: We are going to set the health check protocol to `https` here If `tlsPassThrough` is true and `protocol` happens to be _https_. However, the [Load Balancer API supports "http" and "tcp" only](https://developers.digitalocean.com/documentation/v2/#load-balancers) (which is also what we validate in our health check protocol getter).

I am thinking that we could simplify the entire logic as following:

- If `healthCheckProtocol` is set: use it.
- If `healthCheckProtocol` is _not_ set: use TCP

While we could try to be smart and guess if the user wants TCP or HTTP as the health check protocol based on what the TLS pass-through and protocol values look like, I don't think we can really make any reliable assumptions in this case: say that TLS pass-through is **not** set and protocol is set to HTTPS, does that really mean that the user wants to have HTTP set up as the health check protocol? If the application does not implement HTTP verb semantics correctly, even HTTP GET requests for health checks could mutate the state within the application in undesirable ways. (That's a big no-go you shouldn't do, but sometimes you can't easily do the right thing if you're dealing with third-party application.)

My gut feeling is we should stick to the most simple way that is still helpful, which is to fall back to TCP. This would mean a breaking change in behavior but I think it's the better option than taking fuzzy assumptions; even more so when HTTP/2 becomes supported.

WDYT?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -388,16 +388,26 @@ func buildHealthCheck(service *v1.Service) (*godo.HealthCheck, error) {
...
+		} else {
+			// We already validate that the port.Protocol is TCP, so we
+			// can use it as-is here.
+			healthCheckProtocol = strings.ToLower(string(port.Protocol))
```
timoreimann: I think this should simply assign `protocolTCP` as that constant is specifically designed to match the values that the DO API expects.

cloud-controller-manager/do/loadbalancers.go
snormore: That makes a lot of sense to me. I pushed an update to that effect :+1:

cloud-controller-manager/do/loadbalancers_test.go
```diff
@@ -1328,7 +1378,7 @@ func Test_buildHealthCheck(t *testing.T) {
 					},
 				},
 			},
-			errMsgPrefix: fmt.Sprintf("invalid protocol: %q specified in annotation: %q", "invalid", annDOProtocol),
```
timoreimann: Looks like you replaced the test for the case of an invalid health check protocol specified? This is still possible if the user puts in an unsupported protocol in the annotation, isn't it?

Or is the Github diff view tricking me here?

snormore: The `buildHealthCheck` method no longer does a `getProtocol` so this test case wouldn't get hit anymore.

snormore: To clarify, this wasn't a test for invalid health check protocol, this was a test for invalid service protocol.

timoreimann: Got it, missed what this test was really good for. Thanks.

PR #236, opened 2019-07-05
Clean up LB buildForwardingRules a bit

snormore:
Breaks `buildForwardingRules` up with `buildForwardingRule` and `buildTLSForwardingRule`.

Second of 2 PRs to be extracted from https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/231

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -131,6 +131,14 @@ const (
 	// Sticky sessions types.
 	stickySessionsTypeNone    = "none"
 	stickySessionsTypeCookies = "cookies"
+
+	// Protocol values.
+	protocolTCP   = "tcp"
+	protocolHTTP  = "http"
+	protocolHTTPS = "https"
+
+	// Port protocol values.
+	portProtocolTCP = "TCP"
 )
 
 var errLBNotFound = errors.New("loadbalancer not found")
@@ -426,6 +434,8 @@ func buildHealthCheck(service *v1.Service) (*godo.HealthCheck, error) {
 // buildForwardingRules returns the forwarding rules of the Load Balancer of
 // service.
 func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
+	var forwardingRules []godo.ForwardingRule
+
 	protocol, err := getProtocol(service)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -431,57 +433,73 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
...
+	tlsPortMap := map[int32]bool{}
+	for _, port := range tlsPorts {
+		tlsPortMap[int32(port)] = true
+	}
 
-	if len(tlsPorts) > 0 {
-		if certificateID == "" && !tlsPassThrough {
-			return nil, errors.New("must set certificate id or enable tls pass through")
+	for _, port := range service.Spec.Ports {
+		// We use https for TLS, so set it explicitely if necessary.
```
```suggestion
		// We use https for TLS, so set it explicitly if necessary.
```

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -431,57 +433,73 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	certificateID := getCertificateID(service)
-	tlsPassThrough := getTLSPassThrough(service)
-
-	if (certificateID != "" || tlsPassThrough) && len(tlsPorts) == 0 {
+	if len(tlsPorts) == 0 {
```
timoreimann: Strictly speaking, aren't we diverging from the current behavior with this change?

Before, it was possible for users to run TCP or HTTP traffic on port 443 if they did not specify a certificate ID or TLS passthrough. We now seem to assume that port 443 always means HTTPS.

I realize it'd be somewhat weird to run no-TLS traffic on port 443 (maybe except for testing purposes). If it's possible to implement with the vanilla DO LB product, however, then I believe we should continue to allow so as well with our Kubernetes integration.

WDYT?

snormore: Definitely agree, the behaviour should remain the same, will update :+1:

timoreimann: We may also want to add a test for this case as we are seemingly missing it.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -439,60 +446,78 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
 	certificateID := getCertificateID(service)
 	tlsPassThrough := getTLSPassThrough(service)
 
-	if (certificateID != "" || tlsPassThrough) && len(tlsPorts) == 0 {
+	if len(tlsPorts) == 0 && (protocol == protocolHTTPS || certificateID != "" || tlsPassThrough) {
```
timoreimann: Why do we need the check for the HTTPS protocol? Why wouldn't it suffice to check for the other conditions as before?  🤔

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -439,60 +446,78 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
...
+}
 
-		forwardingRule.EntryProtocol = protocol
-		forwardingRule.TargetProtocol = protocol
+func buildForwardingRule(service *v1.Service, port *v1.ServicePort, protocol, certificateID string, tlsPassThrough bool) (*godo.ForwardingRule, error) {
+	var forwardingRule godo.ForwardingRule
 
-		forwardingRule.EntryPort = int(port.Port)
-		forwardingRule.TargetPort = int(port.NodePort)
+	if port.Protocol != "TCP" {
```
```suggestion
	if port.Protocol != protocolTCP {
```

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -439,60 +446,78 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
...
+		return errors.New("must set certificate id or enable tls pass through")
 	}
 
-	return forwardingRules, nil
+	if certificateID != "" && tlsPassThrough {
+		return errors.New("either certificate id should be set or tls pass through enabled, not both")
+	}
+
+	if tlsPassThrough {
+		forwardingRule.TlsPassthrough = tlsPassThrough
```
timoreimann: Could we add a comment here describing why we won't need to set the protocol?

That should also remind me of any necessary adjustments once we introduce HTTP/2.

cloud-controller-manager/do/loadbalancers.go
snormore: This would have to be something like `if strings.ToLower(port.Protocol) != protocolTCP`, WDYT?

timoreimann: Ah true, missed the case sensitivity. The constants are also not really meant to compare for service port protocols, so please ignore this one.

snormore:
@timoreimann comments addressed, thanks :pray:

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -439,60 +449,82 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
...
+		// We don't explicitely set the TargetProtocol here since in buildForwardingRule
```
```suggestion
		// We don't explicitly set the TargetProtocol here since in buildForwardingRule
```

PR #241, opened 2019-07-16
Bump version to v0.1.16

snormore:
Rebased on top of https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/240 until it's merged.


timoreimann:
@snormore could you take a look at #241? I spotted one or two minor issues with the change log that I think we should fix real quick.

snormore:
`e2e` is 🍏

Issue #243, opened 2019-07-17
Provider fake load-balancer implementation


timoreimann:
Similar to our [fake tag service](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/442aa00e4b84a470298488a055b54d7ee4d6c1b7/cloud-controller-manager/do/faketagssvc_test.go), we should also provide a fake load-balancer service. This will make testing much more simpler because testers won't need to think about what each API method should return anymore. Instead, it allows to focus on what load-balancer resources the fake should be populated with initially and leave the rest to the implemented logic.

cpanato:
@timoreimann can I try this one? Anything special I need to know? Thanks

timoreimann:
Heya @cpanato 👋  

I'd be more than happy to let you take a stab at this one.

I think the basic idea is spelled out in this issue's description. My tl;dr is that we shouldn't have to spend so much time programming/configuring the LB test doubles, at least not for happy path testing: the fake should implement the logic and mimic how the public LB API behaves as good as possible. Ideally, we'd also want this better fake framework to support stubbing out errors (though an argument could be made that we should better refer to stubs/mocks for that purpose).

I previously worked on this task but never managed to finish it off. The basic fake implementation is there (along with a means to fake out errors, modelled about client-go's fake implementation), with the major chunk of remaining work being to update the tests (and make any changes to the fake along the way as it makes sense).

You can take a look at my https://github.com/digitalocean/digitalocean-cloud-controller-manager/tree/add-lb-fake branch if you like. There's a ton of WIP commits, with [the delta itself](https://github.com/digitalocean/digitalocean-cloud-controller-manager/compare/add-lb-fake...58a3698246a6cc192fccac9e4b84444659f70781) being fairly small.

I'm always open to considering alternative approaches If you think it makes sense. Happy to discuss here and/or based on a demonstrative PR.

Thanks for the offer to help out, appreciated!

PR #245, opened 2019-07-17
Annotate Service with LB ID and use for lookup

### LINKED ISSUE #224: Renaming load balancers causes a new instance to be created leaving the previous one dangling
Hey,
I encountered issue, with my managed Digital Ocean Kubernetes. I have exposed my ingress controller with Service type LoadBalancer and it all worked fine, till the automatic upgrade was triggered.

I noticed, that during the upgrade new droplets are created and the yaml-definitions are recreated automatically. For the most part it worked very well, my services/deployments were running, but the cloud controller created new LoadBalancer with new IP-address. I noticed, that there is Kubernetes-native way to prevent it, by specifying `loadBalancerIP` in the yaml, but it didn't worked with DO. To fix this issue, I have changed the Service type to NodePort with reserved ports and created the LoadBalancer manually, so when DO triggers an upgrade, it won't create new LoadBalancer with new IP each time.

Hope you could help to solve this issue, since I spent a whole night fixing your "managed" service.

timoreimann:
This change decouples load-balancer names from their identity, thereby permitting renames that do not impact associations to Services any longer.

Instead of the load-balancer name, we use the immutable load-balancer UUID. As a beneficial side-effect, we can eliminate the local load-balancer cache since the DigitalOcean API allows for direct lookups by UUID in O(1) (vs. O(n) that we previously experienced when iterating through all load-balancers existing in a given account).

Load-balancer IDs are added to their corresponding Service objects as annotations. This happens both on load-balancer creates as well as most retrieval operations (e.g., `EnsureLoadBalancer`) in order to annotate pre-existing Services as well. When no load-balancer ID can be found, we fall back to doing lookups by name.

Apologies for the somewhat bloated PR, which is mostly due to the test changes required by both leveraging the LB UUID as well as removing the LB cache. Unfortunately, these two should really go together. #243 would be helpful to have, but I did not want to add yet another refactoring in the same PR.

Additional changes this PR makes:

- Add new getter method `retrieveAndAnnotateLoadBalancer` to retrieve a load-balancer by ID or name (based on the availability of the LB ID annotation), and patch the associated Service object with the found ID.
- Use new getter method in `{Get,Ensure,Update}LoadBalancer` and `EnsureLoadBalancerDeleted`.
- Exclude load-balancer activity status check from `GetLoadBalancer` as this is not the purpose of the method. (It is a pure getter without any logic.)
- Save on API calls in `EnsureLoadBalancer` by leveraging the LoadBalancer value returned by the new getter method.
- Improve clarity/readability of `EnsureLoadBalancer` by introducing a switch case statement.
- Remove the load-balancer cache and reintroduce logic around name-based lookups that existed previously.
- Remove the extra Kubernetes client maintained on the `ResourcesController` in favor of the client already managed by the `resources` struct.
- Update the tagging service to use LB IDs for lookups.
- Annotate errors.

Additional test-related changes:

- Extend and refactor tests.
- Use bare `godo.Client` for tests instead of `godo.NewClient`. The former comes with no initialization whatsoever, thus preventing us from running against the real DigitalOcean API during tests accidentally. Additionally, we also fail hard if a used API is not stubbed out.

Fixes #224

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Changed
 
+* Annotate Service objects by load-balancer UUIDs to enable free LB renames and improve the DO API consumption performance (@timoreimann)
 * Update Kubernetes dependencies to 1.15.0 (@timoreimann)
 * Set default LB health check protocol to TCP if not specified (@snormore)
 * Default to HTTP for sticky sessions if no protocol is defined (@snormore)
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -34,7 +34,7 @@ import (
 const (
 	// annoDOLoadBalancerID is the annotation specifying the load-balancer ID
 	// used to enable fast retrievals of load-balancers from the API by UUID.
-	annoDOLoadBalancerID = "service.k8s.digitalocean.com/load-balancer-id"
+	annoDOLoadBalancerID = "kubernetes.digitalocean.com/load-balancer-id"
 
 	// annDOProtocol is the annotation used to specify the default protocol
 	// for DO load balancers. For ports specified in annDOTLSPorts, this protocol
@@ -215,15 +215,15 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
 
 	var lb *godo.LoadBalancer
 	lb, err = l.retrieveAndAnnotateLoadBalancer(ctx, service)
-	switch {
-	case err == nil:
+	switch err {
+	case nil:
 		// LB existing
 		lb, _, err = l.client.LoadBalancers.Update(ctx, lb.ID, lbRequest)
 		if err != nil {
 			return nil, fmt.Errorf("failed to update load-balancer with ID %s: %s", lb.ID, err)
 		}
 
-	case err == errLBNotFound:
+	case errLBNotFound:
 		// LB missing
 		lb, _, err = l.client.LoadBalancers.Create(ctx, lbRequest)
 		if err != nil {
@@ -354,7 +354,7 @@ func (l *loadBalancers) ensureLoadBalancerIDAnnot(service *v1.Service, lbID stri
 // lbByName gets a DigitalOcean Load Balancer by name. The returned error will
 // be lbNotFound if the load balancer does not exist.
 func (l *loadBalancers) lbByName(ctx context.Context, name string) (*godo.LoadBalancer, error) {
-	lbs, _, err := l.client.LoadBalancers.List(ctx, &godo.ListOptions{})
+	lbs, err := allLoadBalancerList(ctx, l.client)
 	if err != nil {
 		return nil, err
 	}
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -218,9 +218,10 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
 	switch err {
 	case nil:
 		// LB existing
+		lbID := lb.ID
 		lb, _, err = l.client.LoadBalancers.Update(ctx, lb.ID, lbRequest)
 		if err != nil {
-			return nil, fmt.Errorf("failed to update load-balancer with ID %s: %s", lb.ID, err)
+			return nil, fmt.Errorf("failed to update load-balancer with ID %s: %s", lbID, err)
 		}
 
 	case errLBNotFound:
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -32,6 +32,10 @@ import (
 )
 
 const (
+	// annoDOLoadBalancerID is the annotation specifying the load-balancer ID
+	// used to enable fast retrievals of load-balancers from the API by UUID.
+	annoDOLoadBalancerID = "service.k8s.digitalocean.com/load-balancer-id"
+
 	// annDOProtocol is the annotation used to specify the default protocol
 	// for DO load balancers. For ports specified in annDOTLSPorts, this protocol
 	// is overwritten to https. Options are tcp, http and https. Defaults to tcp.
@@ -172,14 +176,12 @@ func newLoadBalancers(resources *resources, client *godo.Client, region string) 
 //
 // GetLoadBalancer will not modify service.
 func (l *loadBalancers) GetLoadBalancer(ctx context.Context, clusterName string, service *v1.Service) (*v1.LoadBalancerStatus, bool, error) {
-	lbName := l.GetLoadBalancerName(ctx, clusterName, service)
-	lb, found := l.resources.LoadBalancerByName(lbName)
-	if !found {
-		return nil, false, nil
-	}
-
-	if lb.Status != lbStatusActive {
-		return nil, true, fmt.Errorf("load-balancer not active, currently %s", lb.Status)
+	lb, err := l.retrieveAndAnnotateLoadBalancer(ctx, service)
+	if err != nil {
+		if err == errLBNotFound {
+			return nil, false, nil
+		}
+		return nil, false, err
 	}
 
 	return &v1.LoadBalancerStatus{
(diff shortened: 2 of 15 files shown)


timoreimann:
Another note: while the cloud provider interface mandates[ that Service objects should not be modified](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/442aa00e4b84a470298488a055b54d7ee4d6c1b7/vendor/k8s.io/cloud-provider/cloud.go#L111), this actually does not hold in practice anymore: it used to be true when the cloud provider framework still used UPDATE requests to add the external load-balancer IP address (ingress status) to the Service object without handling resource conflicts. These days, it employs PATCH semantics as well.

My tests have shown no problems whatsoever. kubernetes/cloud-provider#27 is touching on the matter as well. (See also [my comment there](https://github.com/kubernetes/cloud-provider/issues/27#issuecomment-508269712) and the following discussion.)

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -32,6 +32,10 @@ import (
 )
 
 const (
+	// annoDOLoadBalancerID is the annotation specifying the load-balancer ID
+	// used to enable fast retrievals of load-balancers from the API by UUID.
+	annoDOLoadBalancerID = "service.k8s.digitalocean.com/load-balancer-id"
```
bouk: I think we should use `doks.digitalocean.com/load-balancer-id` here.

docs/getting-started.md
```diff
@@ -75,6 +75,17 @@ The primary purpose of the variable is to allow DigitalOcean customers to easily
...
+### Load-balancer ID annotations
+
+`digitalocean-cloud-controller-manager` attaches the UUID of load-balancers to the corresponding Service objects (given they are of type `LoadBalancer`) using the `service.k8s.digitalocean.com/load-balancer-id` annotation. This serves two purposes:
+
+1. To support load-balancer renames without impacting the correct working of `digitalocean-cloud-controller-manager`.
+2. To efficiently look up load-balancer resources in the DigitalOcean API.
+
+The annotation is added for both newly created and existing load-balancers. However, existing load-balancers lacking the annotation yet can only be associated correctly if they were not renamed before. Otherwise, the renamed load-balancer will be considered missing and a new one will be created instead, leaving the old one dangling.
+
+It is also possible to start owning an unmanaged load-balancer by adding the annotation manually prior to setting the Service type to `LoadBalancer`. Note however every load-balancer should be owned by exactly one `digitalocean-cloud-controller-manager` instance only to prevent concurrent and presumably conflicting modifications.
```
bouk: > You can import an existing Load Balancer by creating the `Service` annotated with the ID of the Load Balancer. Don’t do this while it’s being managed by another cluster, or the `digitalocean-cloud-controller-managers` will make conflicting modifications to the Load Balancer.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -277,25 +282,92 @@ func (l *loadBalancers) UpdateLoadBalancer(ctx context.Context, clusterName stri
...
+	updated := service.DeepCopy()
+	updated.ObjectMeta.Annotations[annoDOLoadBalancerID] = lbID
+
+	return patchService(l.resources.kclient, service, updated)
+}
+
+// lbByName gets a DigitalOcean Load Balancer by name. The returned error will
+// be lbNotFound if the load balancer does not exist.
+func (l *loadBalancers) lbByName(ctx context.Context, name string) (*godo.LoadBalancer, error) {
+	lbs, _, err := l.client.LoadBalancers.List(ctx, &godo.ListOptions{})
```
bouk: I think this needs to paginate, like `allLoadBalancerList`

jcodybaker:
I was a little concerned that this would cause problems with `kubectl replace` since those objects retain the same uuid but are otherwise set exactly to the state of the new object.  However, it looks like that's likely to fail for other reasons and is there likely unused in common practice.
```
$ kubectl apply -f lb.yaml 
service/server created
$ 
$ kubectl annotate service/server 'doks.digitalocean.com/load-balancer-id=1234'
service/server annotated
$ kubectl get service/server -o jsonpath='{.metadata.uid} {"\n"}{.metadata.annotations.doks\.digitalocean\.com/load-balancer-id} {"\n"}'
dbde9d51-a8a4-11e9-bb23-c2cebb59613b 
1234 
$ kubectl replace -f lb.yaml 
The Service "server" is invalid: spec.clusterIP: Invalid value: "": field is immutable
$ kubectl apply -f lb.yaml 
service/server configured
$ kubectl get service/server -o jsonpath='{.metadata.uid} {"\n"}{.metadata.annotations.doks\.digitalocean\.com/load-balancer-id} {"\n"}'
dbde9d51-a8a4-11e9-bb23-c2cebb59613b 
```

cloud-controller-manager/do/resources.go
```diff
@@ -348,8 +261,13 @@ func (r *ResourcesController) syncTags() error {
...
 
-		name := getDefaultLoadBalancerName(svc)
-		if id, ok := loadBalancers[name]; ok {
+		id := getLoadBalancerID(svc)
+		if id == "" {
+			name := getDefaultLoadBalancerName(svc)
+			id = loadBalancerIDsByName[name]
+		}
+
+		if id != "" {
```
jcodybaker: You could either do an `else` here or just put a `continue` at the end of the `if id == "" {}` block.

docs/getting-started.md
```diff
@@ -75,6 +75,17 @@ The primary purpose of the variable is to allow DigitalOcean customers to easily
...
 When a cluster is created in a non-default VPC for the region, the environment variable `DO_CLUSTER_VPC_ID` must be specified or Load Balancer creation for services will fail.
 
+### Load-balancer ID annotations
+
+`digitalocean-cloud-controller-manager` attaches the UUID of load-balancers to the corresponding Service objects (given they are of type `LoadBalancer`) using the `service.k8s.digitalocean.com/load-balancer-id` annotation. This serves two purposes:
+
+1. To support load-balancer renames without impacting the correct working of `digitalocean-cloud-controller-manager`.
+2. To efficiently look up load-balancer resources in the DigitalOcean API.
+
+The annotation is added for both newly created and existing load-balancers. However, existing load-balancers lacking the annotation yet can only be associated correctly if they were not renamed before. Otherwise, the renamed load-balancer will be considered missing and a new one will be created instead, leaving the old one dangling.
```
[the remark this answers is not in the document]
jcodybaker: s/added gets lost/added will be lost/

vendor/modules.txt
```diff
@@ -267,6 +267,7 @@ k8s.io/apiextensions-apiserver/pkg/features
 # k8s.io/apimachinery v0.0.0 => k8s.io/apimachinery v0.0.0-20190612205821-1799e75a0719
 k8s.io/apimachinery/pkg/labels
 k8s.io/apimachinery/pkg/types
+k8s.io/apimachinery/pkg/util/strategicpatch
```
jcodybaker: We're not actually adding/removing anything from the vendor here.  Why did this file change?  Do we want to include these changes?

cloud-controller-manager/do/loadbalancers.go
timoreimann: Replaced by `kubernetes.digitalocean.com/load-balancer-id` as discussed via DM to make this less specific about DOKS. (DIY clusters will use the annotation as well.)

docs/getting-started.md
```diff
@@ -75,6 +75,17 @@ The primary purpose of the variable is to allow DigitalOcean customers to easily
 
 When a cluster is created in a non-default VPC for the region, the environment variable `DO_CLUSTER_VPC_ID` must be specified or Load Balancer creation for services will fail.
 
+### Load-balancer ID annotations
+
```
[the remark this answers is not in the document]
timoreimann: Pretty much all references in CCM use the _load-balancer_ style right now. If we want to adapt the marketing speech, let's do this in a separate PR.

docs/getting-started.md
timoreimann: I went with your suggestion with two exceptions:

- I think "owning" is clearer than "importing"
- You can also own/import a load-balancer after the Service has been created as long as the type isn't set to `LoadBalancer` yet. I think the new wording makes this easy enough to understand now.

cloud-controller-manager/do/loadbalancers.go
timoreimann: Good catch!

Lucky for us, we already had a `allLoadBalancerList` function, so I just used that.

cloud-controller-manager/do/resources.go
timoreimann: We cannot assume that we have an ID for sure at this point: for the case that the user renamed the LB, `loadBalancerIDsByName[name]` won't yield anything. I added a comment to make that clearer.

vendor/modules.txt
timoreimann: I think it changed because we started to use the referenced package as a direct dependency now, whereas it was consumed transitively only before.

docs/getting-started.md
```diff
@@ -75,6 +75,17 @@ The primary purpose of the variable is to allow DigitalOcean customers to easily
...
+`digitalocean-cloud-controller-manager` attaches the UUID of load-balancers to the corresponding Service objects (given they are of type `LoadBalancer`) using the `service.k8s.digitalocean.com/load-balancer-id` annotation. This serves two purposes:
```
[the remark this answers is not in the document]
timoreimann: I still need to address this one.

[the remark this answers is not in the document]
timoreimann: Done.

timoreimann:
@bouk @jcodybaker addressed all comments. PTAL again.

docs/getting-started.md
bouk: Maybe we can just call it ‘managing’ instead of owning, because that’s what you call it in the next sentence too.

> You can also own/import a load-balancer after the Service has been created as long as the type isn't set to LoadBalancer yet

I think the way you’re wording this is confusing, because it makes it sounds like you need to:

1. Create a Service
2. Set the annotation
3. Change the type to LoadBalancer

While this is possible, it’s a very roundabout way of doing it. You can just create a service of type LoadBalancer with the annotation in one go. Perhaps you could say ‘You [...] manage an existing load-balancer by creating a LoadBalancer Service annotated with the UUID of the load-balancer.’

Also `or otherwise` is a tautology.

timoreimann: Adopted your suggestion and rephrased the second sentence to be less tautological.

timoreimann: Just dropped "otherwise" in the last sentence.

cloud-controller-manager/do/resources.go
jcodybaker: Gotcha.  Yeah, now I see this.  Sorry for the unnecessary comment?

timoreimann: Not at all -- it pointed out that the reasoning behind this wasn't clear.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Changed
 
+* Annotate Service objects by load-balancer UUIDs to enable free LB renames and improve the DO API consumption performance (@timoreimann)
```
snormore: Just noticed this got merged into the `v0.1.16` section

timoreimann: Good catch -- submitted #249 to fix.

PR #248, opened 2019-07-18
Fix e2e instructions and add a Docker runner for them

snormore:
Noticed a couple things that were off in the e2e instructions, and also wanted an isolated way to run this in a Docker image.

### DIFF
--- a/e2e/Dockerfile
+++ b/e2e/Dockerfile
@@ -1,0 +1,37 @@
+# Copyright 2017 DigitalOcean
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM golang:1.12
+
+WORKDIR /root
+
+# Install kubectl
+RUN curl -LfO https://storage.googleapis.com/kubernetes-release/release/v1.15.0/bin/linux/amd64/kubectl && \
+    chmod +x ./kubectl && \
+    mv ./kubectl /usr/local/bin/kubectl
+
+# Install kops
+RUN curl -Lfo kops https://github.com/kubernetes/kops/releases/download/$(curl -s https://api.github.com/repos/kubernetes/kops/releases/latest | grep tag_name | cut -d '"' -f 4)/kops-linux-amd64 && \
+    chmod +x ./kops && \
+    mv ./kops /usr/local/bin/
+
+# Install doctl
+ARG DOCTL_VERSION="1.22.0"
+RUN curl -LfO https://github.com/digitalocean/doctl/releases/download/v${DOCTL_VERSION}/doctl-${DOCTL_VERSION}-linux-amd64.tar.gz && \
+    tar xf doctl-${DOCTL_VERSION}-linux-amd64.tar.gz && \
+    mv doctl /usr/local/bin
+
+# Install s3cmd and jq
+RUN apt-get update -qq && \
+    apt-get install -y s3cmd
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -6,10 +6,10 @@ It may be run either locally or on the CI.
 
 ## usage
 
-The easiest way to run the end-to-end test is to execute `e2e.sh`. It requires the following environment variables to be set:
+The easiest way to run the end-to-end test is to execute `docker.sh test`. It requires the following environment variables to be set or available in a `.env` file:
 
 - `DIGITALOCEAN_ACCESS_TOKEN`: the DigitalOcean access token
-- `KOPS_REGION`: the DigitalOcean region (e.g., _SFO2_)
+- `KOPS_REGION`: the DigitalOcean region (e.g., _sfo2_)
 - `S3_ENDPOINT`: the Spaces endpoint; should consist of the `KOPS_REGION` followed by the official Spaces API host (e.g., _sfo2.digitaloceanspaces.com_)
 - `S3_ACCESS_KEY_ID`: the Spaces access key ID
 - `S3_SECRET_ACCESS_KEY`: the Spaces secret access key ID
@@ -25,6 +25,6 @@ The following environment parameters are optional:
 
 ## resource cleanup
 
-The end-to-end tests clean up resources used on the DigitalOcean cloud after themselves on completion (either successful or erroneous). In the case that teardown does not complete for whatever reason (say, because of a crash of the tests or resources being in an irreparable state), the `cleanup-resources.sh` script can be used to remove the resources explicitly.
+The end-to-end tests clean up resources used on the DigitalOcean cloud after themselves on completion (either successful or erroneous). In the case that teardown does not complete for whatever reason (say, because of a crash of the tests or resources being in an irreparable state), the `docker.sh clean <ID> <NAME>` script can be used to remove the resources explicitly.
 
 See the script's header for details.
--- a/e2e/e2e.sh
+++ b/e2e/e2e.sh
@@ -38,5 +38,5 @@ echo "==> installing dependencies..."
   cd "${SCRIPT_DIR}"
   echo "==> running E2E tests..."
   # shellcheck disable=SC2086
-  go test -v -timeout 1h -count 1 -tags integration ${RUN} "./..."
+  GO111MODULE=on go test -mod=vendor -v -timeout 1h -count 1 -tags integration ${RUN} "./..."
 )
--- a/e2e/scripts/setup_cluster.sh
+++ b/e2e/scripts/setup_cluster.sh
@@ -34,7 +34,16 @@ fi
 readonly KUBERNETES_VERSION="$1"
 readonly NUM_NODES="$2"
 
-SSH_PUBLIC_KEYFILE="${SSH_PUBLIC_KEYFILE:-${SCRIPT_DIR}/id_rsa.pub}"
+SSH_PUBLIC_KEYFILE=${SSH_PUBLIC_KEYFILE:-${HOME}/.ssh/id_rsa.pub}
+if [ ! -f ${SSH_PUBLIC_KEYFILE} ]; then
+  echo "${SSH_PUBLIC_KEYFILE} does not exist"
+  exit 1
+fi
+SSH_KEY_FINGERPRINT="$(ssh-keygen -E md5 -lf ${SSH_PUBLIC_KEYFILE} | cut -d " " -f 2 | sed -e "s/^MD5://")"
+if ! doctl compute ssh-key get ${SSH_KEY_FINGERPRINT}; then
+  echo "${SSH_PUBLIC_KEYFILE} (${SSH_KEY_FINGERPRINT}) does not exist in your DO account"
+  exit 1
+fi
 
 ensure_deps
 
--- a/e2e/.dockerignore
+++ b/e2e/.dockerignore
@@ -1,0 +1,1 @@
+kubeconfig-e2e.*
--- a/e2e/.env.sample
+++ b/e2e/.env.sample
@@ -1,0 +1,6 @@
+DIGITALOCEAN_ACCESS_TOKEN=secret
+KOPS_REGION=sfo2
+S3_ENDPOINT=sfo2.digitaloceanspaces.com
+S3_ACCESS_KEY_ID=id
+S3_SECRET_ACCESS_KEY=secret
+KOPS_CLUSTER_NAME=ccme2e.example.com
--- a/e2e/.gitignore
+++ b/e2e/.gitignore
@@ -1,0 +1,3 @@
+# environment variables file
+!.env.sample
+.env*
--- a/e2e/docker.sh
+++ b/e2e/docker.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+
+# Copyright 2017 DigitalOcean
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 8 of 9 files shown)


scripts/e2e.sh
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+docker build -t do-ccm-e2e e2e
+docker run --rm -it -v ~/.ssh:/root/.ssh -v $PWD:/app -w /app --env-file e2e/.env do-ccm-e2e ./e2e/e2e.sh
```
timoreimann: When I run the container as-is, it downloads all dependencies via Go modules. I suppose that's going to happen on every run since we don't mount in the cache?

I'd also like to be able to just use the vendor folder. Adding `-e GO111MODULE` to docker run disables Go modules in the container (as I have that variable set to `off`) but I get a build error afterwards. I suppose it's because the files aren't copied below the `GOPATH` or `/go`?

e2e/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+FROM golang:1.12
+
+# Install kubectl
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.15.0/bin/linux/amd64/kubectl && \
+    chmod +x ./kubectl && \
+    mv ./kubectl /usr/local/bin/kubectl
+
+# Install kops
+RUN curl -Lo kops https://github.com/kubernetes/kops/releases/download/$(curl -s https://api.github.com/repos/kubernetes/kops/releases/latest | grep tag_name | cut -d '"' -f 4)/kops-linux-amd64 && \
```
timoreimann: The latest kops release won't work for us until kubernetes/kops#7221 is merged and backported to the release versions. But even after that happens, the kops release won't contain the CCM manifest of the version under development.

I suppose the best we can do for now is check out the source code of kops in a version that works for us (1.10 right now, soon master once that upstream PR merges), change the built-in CCM manifest to point at something like `digitalocean/digitalocean-cloud-controller-manager:dev` (that would need to correspond to the version we want to test), compile, and use the built binary for our tests.

I'm completely okay with doing that in a follow-up PR -- the containerized end-to-end test execution is already a great step forward.

timoreimann: Update: the kops PR was merged, we could use kops master now.

e2e/scripts/setup_cluster.sh
```diff
@@ -34,7 +34,7 @@ fi
 readonly KUBERNETES_VERSION="$1"
 readonly NUM_NODES="$2"
 
-SSH_PUBLIC_KEYFILE="${SSH_PUBLIC_KEYFILE:-${SCRIPT_DIR}/id_rsa.pub}"
+SSH_PUBLIC_KEYFILE="${SSH_PUBLIC_KEYFILE:-${HOME}/.ssh/id_rsa.pub}"
```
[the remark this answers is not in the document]
snormore: Reasoning here was that the instructions in the README indicates that `${HOME}/.ssh/id_rsa.pub` is the default. I don't have a strong opinion either way TBH, we can make it work both ways.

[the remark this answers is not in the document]
snormore: I'll update to have it generate a one-time e2e-specific SSH key key and push it via doctl for the tests.

scripts/e2e.sh
snormore: I can get this to work by keeping go modules enabled and using `-mod=vendor` :+1:

snormore:
Comments addressed and should be 👌 for another round of 👀 whenever you get a chance.

e2e/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    mv ./kops /usr/local/bin/
+
+# Install doctl
+RUN go get -u github.com/digitalocean/doctl/cmd/doctl
+
+# Install s3cmd and jq
+RUN apt-get update -qq && \
+    apt-get install -y s3cmd jq
+
+CMD ["./e2e/e2e"]
```
timoreimann: I think the file is called `e2e.sh`.

e2e/docker-run.sh
```diff
@@ -0,0 +1,46 @@
...
+        -e DIGITALOCEAN_ACCESS_TOKEN \
+        -e KOPS_REGION \
+        -e S3_ENDPOINT \
+        -e S3_ACCESS_KEY_ID \
+        -e S3_SECRET_ACCESS_KEY \
+        -e KOPS_CLUSTER_NAME \
+        ${DOCKER_RUN_ARGS} \
+        do-ccm-e2e \
+        ./e2e/e2e.sh
+)
```
timoreimann: The script seems almost identical to `docker-clean.sh` except for which script is invoked and from where. Should we use a single docker runner at this point?

(For extra convenience, we could have a very small entry point script (similar to [this one](https://success.docker.com/article/use-a-script-to-initialize-stateful-container-data)) to accept parameters like `e2e` and `cleanup` and calls the appropriate scripts.)

e2e/docker-run.sh
```diff
@@ -0,0 +1,46 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+PROJECT_DIR=$(dirname ${SCRIPT_DIR})
```
timoreimann: A tad simpler might be

```
PROJECT_DIR="${SCRIPT_DIR}/.."
```

e2e/e2e.sh
```diff
@@ -38,5 +40,5 @@ echo "==> installing dependencies..."
   cd "${SCRIPT_DIR}"
   echo "==> running E2E tests..."
   # shellcheck disable=SC2086
-  go test -v -timeout 1h -count 1 -tags integration ${RUN} "./..."
+  go test -mod=vendor -v -timeout 1h -count 1 -tags integration ${RUN} "./..."
```
timoreimann: `-mod=vendor` is only valid when Go modules are used. We should set `GO111MODULE=on` on the invocation to be sure.

e2e/e2e.sh
```diff
@@ -18,6 +18,8 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
+export KOPS_STATE_STORE=${KOPS_STATE_STORE:-"state-${KOPS_CLUSTER_NAME}"}
```
timoreimann: If we want this, then we should check if `KOPS_CLUSTER_NAME` is set and error out if it doesn't. Otherwise, the export is going to fail if `KOPS_CLUSTER_NAME` isn't set due to `nounset`.

I'm wondering though if a default `KOPS_STATE_STORE` makes sense here -- after all, the user needs to make sure it exists, so my thinking is that we could just as well expect them to specify the store name (as we do right now).

Maybe I missed something about your intentions, so please let me know.

e2e/scripts/setup_cluster.sh
```diff
@@ -34,7 +34,19 @@ fi
...
+  SSH_KEY_NAME="do-ccm-e2e"
+  SSH_PUBLIC_KEYFILE="id_rsa.pub"
+  SSH_KEY_ID="$(doctl compute ssh-key list -ojson | jq --arg NAME ${SSH_KEY_NAME} '.[] | select(.name == $NAME) | .id')"
+  if [ "${SSH_KEY_ID}" ]; then
+    doctl compute ssh-key delete ${SSH_KEY_ID} -f
+  fi
+  if [ ! -f ${SSH_PUBLIC_KEYFILE} ]; then
+    ssh-keygen -t rsa -N "" -f id_rsa
+  fi
+  doctl compute ssh-key import ${SSH_KEY_NAME} --public-key-file ${SSH_PUBLIC_KEYFILE}
```
timoreimann: What's the reason we are importing the key?

FWIW, you should be able to log into master and worker nodes without importing it. At least, that's what I tend to do when debugging.

e2e/e2e.sh
snormore: I don't think the user has to make sure this exists on their own (I've never created it myself), but it is checked as required and was erroring previously to this default being set here:
```
    e2e_test.go:57: missing required environment variable(s): [KOPS_STATE_STORE]
```
An alternative is that we set it to a derived default like this in the Go code closer to where it's actually needed.

e2e/scripts/setup_cluster.sh
snormore: The key needs to be available in the users DO account or else we get these errors:
```
W0722 12:04:11.159615    3872 executor.go:130] error running task "Droplet/master-sfo2.masters.ccme2e.snodo.co" (29s remaining to succeed): POST https://api.digitalocean.com/v2/droplets: 422 redacted are invalid key identifiers for Droplet creation.
W0722 12:04:11.159694    3872 executor.go:130] error running task "Droplet/nodes.ccme2e.snodo.co" (29s remaining to succeed): POST https://api.digitalocean.com/v2/droplets: 422 redacted are invalid key identifiers for Droplet creation.
```

e2e/docker-run.sh
snormore: I like the rule of 3 in these situations (abstract at the 3rd instance), but will DRY this up a bit 😄

e2e/e2e.sh
timoreimann: You're right, the variable shouldn't need to be set for the end-to-end tests because [we're making sure it is set](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/47731f5a0c87b53ebb06171e6851b8233e679eca/e2e/cluster_test.go#L78) prior to calling `setup_cluster.sh`. It is only needed when someone invokes that script directly (which I happen to do every now and then).

I think it's okay to just exclude `KOPS_STATE_STORE` from [the set of variables checked for existence](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/47731f5a0c87b53ebb06171e6851b8233e679eca/e2e/e2e_test.go#L51) then in the Go test. The [check we have in `setup_cluster.sh`](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/47731f5a0c87b53ebb06171e6851b8233e679eca/e2e/scripts/setup_cluster.sh#L26) suffices, and so we can drop this extra export. WDYT?

e2e/scripts/setup_cluster.sh
timoreimann: Ah of course, I completely forgot that I added my key a long long time ago. 🤦‍♂

e2e/docker-run.sh
timoreimann: That's actually a very good rule, so I'm also okay with sticking to the duplication for now. 👍

e2e/e2e.sh
snormore: That makes sense to me :+1:

e2e/docker-run.sh
snormore: DRYed this up into just a `e2e/docker.sh test|clean` script.

e2e/scripts/cleanup-resources.sh
```diff
@@ -51,18 +48,8 @@ if ! type s3cmd > /dev/null 2>&1; then
...
-
-if [[ $# -ne 2 ]]; then
-  usage
-  exit 1
-fi
-
-readonly ID="$1"
-readonly DOMAIN="$2"
+readonly ID="${KOPS_CLUSTER_NAME}"
+readonly DOMAIN="${KOPS_CLUSTER_NAME}"
```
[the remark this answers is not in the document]
snormore: I've been using `ccme2e.snodo.co` and everything seems to be getting cleaned up for me. 🤔 

It looks like it's because of:
```
$ doctl compute droplet list --format ID,Tags | grep "ccme2e.snodo.co"
152032167    KubernetesCluster:ccme2e-snodo-co
152032186    KubernetesCluster:ccme2e-snodo-co
152032187    KubernetesCluster:ccme2e-snodo-co
```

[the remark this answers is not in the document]
timoreimann: @snormore hmm how does that `grep` work when there are dots in `ccme2e.snodo.co` but the tags use dashes like `ccme2e-snodo-co`? 🤔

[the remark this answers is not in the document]
snormore: Apparently `.` is the single-character-wildcard in `grep` (TIL) 🤷‍♂

.gitignore
```diff
@@ -6,3 +6,6 @@ generated-secret.yml
 
 # generated e2e artifacts
 e2e/kubeconfig-e2e.*
+
+# ignore SSH key files for e2e tests
+e2e/id_rsa*
```
timoreimann: This can now be removed as we decided to go with `$HOME/.ssh`.

e2e/README.md
```diff
@@ -6,25 +6,25 @@ It may be run either locally or on the CI.
...
+- `KOPS_REGION`: the DigitalOcean region (e.g., _sfo2_)
 - `S3_ENDPOINT`: the Spaces endpoint; should consist of the `KOPS_REGION` followed by the official Spaces API host (e.g., _sfo2.digitaloceanspaces.com_)
 - `S3_ACCESS_KEY_ID`: the Spaces access key ID
 - `S3_SECRET_ACCESS_KEY`: the Spaces secret access key ID
 - `KOPS_CLUSTER_NAME`: The cluster name; needs to be managed by the DigitalOcean nameserver (e.g., _k8stest.mydomain.com_)
 
 The following environment parameters are optional:
 
-- `SSH_PUBLIC_KEYFILE`: path to the public SSH key matching the private key configured on DigitalOcean (default: `$HOME/.ssh/id_rsa.pub`)
+- `SSH_PUBLIC_KEYFILE`: path to the public SSH key matching the private key configured on DigitalOcean (default: `e2e/id_rsa.pub`)
```
timoreimann: This one must be reverted as well.

PR #251, opened 2019-07-19
Use provider IDs for setting LB droplet targets

timoreimann:
Node objects may contain a provider ID of the kind `digitalocean://<droplet ID>`. At times when we need to specify the set of target droplets for load-balancer creates and updates, we now check
for the availablity of provider IDs and use them directly without relying on the local droplets cache.

This change renders the droplets cache obsolete: it was originally introduced to better support accounts with a large number of droplets. For such cases, the provider ID should be injected through the kubelet directly. Consequently, we drop the cache and resort to listing all droplets from the API only when nodes do not come with a provider ID, which should be acceptable for smaller accounts.

### DIFF
--- a/cloud-controller-manager/do/droplets.go
+++ b/cloud-controller-manager/do/droplets.go
@@ -225,5 +225,11 @@ func dropletIDFromProviderID(providerID string) (int, error) {
 		return 0, fmt.Errorf("provider name from providerID should be %s: %s", ProviderName, providerID)
 	}
 
-	return strconv.Atoi(split[2])
+	digits := split[2]
+	dropletID, err := strconv.Atoi(digits)
+	if err != nil {
+		return 0, fmt.Errorf("failed to convert digits %q: %s", digits, err)
+	}
+
+	return dropletID, nil
 }
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -36,9 +36,13 @@ A solution to the problem is to pass the provider ID (aka droplet ID) via the ku
 
 DigitalOcean's managed Kubernetes offering DOKS sets the provider ID on each worker node kubelet instance.
 
-### Kubernetes node names must match the droplet name, private ipv4 ip or public ipv4 ip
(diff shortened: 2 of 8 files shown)


docs/getting-started.md
```diff
@@ -36,9 +36,13 @@ A solution to the problem is to pass the provider ID (aka droplet ID) via the ku
...
+By default, the kubelet will name nodes based on the node's hostname. On DigitalOcean, node hostnames are set based on the name of the droplet. If you decide to override the hostname on kubelets with `--hostname-override`, this will also override the node name in Kubernetes.
+
+Overriding the hostname is okay if provider IDs are injected by the kubelet. (See the previous section.) If that is not the case or there are nodes lacking the provider ID, however, then the Kubenretes node name must match either the droplet name, private ipv4 ip, or the public ipv4 ip. Otherwise, `cloud-controller-manager` won't be able to find the corresponding droplets in the DigitalOcean API and consequently fail to bootstrap nodes.
```
snormore: s/ip/IP

cloud-controller-manager/do/droplets.go
```diff
@@ -225,5 +225,11 @@ func dropletIDFromProviderID(providerID string) (int, error) {
...
 	}
 
-	return strconv.Atoi(split[2])
+	digits := split[2]
+	dropletID, err := strconv.Atoi(digits)
+	if err != nil {
+		return 0, fmt.Errorf("failed to convert digits %q: %s", digits, err)
+	}
+
+	return dropletID, nil
```
bouk: Can you please rewrite this function to use `net/url`, this is so hacky.

timoreimann: Yeah I can -- let me do so in a follow-up PR. It's hacky but functional, and I don't want to risk breaking stuff.

Will put together something right after this one is merged.

timoreimann: Submitted #253.

PR #255, opened 2019-07-29
Support custom annotation to specify HTTP2 ports

### LINKED ISSUE #254: tls-ports annotation is ignored when using http2
Just updated my cluster hopping to switch to http2 but when changing from `http` to `http2` it also switched port 80 to http2 which I don't think is the expected behaviour. 

The docs state...

> service.beta.kubernetes.io/do-loadbalancer-tls-ports
> Specify which ports of the loadbalancer should use the https or http2 protocol.

But the following config produces a LB with http2 as the protocol for both ports 80 and 443...

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    service.beta.kubernetes.io/do-loadbalancer-algorithm: least_connections
    service.beta.kubernetes.io/do-loadbalancer-certificate-id: xxx
    service.beta.kubernetes.io/do-loadbalancer-healthcheck-path: /_health
    service.beta.kubernetes.io/do-loadbalancer-healthcheck-protocol: http
    service.beta.kubernetes.io/do-loadbalancer-protocol: http2
    service.beta.kubernetes.io/do-loadbalancer-redirect-http-to-https: "true"
    service.beta.kubernetes.io/do-loadbalancer-tls-ports: "443"
  labels:
    app: days
spec:
  ports:
  - name: http
    port: 80
    targetPort: 4000
  - name: https
    port: 443
    targetPort: 4000
```

Switching the protocol to `http` has the expected behaviour of port 80 - http, port 443 - https.

timoreimann:
The new annotation enables to run all protocols concurrently by allowing to specify HTTP* ports independently while maintaining backwards compatibility with the existing logic. In particular, port 443 continues to default to HTTPS unless it is explicitly bound to HTTPS.

We also add a guard preventing conflicting port sharing across both the HTTPS and HTTP2 protocols.

Fixes #254

### DIFF
--- a/docs/controllers/services/examples/http2-with-cert-nginx.yml
+++ b/docs/controllers/services/examples/http2-with-cert-nginx.yml
@@ -4,8 +4,7 @@ apiVersion: v1
 metadata:
   name: http2-with-cert-minimal
   annotations:
-    # HTTP2 port annotation for 443 needed since it defaults to HTTPS.
-    service.beta.kubernetes.io/do-loadbalancer-http2-ports: "443"
+    service.beta.kubernetes.io/do-loadbalancer-protocol: "http2"
     service.beta.kubernetes.io/do-loadbalancer-certificate-id: "c019bbf4-cad2-4884-8a20-410ddad7f54a"
 spec:
   type: LoadBalancer
--- a/docs/controllers/services/examples/http2-with-cert-nginx.yml
+++ b/docs/controllers/services/examples/http2-with-cert-nginx.yml
@@ -2,10 +2,11 @@
 kind: Service
 apiVersion: v1
 metadata:
-  name: https-with-cert-minimal
+  name: http2-with-cert-minimal
   annotations:
+    # HTTP2 port annotation for 443 needed since it defaults to HTTPS.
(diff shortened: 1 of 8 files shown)


docs/controllers/services/examples/http2-with-cert-nginx.yml
```diff
@@ -2,10 +2,11 @@
 kind: Service
 apiVersion: v1
 metadata:
-  name: https-with-cert-minimal
+  name: http2-with-cert-minimal
   annotations:
+    # HTTP2 port annotation for 443 needed since it defaults to HTTPS.
+    service.beta.kubernetes.io/do-loadbalancer-http2-ports: "443"
```
snormore: Can you make this a separate example? I think it's still nice to have an example without using the `http2-ports` annotation for the simple case.

timoreimann: Good point.

I changed the example by swapping the HTTP2 port annotation with a protocol annotation as we have the port case covered by the pass-through example already. LMK if that makes sense.

snormore: WFM :+1:

PR #258, opened 2019-07-30
Support LB with status.Hostname instead of status.IP

snormore:
Annotation for specifying the hostname used for the LB `status.Hostname` instead of assigning `status.IP`. This can be used to avoid this [iptables rule being written by kube-proxy](https://github.com/kubernetes/kubernetes/blob/release-1.15/pkg/proxy/iptables/proxier.go#L958-L979) for the IP. This is useful when using SSL or proxy protocol on the LB, which will break if the request doesn't goes through it, but instead is routed directly in-cluster.

## Example

 1. Deploy the manifest below.
 1. Wait for the service external IP to be available.
 1. Add a DNS record for your hostname
 1. Uncomment the hostname annotation in the manifest below. Deploy it.

```yaml
kind: Service
apiVersion: v1
metadata:
  name: hello
  annotations:
    service.beta.kubernetes.io/do-loadbalancer-certificate-id: "1234-5678-9012-3456"
    service.beta.kubernetes.io/do-loadbalancer-protocol: "https"
    # service.beta.kubernetes.io/do-loadbalancer-hostname: "hello.example.com"
spec:
  type: LoadBalancer
  selector:
    app: hello
  ports:
    - name: https
      protocol: TCP
      port: 443
      targetPort: 80
---
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: hello
spec:
  replicas: 3
  template:
    metadata:
      labels:
        app: hello
    spec:
      containers:
      - name: hello
        image: snormore/hello
        ports:
        - containerPort: 80
          protocol: TCP

```

### Before

```
$ k get svc hello
NAME    TYPE           CLUSTER-IP      EXTERNAL-IP      PORT(S)         AGE
hello   LoadBalancer   10.245.83.169   63.133.251.133   443:31962/TCP   3d8h
```

```
root@hello-848657bd59-2j4h2:/go# curl https://hello.example.com
curl: (35) error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol
```

### After

```
$ k get svc hello
NAME    TYPE           CLUSTER-IP      EXTERNAL-IP      PORT(S)         AGE
hello   LoadBalancer   10.245.83.169   hello.example.com   443:31962/TCP   3d8h
```

```
root@hello-848657bd59-2j4h2:/go# curl https://hello.example.com
Hello, you've requested: /
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+* Support LB with status.Hostname instead of status.IP (@snormore)
 * Set default health check protocol to HTTP if health check path is given (@snormore)
 * Support custom annotation to specify HTTP2 ports (@timoreimann)
 * Use provider ID for setting LB droplet targets (@timoreimann)
--- a/docs/controllers/services/annotations.md
+++ b/docs/controllers/services/annotations.md
@@ -64,6 +64,10 @@ Specify whether the DigitalOcean Load Balancer should pass encrypted data to bac
 
 Specifies the certificate ID used for https. To list available certificates and their IDs, install [doctl](https://github.com/digitalocean/doctl) and run `doctl compute certificate list`.
 
+## service.beta.kubernetes.io/do-loadbalancer-hostname
+
+Specifies the hostname used for the Service `status.Hostname` instead of assigning `status.IP` directly. This can be used to workaround the issue of [kube-proxy adding external LB address to node local iptables rule](https://github.com/kubernetes/kubernetes/issues/66607), which will break requests to an LB from in-cluster if the LB is expected to terminate SSL or proxy protocol. See the [examples/README](examples/README.md) for more detail.
+
 ## service.beta.kubernetes.io/do-loadbalancer-algorithm
 
 Specifies which algorithm the Load Balancer should use. Options are `round_robin`, `least_connections`. Defaults to `round_robin`.
--- a/docs/controllers/services/examples/README.md
+++ b/docs/controllers/services/examples/README.md
@@ -98,3 +98,54 @@ ID                                      IP                Name                  
 105c2071-dcfd-479c-8e84-883d66ba4dec    159.203.52.207    a16bdac1c797611e7af69267d6e0a2ca    active    2017-08-05T00:36:16Z    round_robin          tor1             55581290,55581291,55581292    false    type:none,cookie_name:,cookie_ttl_seconds:0    protocol:http,port:31644,path:/,check_interval_seconds:3,response_timeout_seconds:5,healthy_threshold:5,unhealthy_threshold:3    entry_protocol:http,entry_port:80,target_protocol:http,target_port:31644,certificate_id:,tls_passthrough:false entry_protocol:https,entry_port:443,target_protocol:https,target_port:30566,certificate_id:,tls_passthrough:true
 
 ```
+
+## HTTPS or HTTP2 Load Balancer with Hostname
+
+The `service.beta.kubernetes.io/do-loadbalancer-hostname` annotation can be used to specify the hostname used for the Service `status.Hostname` instead of assigning `status.IP` directly. This is useful as a workaround for the issue of [kube-proxy adding external LB address to node local iptables rule](https://github.com/kubernetes/kubernetes/issues/66607), which will break requests to an LB from in-cluster if the LB is expected to terminate SSL or proxy protocol.
+
+The workflow for this scenario is generally:
+
+ 1. Deploy the manifest with your Service (example below).
+ 1. Wait for the service external IP to be available.
+ 1. Add a DNS record for your hostname pointing to the external IP.
+ 1. Add the hostname annotation to your manifest (example below). Deploy it.
+
+```yaml
+kind: Service
+apiVersion: v1
+metadata:
+  name: hello
+  annotations:
+    service.beta.kubernetes.io/do-loadbalancer-certificate-id: "1234-5678-9012-3456"
+    service.beta.kubernetes.io/do-loadbalancer-protocol: "https"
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Support custom annotation to specify HTTP2 ports (@timoreimann)
 * Use provider ID for setting LB droplet targets (@timoreimann)
 * Annotate Service objects by load-balancer UUIDs to enable free LB renames and improve the DO API consumption performance (@timoreimann)
+* Support LB with status.Hostname instead of status.IP (@snormore)
```
timoreimann: Nit: should go at the top.

docs/controllers/services/annotations.md
```diff
@@ -64,6 +64,10 @@ Specify whether the DigitalOcean Load Balancer should pass encrypted data to bac
 
 Specifies the certificate ID used for https. To list available certificates and their IDs, install [doctl](https://github.com/digitalocean/doctl) and run `doctl compute certificate list`.
 
+## service.beta.kubernetes.io/do-loadbalancer-hostname
+
+Specifies the hostname used for the LB `status.Hostname` instead of assigning `status.IP` directly. This can be used to avoid this [iptables rule being written by kube-proxy](https://github.com/kubernetes/kubernetes/blob/release-1.15/pkg/proxy/iptables/proxier.go#L958-L979) for the IP. This is useful when using SSL or proxy protocol on the LB, which will break if the request doesn't goes through it, but instead is routed directly in-cluster.
```
timoreimann: IMO, pointing at the iptables rules is a bit too technical / detailed. I'd keep it at the behavioral level and reference the upstream issue. Suggestion:

```suggestion
Specifies the hostname used for the Service `status.Hostname` instead of assigning `status.IP` directly. This can be used to avoid bypassing the external LB when requests originate at a pod in order to enforce SSL or proxy protocol processing.

See also [Kubernetes issue 66607](github.com/kubernetes/kubernetes/issues/66607).
```

docs/controllers/services/examples/http2-with-cert-nginx.yml
```diff
@@ -6,6 +6,7 @@ metadata:
   annotations:
     service.beta.kubernetes.io/do-loadbalancer-protocol: "http2"
     service.beta.kubernetes.io/do-loadbalancer-certificate-id: "c019bbf4-cad2-4884-8a20-410ddad7f54a"
+    # service.beta.kubernetes.io/do-loadbalancer-hostname: "example.com"
```
timoreimann: How about putting this in an extra example and leaving the annotation uncommented while putting an explanatory comment right above it, e.g.:

```
annotations:
  service.beta.kubernetes.io/do-loadbalancer-protocol: "http2"
  service.beta.kubernetes.io/do-loadbalancer-certificate-id: "c019bbf4-cad2-4884-8a20-410ddad7f54a"	    
  # "example.com" should be configured to point at the IP address of the DO load-balancer
  service.beta.kubernetes.io/do-loadbalancer-hostname: "example.com"
```

docs/controllers/services/examples/README.md
```diff
@@ -98,3 +98,53 @@ ID                                      IP                Name
...
+
+## HTTPS or HTTP2 Load Balancer with Hostname
+
+The `service.beta.kubernetes.io/do-loadbalancer-hostname` annotation can be used to specify the hostname used for the Service `status.Hostname` instead of assigning `status.IP` directly. This is useful as a workaround for the issue of [kube-proxy adding external LB address to node local iptables rule](https://github.com/kubernetes/kubernetes/issues/66607), which will break requests to an LB from in-cluster if the LB is expected to terminate SSL or proxy protocol.
+
+The workflow for this scenario is generally:
+
+ 1. Deploy the manifest with your Service (example below).
+ 1. Wait for the service external IP to be available.
+ 1. Add a DNS record for your hostname pointing to the external IP
```
timoreimann: Nit: trailing period missing.

docs/controllers/services/examples/README.md
```diff
@@ -98,3 +98,53 @@ ID                                      IP                Name
...
+
+```yaml
+kind: Service
+apiVersion: v1
+metadata:
+  name: hello
+  annotations:
+    service.beta.kubernetes.io/do-loadbalancer-certificate-id: "1234-5678-9012-3456"
+    service.beta.kubernetes.io/do-loadbalancer-protocol: "https"
+    # service.beta.kubernetes.io/do-loadbalancer-hostname: "hello.example.com"
```
timoreimann: Can we prepend the commented out line with something like

```
# Uncomment once hello.example.com points to the external IP address of the DO load-balancer.
```

timoreimann:
👍 Ready to squash and merge.

ScopeyNZ:
Hey all. Not sure if this is the best place for this comment. We just spent a bit of time debugging an issue with this workaround, where we needed to communicate externally between two different problems. We eventually discovered that our issue was happening because we needed at least one replica of each pod running on the same node for this to actually work, otherwise we were just getting connection timeouts. Not sure if this is expected, but might be worth adding to the docs if it is.

jcodybaker:
@ScopeyNZ - I suspect you're seeing this issue because your Service resource has `.spec.externalTrafficPolicy` set to "Local".  Setting this value to "Cluster" should let any node respond.

ScopeyNZ:
Thanks for the feedback @jcodybaker. FYI we did have our `externalTrafficPolicy` set to Cluster. We did have it at Local at one point during our debugging though. I'll double check tomorrow if I read the wrong configuration or didn't apply it properly.

ScopeyNZ:
Hey Cody. I just double checked, all the applicable services have `externalTrafficPolicy` set to Cluster or no policy set - I think the default is `Cluster` though.

PR #260, opened 2019-08-01
Default to health check protocol http if health check path is specified

snormore:
In the v0.1.16 release we changed the behaviour slightly around how we set the health check protocol default if not specified explicitly. Previously we would use the service protocol as the default. Currently we default to TCP if not specified explicitly. This breaks the case where a health check path is specified but no health check protocol, because now we get an error back from the DO LB API of `health check path isn't available for TCP`. This PR updates the behaviour so that if a health check path is specified, but no health check protocol, then default to HTTP as the health check protocol.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## unreleased
 
+* Set default health check protocol to HTTP if health check path is given (@snormore)
+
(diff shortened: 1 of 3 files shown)


snormore:
Going to close this as it represents the hotfix off of the `v0.1.16` tag. We'll open another for merging this change into master next.

Issue #262, opened 2019-08-05
Loadbalancer forces http port-forwarding


bjg2:
Hi,

I used these settings for my nginx controller annotations and they were working:
```
                        "annotations": {
                            "service.beta.kubernetes.io/do-loadbalancer-sticky-sessions-type": "cookies",
                            "service.beta.kubernetes.io/do-loadbalancer-sticky-sessions-cookie-name": "vb-do-sticky",
                            "service.beta.kubernetes.io/do-loadbalancer-sticky-sessions-cookie-ttl": "34650",
                            "service.beta.kubernetes.io/do-loadbalancer-enable-proxy-protocol": true
                        }
```

Today I created new cluster, and the same configurations created load balancer with HTTP forwarding instead of TCP:
![image](https://user-images.githubusercontent.com/6639273/62477663-bb39e380-b7a9-11e9-98a5-680a549da8c9.png)

Even though [this page](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/docs/controllers/services/annotations.md) states that default forwarding rule is TCP, I tried to fix the issue by explicitely setting TCP in service annotation, but it still created load balancer with HTTP forwarding rules:
![image](https://user-images.githubusercontent.com/6639273/62477910-356a6800-b7aa-11e9-8027-0ba4b76bc6d4.png)
![image](https://user-images.githubusercontent.com/6639273/62477983-592dae00-b7aa-11e9-8722-4cbd69524ec9.png)

This was not happening on k8s cluster v1.13 few months ago. I created v1.14 today and now I hit the issue. I manually changed forwarding rules to TCP through GUI, can you please take a look to see why this is happening? It is impacting our scripting and automation...

PS. This did not happen for another service of mine, which got correct port forwarding.

Service that had issues:
```
{
    "apiVersion": "v1",
    "kind": "Service",
    "metadata": {
        "annotations": {
            "service.beta.kubernetes.io/do-loadbalancer-enable-proxy-protocol": "true",
            "service.beta.kubernetes.io/do-loadbalancer-protocol": "tcp",
            "service.beta.kubernetes.io/do-loadbalancer-sticky-sessions-cookie-name": "vb-do-sticky",
            "service.beta.kubernetes.io/do-loadbalancer-sticky-sessions-cookie-ttl": "34650",
            "service.beta.kubernetes.io/do-loadbalancer-sticky-sessions-type": "cookies"
        },
[...]

PR #263, opened 2019-08-05
Do not force HTTP with sticky-sessions

snormore:
This PR reverts https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/234 and adds a warning/note about using sticky sessions with only a TCP forwarding rule and how it will not work as expected.


timoreimann:
Refs #262.

PR #264, opened 2019-08-06
Ensure annotation map exists prior to adding LB ID

timoreimann:
Needed to prevent in panic when the service has no annotations whatsoever, which is the default state of annotation-less services.

Also update test service builder to leave annotations uninitialized by default in order to match the production state.

Follow-up of #245 that was discovered during end-to-end test execution.


fatih:
lgtm

PR #268, opened 2019-08-18
Overwrite service load-balancer ID on mismatch

timoreimann:
This change causes the service load-balancer ID annotation to be overwritten if it does not match the given load-balancer ID.

It is required to properly handle the case where the load-balancer is deleted manually: in that case, CCM won't be able to look up the load-balancer and create a new one. Without updating the ID, however, subsequent reconciliations will continue to use the old ID, fail to find one, try to create a new but fail because a load-balancer with the given name (based on the service UUID) already exists. At this point, the reconciliation attempt will loop endlessly.

We need to break the cycle by annotating the ID of the newly created load-balancer.

### DIFF
--- a/cloud-controller-manager/do/loadbalancers_test.go
+++ b/cloud-controller-manager/do/loadbalancers_test.go
@@ -3412,16 +3412,17 @@ func Test_GetLoadBalancer(t *testing.T) {
 
 func Test_EnsureLoadBalancer(t *testing.T) {
 	testcases := []struct {
-		name     string
-		droplets []godo.Droplet
-		getFn    func(context.Context, string) (*godo.LoadBalancer, *godo.Response, error)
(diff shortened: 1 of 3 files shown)


cloud-controller-manager/do/loadbalancers_test.go
```diff
@@ -3691,6 +3694,7 @@ func Test_EnsureLoadBalancer(t *testing.T) {
 					},
 				},
 			},
+			wantLoadBalancerID: "other-load-balancer-id",
```
snormore: How about `newLoadBalancerID` instead of `wantLoadBalancerID`?

timoreimann: Yep, good idea -- done.

timoreimann:
Validated that the PR fixes the outlined use case. (My explanation on how things would go wrong was slightly but the effect [endless reconciliation] still holds; I updated the description.)

Issue #92, opened 2018-05-09
Support Let's Encrypt Certificates for LoadBalancers


andrewsykim:
Need to pull in the latest version of godo that includes https://github.com/digitalocean/godo/pull/165.

xmudrii:
`godo` updated in #93.

I don't believe `digitalocean-cloud-controller-manager` currently supports Certificates (IIRC, you can only uses existing one with ID). So, we'll need to add annotations for Certificate payload (private key, certificate chain, leaf certificate) and type (`custom` or `lets_encrypt`).

DNS names maybe could be obtained from API, not sure are there annotations by default.

andrewsykim:
That is true, the current implementation of LBs require that the certificate was already created and that you know the certificate ID. The new Let's Encrypt feature means that we don't have to have the certificate created, we just need a DigitalOcean domain that we can use. 

One thing to note is that the Let's Encrypt feature currently does not support wildcard domains, meaning that though users don't need to manually create certificates, they still need to manually create a domain. However, we can programmatically create subdomains once we have a domain to work so users only have to do the manual step once.

sonamsamdupkhangsar:
Does DigitalOcean Kubernetes have the ability to use lets encrypt to use a domain name with https? 
There are docs online but not sure if there is an update or examples on how to do this.

EnigmaCurry:
I have a custom integration with traefik+Let's Encrypt documented here -
https://github.com/EnigmaCurry/kubelab

Miniland1333:
![letsdothis](https://user-images.githubusercontent.com/12994749/51502187-d2732f80-1d89-11e9-9a77-1b5ffc525436.PNG)
Just to make sure, are you looking for the ability to utilize Let's Encrypt Certificates on DigitalOcean loadbalancer for Kubernetes? I had been delaying a project of mine because of this, but I found out that it is already baked into the loadbalancer that is pointing at my cluster. If you select HTTPS ingress, it just asks for which certificate to use exactly as it would for a non-kubernetes pointing loadbalancer!

wayne-o:
I have a chicken and egg problem with this that i don't know how to fix unless the LE inttegration is there? 

The LB needs to have the cert ID added as an annotation but I don't know the IP of the LB until it is created.

How are people getting around this?

nickpack:
@wayne-o I gave up on this entirely and set up an nginx ingress controller and cert-manager to handle it instead.

[...]

PR #274, opened 2019-09-04
Gracefully handle DO lets_encrypt cert rotation

waynr:
While DOKS does not officially support Let's Encrypt, it is possible for users to create a DO Let's Encrypt certificate and use that cert UUID for the "service.beta.kubernetes.io/do-loadbalancer-certificate-id" on Services they create in their cluster.

This is problematic because some combination of the Certificates Service and LBaaS will automatically rotate Let's Encrypt certificates which deletes the old Certificate (resulting an invalid UUID on the "service.beta.kubernetes.io/do-loadbalancer-certificate-id" Service annotation) then updates the Loadbalancer to point at a new certificate with CCM none the wiser.

Because of the following facts:

* CCM doesn't "know" about Let's Encrypt certificates
* CCM always attempts to set the cert UUID from the Service on the LoadBalancer when making any kind of update to it
* LBaaS will return HTTP 404 status code when it determines that there is no Certificate pointed to by the given cert UUID

CCM subsequently cannot make any change to LoadBalancers in these scenarios. Actually, this can happen when users specify their own custom Certificate and update a LoadBalancer manually without also updating the Service object, but that is probably more rare than the Let's Encrypt scenario and is arguably the user's responsibility not to shoot themselves in the foot by manual changing a LoadBalancer managed by CCM.

This PR proposes a fix that changes the way CCM behaves when it encounters a Let's Encrypt certificate associated with the LoadBalancer that is different from the one found on the Service annotation; specifically, it makes an effort to update the Service annotation with the correct cert UUID.

Left to do before merging this PR:

* [x] Update broken test cases
* [x] add new test cases to cover the different cert type and existence scenarios
* [x] Consider how to handle the known edge case where the load balancer has a valid LE cert but the user may have specified a new certificate ID manually
  * Internal discussion within the DOKS team has initially proposed that we don't support this use case, but we should consider documenting that a user can't (easily) switch from a lets_encrypt to a user-managed certificate.
  * Use certificate creation datetime to determine precedence when the LB and the service certificate differe and both are valid certs.
* [x] collapse service patching code into single call
* [x] create separate PR for service cert id missing 
* [x] use loadbalancer request creation for validation (use in-memory validation where possible before reaching out to APIs)
* [x] if we can't find the loadbalancer certificate we should not fail immediately but should still attempt to update from service (add test case to validate the correct behavior here)
* [ ] manually create service/lb using k8saas-images PR image

Refs #92

### DIFF
--- a/cloud-controller-manager/do/certificates.go
+++ b/cloud-controller-manager/do/certificates.go
@@ -1,0 +1,7 @@
+package do
+
+const (
+	// DO Certificate types
+	certTypeLetsEncrypt = "lets_encrypt"
+	certTypeCustom      = "custom"
+)
--- a/cloud-controller-manager/do/certificates.go
+++ b/cloud-controller-manager/do/certificates.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2017 DigitalOcean
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package do
 
 const (
--- a/cloud-controller-manager/do/certificates_test.go
+++ b/cloud-controller-manager/do/certificates_test.go
@@ -19,6 +19,7 @@ package do
 import (
 	"context"
 	"errors"
+	"fmt"
 	"reflect"
 	"testing"
 
@@ -107,6 +108,76 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 					},
 				},
 			},
+		},
+		{
+			name: "validate behavior when LB cert ID and service cert ID match and correspond to non-existent cert",
+			fakeLBServiceFn: func() fakeLBService {
+				s := defaultLBService
+				s.getFn = func(context.Context, string) (*godo.LoadBalancer, *godo.Response, error) {
+					return createHTTPSLB(443, 30000, "test-lb-id", "test-cert-id"), newFakeOKResponse(), nil
+				}
+				return s
+			},
+			fakeCertServiceFn: func() fakeCertService {
+				s := defaultCertService
+				s.getFn = func(ctx context.Context, certID string) (*godo.Certificate, *godo.Response, error) {
+					return nil, nil, newFakeNotFoundErrorResponse()
+				}
+				return s
+			},
+			service: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+					UID:  "foobar123",
+					Annotations: map[string]string{
+						annDOProtocol:      "http",
+						annDOCertificateID: "test-cert-id",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Ports: []v1.ServicePort{
+						{
+							Name:     "test",
+							Protocol: "TCP",
+							Port:     int32(443),
+							NodePort: int32(30000),
+						},
+					},
+				},
+			},
+			expectedServiceCertID: "test-cert-id",
+			err:                   fmt.Errorf("the %q service annotation refers to nonexistent DO Certificate %q", annDOCertificateID, "test-cert-id"),
+		},
+		{
+			name: "validate behavior when LB cert ID and service cert ID match ",
+			fakeLBServiceFn: func() fakeLBService {
+				s := defaultLBService
+				s.getFn = func(context.Context, string) (*godo.LoadBalancer, *godo.Response, error) {
+					return createHTTPSLB(443, 30000, "test-lb-id", "test-cert-id"), newFakeOKResponse(), nil
+				}
+				return s
+			},
+			service: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+					UID:  "foobar123",
+					Annotations: map[string]string{
+						annDOProtocol:      "http",
+						annDOCertificateID: "test-cert-id",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Ports: []v1.ServicePort{
+						{
+							Name:     "test",
+							Protocol: "TCP",
+							Port:     int32(443),
+							NodePort: int32(30000),
+						},
+					},
+				},
+			},
+			expectedServiceCertID: "test-cert-id",
 		},
 	}
 
--- a/cloud-controller-manager/do/certificates_test.go
+++ b/cloud-controller-manager/do/certificates_test.go
@@ -176,6 +176,19 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 			expectedServiceCertID: "test-cert-id",
 			expectedLBCertID:      "test-cert-id",
 		},
+		{
+			name: "[lets_encrypt] LB cert ID does not exit and service cert ID does",
+			setupFn: func(lbService fakeLBService, certService kvCertService) *v1.Service {
+				lb, _ := createHTTPSLB(443, 30000, "test-lb-id", "test-cert-id", certTypeLetsEncrypt)
+				lbService.store[lb.ID] = lb
+
+				service, cert := createServiceAndCert(lb.ID, "service-cert-id", certTypeLetsEncrypt)
+				certService.store[cert.ID] = cert
+				return service
+			},
+			expectedServiceCertID: "service-cert-id",
+			expectedLBCertID:      "service-cert-id",
+		},
 
 		// custom test cases
 		{
@@ -227,6 +240,19 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 			expectedServiceCertID: "service-cert-id",
 			expectedLBCertID:      "test-cert-id",
 			err:                   fmt.Errorf("the %q service annotation refers to nonexistent DO Certificate %q", annDOCertificateID, "service-cert-id"),
+		},
+		{
+			name: "[custom] LB cert ID does not exit and service cert ID does",
+			setupFn: func(lbService fakeLBService, certService kvCertService) *v1.Service {
+				lb, _ := createHTTPSLB(443, 30000, "test-lb-id", "test-cert-id", certTypeCustom)
+				lbService.store[lb.ID] = lb
+
+				service, cert := createServiceAndCert(lb.ID, "service-cert-id", certTypeLetsEncrypt)
+				certService.store[cert.ID] = cert
+				return service
+			},
+			expectedServiceCertID: "service-cert-id",
+			expectedLBCertID:      "service-cert-id",
 		},
 	}
 
--- a/cloud-controller-manager/do/certificates_test.go
+++ b/cloud-controller-manager/do/certificates_test.go
@@ -18,7 +18,6 @@ package do
 
 import (
 	"context"
-	"fmt"
 	"reflect"
 	"testing"
 
@@ -140,17 +139,6 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 			expectedLBCertID:      "test-cert-id",
 		},
 		{
-			name: "[letsencrypt] LB cert ID and service cert ID match and correspond to non-existent cert",
-			setupFn: func(lbService fakeLBService, certService kvCertService) *v1.Service {
-				lb, cert := createHTTPSLB(443, 30000, "test-lb-id", "test-cert-id", certTypeLetsEncrypt)
-				lbService.store[lb.ID] = lb
-				return createServiceWithCert(lb.ID, cert.ID)
-			},
-			expectedLBCertID:      "test-cert-id",
-			expectedServiceCertID: "test-cert-id",
-			err:                   fmt.Errorf("the %q service annotation refers to nonexistent DO Certificate %q", annDOCertificateID, "test-cert-id"),
-		},
-		{
 			name: "[letsencrypt] LB cert ID and service cert ID differ and both certs exist",
 			setupFn: func(lbService fakeLBService, certService kvCertService) *v1.Service {
 				lb, cert := createHTTPSLB(443, 30000, "test-lb-id", "test-cert-id", certTypeLetsEncrypt)
@@ -203,17 +191,6 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 			expectedLBCertID:      "test-cert-id",
 		},
 		{
-			name: "[custom] LB cert ID and service cert ID match and correspond to non-existent cert",
-			setupFn: func(lbService fakeLBService, certService kvCertService) *v1.Service {
-				lb, cert := createHTTPSLB(443, 30000, "test-lb-id", "test-cert-id", certTypeCustom)
-				lbService.store[lb.ID] = lb
-				return createServiceWithCert(lb.ID, cert.ID)
-			},
-			expectedLBCertID:      "test-cert-id",
-			expectedServiceCertID: "test-cert-id",
-			err:                   fmt.Errorf("the %q service annotation refers to nonexistent DO Certificate %q", annDOCertificateID, "test-cert-id"),
-		},
-		{
 			name: "[custom] LB cert ID and service cert ID differ and both certs exist",
 			setupFn: func(lbService fakeLBService, certService kvCertService) *v1.Service {
 				lb, cert := createHTTPSLB(443, 30000, "test-lb-id", "test-cert-id", certTypeCustom)
@@ -226,20 +203,6 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 			},
 			expectedServiceCertID: "service-cert-id",
 			expectedLBCertID:      "service-cert-id",
-		},
-		{
-			name: "[custom] LB cert ID exists and service cert ID does not",
-			setupFn: func(lbService fakeLBService, certService kvCertService) *v1.Service {
-				lb, cert := createHTTPSLB(443, 30000, "test-lb-id", "test-cert-id", certTypeCustom)
-				lbService.store[lb.ID] = lb
-				certService.store[cert.ID] = cert
-
-				service, _ := createServiceAndCert(lb.ID, "service-cert-id", certTypeCustom)
-				return service
-			},
-			expectedServiceCertID: "service-cert-id",
-			expectedLBCertID:      "test-cert-id",
-			err:                   fmt.Errorf("the %q service annotation refers to nonexistent DO Certificate %q", annDOCertificateID, "service-cert-id"),
 		},
 		{
 			name: "[custom] LB cert ID does not exit and service cert ID does",
--- a/cloud-controller-manager/do/certificates_test.go
+++ b/cloud-controller-manager/do/certificates_test.go
@@ -276,9 +276,7 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 				}
 
 				if !reflect.DeepEqual(err, test.err) {
-					t.Error("error does not match test case expectation")
-					t.Logf("expected: %v", test.err)
-					t.Logf("actual: %v", err)
+					t.Errorf("got error %q, want: %q", err, test.err)
 				}
 
 				service, err = fakeResources.kclient.CoreV1().Services(service.Namespace).Get(service.Name, metav1.GetOptions{})
@@ -288,7 +286,7 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 
 				serviceCertID := getCertificateID(service)
 				if test.expectedServiceCertID != serviceCertID {
-					t.Errorf("got service certificate ID: %s, wanted: %s", test.expectedServiceCertID, serviceCertID)
+					t.Errorf("got service certificate ID: %s, want: %s", test.expectedServiceCertID, serviceCertID)
 				}
 
 				godoLoadBalancer, _, err := lbService.Get(context.Background(), getLoadBalancerID(service))
@@ -297,7 +295,7 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 				}
 				lbCertID := getCertificateIDFromLB(godoLoadBalancer)
 				if test.expectedLBCertID != lbCertID {
-					t.Errorf("got load-balancer certificate ID: %s, wanted: %s", test.expectedLBCertID, lbCertID)
+					t.Errorf("got load-balancer certificate ID: %s, want: %s", test.expectedLBCertID, lbCertID)
 				}
 			})
 		}
--- a/cloud-controller-manager/do/certificates_test.go
+++ b/cloud-controller-manager/do/certificates_test.go
@@ -89,7 +89,7 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 		err                   error
 	}{
 		{
-			name: "validate default test values, tls not enabled",
+			name: "default test values, tls not enabled",
 			service: &v1.Service{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: "test",
--- a/cloud-controller-manager/do/certificates_test.go
+++ b/cloud-controller-manager/do/certificates_test.go
@@ -95,7 +95,8 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 					Name: "test",
 					UID:  "foobar123",
 					Annotations: map[string]string{
-						annDOProtocol: "http",
+						annoDOLoadBalancerID: "load-balancer-id",
+						annDOProtocol:        "http",
 					},
 				},
 				Spec: v1.ServiceSpec{
@@ -111,7 +112,40 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 			},
 		},
 		{
-			name: "validate behavior when LB cert ID and service cert ID match and correspond to non-existent cert",
+			name: "[letsencrypt] LB cert ID and service cert ID match ",
+			fakeLBServiceFn: func() fakeLBService {
+				s := defaultLBService
+				s.getFn = func(context.Context, string) (*godo.LoadBalancer, *godo.Response, error) {
+					return createHTTPSLB(443, 30000, "test-lb-id", "test-cert-id"), newFakeOKResponse(), nil
+				}
+				return s
+			},
+			service: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+					UID:  "foobar123",
+					Annotations: map[string]string{
+						annDOProtocol:        "http",
+						annoDOLoadBalancerID: "test-lb-id",
+						annDOCertificateID:   "test-cert-id",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Ports: []v1.ServicePort{
+						{
+							Name:     "test",
+							Protocol: "TCP",
+							Port:     int32(443),
+							NodePort: int32(30000),
+						},
+					},
+				},
+			},
+			expectedServiceCertID: "test-cert-id",
+			expectedLBCertID:      "test-cert-id",
+		},
+		{
+			name: "[letsencrypt] LB cert ID and service cert ID match and correspond to non-existent cert",
 			fakeLBServiceFn: func() fakeLBService {
 				s := defaultLBService
 				s.getFn = func(context.Context, string) (*godo.LoadBalancer, *godo.Response, error) {
@@ -131,8 +165,9 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 					Name: "test",
 					UID:  "foobar123",
 					Annotations: map[string]string{
-						annDOProtocol:      "http",
-						annDOCertificateID: "test-cert-id",
+						annDOProtocol:        "http",
+						annoDOLoadBalancerID: "test-lb-id",
+						annDOCertificateID:   "test-cert-id",
 					},
 				},
 				Spec: v1.ServiceSpec{
@@ -151,7 +186,7 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 			err:                   fmt.Errorf("the %q service annotation refers to nonexistent DO Certificate %q", annDOCertificateID, "test-cert-id"),
 		},
 		{
-			name: "validate behavior when LB cert ID and service cert ID match ",
+			name: "[letsencrypt] LB cert ID and service cert ID differ and both certs exist",
 			fakeLBServiceFn: func() fakeLBService {
 				s := defaultLBService
 				s.getFn = func(context.Context, string) (*godo.LoadBalancer, *godo.Response, error) {
@@ -164,8 +199,9 @@ func Test_LBaaSCertificateScenarios(t *testing.T) {
 					Name: "test",
 					UID:  "foobar123",
 					Annotations: map[string]string{
-						annDOProtocol:      "http",
-						annDOCertificateID: "test-cert-id",
+						annDOProtocol:        "http",
+						annoDOLoadBalancerID: "test-lb-id",
+						annDOCertificateID:   "test-cert-id-2",
 					},
 				},
 				Spec: v1.ServiceSpec{
--- a/cloud-controller-manager/do/certificates_test.go
+++ b/cloud-controller-manager/do/certificates_test.go
@@ -18,7 +18,6 @@ package do
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"reflect"
 	"testing"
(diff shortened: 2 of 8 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -275,6 +274,57 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
...
+	checkServiceCertID := true
+	serviceCertID := getCertificateID(service)
+	if lbCertID != "" && lbCertID != serviceCertID {
+		lbCert, _, err := l.resources.gclient.Certificates.Get(ctx, lbCertID)
+		if err != nil {
+			return fmt.Errorf("failed to get current certificate for lb")
+		}
+
+		if lbCert.Type == "lets_encrypt" {
+			l.ensureCertificateIDAnnot(service, lbCertID)
```
snormore: Should this set a the new `serviceCertID = getCertificateID(service)` here too?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -275,6 +274,57 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
...
+
+	checkServiceCertID := true
+	serviceCertID := getCertificateID(service)
+	if lbCertID != "" && lbCertID != serviceCertID {
+		lbCert, _, err := l.resources.gclient.Certificates.Get(ctx, lbCertID)
+		if err != nil {
+			return fmt.Errorf("failed to get current certificate for lb")
+		}
+
+		if lbCert.Type == "lets_encrypt" {
```
snormore: Pull this out into a const?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -275,6 +274,57 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
...
+			l.ensureCertificateIDAnnot(service, lbCertID)
+			checkServiceCertID = false
+		}
+	}
+
+	if checkServiceCertID {
+		_, _, err := l.resources.gclient.Certificates.Get(ctx, serviceCertID)
+		if err != nil {
+			respErr, ok := err.(*godo.ErrorResponse)
+			if ok && respErr.Response.StatusCode == 404 {
```
snormore: s/`404`/`http.StatusNotFound`

cloud-controller-manager/do/loadbalancers.go
waynr: It's not strictly necessary; at the line just below this we set `checkServiceCertID` to `false` and the block conditioned on that variable is the only place where the service cert id is used.

That being said, I can see how someone else coming at this code later on might be confused so it probably does make sense to set it to over write the old value with that of `lbCertID`.

waynr: Alternatively, it `getCertificateID` is a fairly cheap call that just indexes into a map to return a value...while I know next to nothing about the go compiler or compilers in general, i wouldn't be surprised to learn functions like that are simply inlined... [the doc on go compiler optimizations suggests this is probably the case](https://github.com/golang/go/wiki/CompilerOptimizations#function-inlining)

So where I'm going with this is, I think I'll just replace all `serviceCertID` instances with the function call and avoid the potential for confusion altogether.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -275,6 +274,56 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
...
+		_, _, err := l.resources.gclient.Certificates.Get(ctx, getCertificateID(service))
+		if err != nil {
+			respErr, ok := err.(*godo.ErrorResponse)
+			if ok && respErr.Response.StatusCode == http.StatusNotFound {
+				fmtStr := "certificate UUID %q not found; the %q service annotation refers to nonexistent DO Certificate"
+				respErr.Message = fmt.Sprintf(fmtStr, getCertificateID(service), annDOCertificateID)
+			}
+			return nil, err
+		}
+	}
```
timoreimann: I think the logic around certificate validation should be moved into a dedicated function. AFAICS, that would also allow us to replace the `checkServiceCertID` variable / check by a premature return.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -275,6 +274,56 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
...
+
+	if checkServiceCertID {
+		_, _, err := l.resources.gclient.Certificates.Get(ctx, getCertificateID(service))
+		if err != nil {
+			respErr, ok := err.(*godo.ErrorResponse)
+			if ok && respErr.Response.StatusCode == http.StatusNotFound {
+				fmtStr := "certificate UUID %q not found; the %q service annotation refers to nonexistent DO Certificate"
+				respErr.Message = fmt.Sprintf(fmtStr, getCertificateID(service), annDOCertificateID)
```
timoreimann: I think we shouldn't overwrite the error message from an error we do not own. Creating a new error seems like the better option to me.

Doing so would also allow us to replace the type assertion by checking the response directly like

```
if resp != nil && resp.StatusCode == http.StatusNotFound {
  return fmt.Errorf(...)
}
```

(This would replace line 310 forward.)

cloud-controller-manager/do/loadbalancers.go
timoreimann: The extra lookup is certainly not going to affect us performance-wise.

cloud-controller-manager/do/certificates.go
```diff
@@ -0,0 +1,7 @@
+package do
```
timoreimann: You need to add the licence header to this file as well.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -290,12 +339,9 @@ func (l *loadBalancers) UpdateLoadBalancer(ctx context.Context, clusterName stri
 		return err
 	}
 
-	_, _, err = l.resources.gclient.LoadBalancers.Update(ctx, lb.ID, lbRequest)
-	if err != nil {
-		return fmt.Errorf("failed to update load-balancer with ID %s: %s", lb.ID, err)
```
timoreimann: The intention here is to annotate the error because the upstream error itself does not provide enough context (at least it didn't last time I checked).

cloud-controller-manager/do/loadbalancers_test.go
```diff
@@ -3411,13 +3411,24 @@ func Test_GetLoadBalancer(t *testing.T) {
...
+	}
+	defaultGetCertFn := getGetCertFnFn(certTypeCustom)
 	testcases := []struct {
 		name              string
 		droplets          []godo.Droplet
 		getFn             func(context.Context, string) (*godo.LoadBalancer, *godo.Response, error)
 		listFn            func(context.Context, *godo.ListOptions) ([]godo.LoadBalancer, *godo.Response, error)
 		createFn          func(context.Context, *godo.LoadBalancerRequest) (*godo.LoadBalancer, *godo.Response, error)
 		updateFn          func(ctx context.Context, lbID string, lbr *godo.LoadBalancerRequest) (*godo.LoadBalancer, *godo.Response, error)
+		getCertFn         func(context.Context, string) (*godo.Certificate, *godo.Response, error)
```
timoreimann: We never seem to set this in any test case?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -275,6 +274,56 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
...
+			// need to grab the first one we find
+			break
+		}
+	}
+
+	checkServiceCertID := true
+	if lbCertID != "" && lbCertID != getCertificateID(service) {
+		lbCert, _, err := l.resources.gclient.Certificates.Get(ctx, lbCertID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get current certificate for lb")
```
timoreimann: This one should include the `err`.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -275,6 +274,57 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
...
+		if rule.TlsPassthrough {
+			continue
+		}
+		if rule.CertificateID != "" {
+			lbCertID = rule.CertificateID
+			// it's not clear that it is even possible for different forwarding rules
+			// on the same loadbalancer to have different certificates associated
+			// with them -- as far as we are concerned where a load balancer is
+			// created by CCM in the first place it doesn't seem possible so we only
+			// need to grab the first one we find
```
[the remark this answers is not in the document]
waynr: > The LB API supports different certificates in different forwarding rules;

Yeah, my comment is intended to explain why we are only paying attention to the first one found.

> terms of what is possible

So just to be clear, you want me to clarify in my comment (which already frames it in terms of what is possible for a CCM-managed LB) that CCM support for multiple TLS serving certificates may be possible in the future?

cloud-controller-manager/do/certificates.go
waynr: doh, thanks for the reminder

cloud-controller-manager/do/loadbalancers.go
waynr: I replaced the `l.resources.gclient.LoadBalancers.Update` and subsequent error check with a call to `l.getLoadBalancer` which [returns an identically-formatted error message when the update fails](https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/274/files#diff-0f171474489b3a27add24a94c0da1505R321)

cloud-controller-manager/do/loadbalancers.go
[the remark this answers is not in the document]
waynr: > I think it's not hurtful but it indicates that the design could be improved.

I don't think it's hurtful either, given that updating the service annotation is going to be a relatively rare edge case. If I remember correctly, Let's Encrypt certificates have a default lifetime of 3 months which is about how often we would end up updating the cert id annotation here.

I think additional effort to reduce the number of resource updates we make when handling this particular edge case would be more hurtful in terms of the opportunity cost of the time spent.

cloud-controller-manager/do/loadbalancers.go
waynr: My inclination is to factor out functions in the following situations:

* there is an immediate need to reuse a snippet of code
* there will obivously be need to reuse a snippet of code
* the function where the snippet currently resides is getting too unwieldy

I don't think any of those criteria apply in this situation, but I'll take a look once I've had an opportunity to add some test cases to validate what is here so far.

timoreimann: Code fluency and natural scoping of functionality are additional concerns to me. Right now, `updateLoadBalancer` consists of "do stuff around certificates" and the actual update. The former isn't obvious on first sight, especially since it's not core business to updating.

I also think the code would be easier to understand as we could get rid of `checkServiceCertID`: No need to set and check conditionals when a `return nil` has the same effect.

cloud-controller-manager/do/loadbalancers_test.go
waynr: This is only intended to keep the existing tests from breaking with the addition of `kvCertService` to the `newFakeClient` signature.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -184,8 +197,19 @@ func newLoadBalancers(resources *resources, client *godo.Client, region string)
 // GetLoadBalancer returns the *v1.LoadBalancerStatus of service.
 //
 // GetLoadBalancer will not modify service.
-func (l *loadBalancers) GetLoadBalancer(ctx context.Context, clusterName string, service *v1.Service) (*v1.LoadBalancerStatus, bool, error) {
-	lb, err := l.retrieveAndAnnotateLoadBalancer(ctx, service)
+func (l *loadBalancers) GetLoadBalancer(ctx context.Context, clusterName string, service *v1.Service) (lbs *v1.LoadBalancerStatus, truthy bool, err error) {
+	patcher := newServicePatcher(service)
```
waynr: @timoreimann @snormore here is an example of my solution for aggregating in-memory service modifications into a single PATCH call to the kubernetes api. The reason I chose to defer the Patch call is to reduce the chances of service changes happening that don't get captured before return. If you think that's too convoluted I'm fine identifying a more appropriate place to call `patcher.Patch`.

One thing that's worth noting is that in the defer methods as written if the function returns an error then we currently skip the patch. Do you think we should instead unconditionally always patch? I'm not certain myself which would be best.

waynr:
@timoreimann I've moved the service certificate check into a separate PR: https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/275

cloud-controller-manager/do/loadbalancers.go
timoreimann: In general, I like the defer approach a lot. I'll comment around a few design details on more specific lines, but overall it seems like a good approach to ensure we include all patch-worthy annotations in a single invocation.

Yes, I think we should patch independently of whether we hit an error or not. That's what we did before, and the reason is this: even if our request to the LB API fails overall, the kind of patches we intend to do (LB ID / certificate updates) will help us avoid unnecessary follow-up API requests. For example, if we did not have the LB ID recorded in the Service yet and we only failed in a final LB update API request, persisting the LB ID annotation already would ensure that our API requests we executed in the controller-driven retry attempts could fetch the LB object efficiently in O(1) (vs. having to iterate through all LBs exhaustively in the worst case if the LB UUID was still lacking).

Similarly for the certificate ID where we could avoid reaching out to the Certificate API as soon as the certificate IDs matched.

The only thing we need to do then is to short-circuit our `Patch` call if the annotations match. That way, we don't even reach out to the API server if there's nothing worthwhile patching (and thus rule out a potential source of I/O erroring). Would appreciate if you could do the necessary adjustments. (We can copy [what upstream does](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/67c685a86ca3b54897fa4f835c0e652844466a9a/vendor/k8s.io/kubernetes/pkg/controller/service/service_controller.go#L484-L486), which is to invoke `reflect.DeepCopy` on the annotations.)

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -184,8 +197,19 @@ func newLoadBalancers(resources *resources, client *godo.Client, region string)
...
+	defer func() {
+		if err == nil {
+			err = patcher.Patch(l.resources.kclient, service)
+			if err != nil {
+				err = fmt.Errorf("failed to patch service: %s", err)
```
timoreimann: `patcher.Patch` returns annotated errors (that also include the Service name and namespace which this error currently doesn't), so we should return a bare `err` from that method call to avoid stuttering.

timoreimann: If we go with patching regardless of the outer error (i.e., the value of `err` in line 203), we'll have to make sure that both an outer non-nil error and any non-nil error returned from `patcher.Patch` are propagated to the caller.

A convenient way to do that is to use [apimachinery's `Aggregate`](https://github.com/kubernetes/apimachinery/blob/release-1.15/pkg/util/errors/errors.go) helper: it basically aggregates (or wraps) multiple error values into a single error type that handles `nil` values appropriately and returns a single string representation of the errors. The usage in our case would look like this:

```go
import utilerrors "k8s.io/apimachinery/util/errors"
[...]

defer func() {
  perr := patcher.Patch(l.resources.kclient, service)
  errs := []errors{err, perr} // Errors can be nil; utilerrors.Aggregate will handle it correctly.
  err = utilerrors.NewAggregate(errs)
}
```

If you're curious, check out the [`Error()` implementation](https://github.com/kubernetes/apimachinery/blob/release-1.15/pkg/util/errors/errors.go#L64) to see what a string representation would look like.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -171,6 +172,18 @@ type loadBalancers struct {
...
+type servicePatcher struct {
+	base *v1.Service
+}
+
+func newServicePatcher(base *v1.Service) servicePatcher {
+	return servicePatcher{base: base.DeepCopy()}
+}
+
+func (sp *servicePatcher) Patch(kclient kubernetes.Interface, updated *v1.Service) error {
+	return patchService(kclient, sp.base, updated)
```
timoreimann: `updated` is always pointing at the same `base` Service object we received prior to deep-copying it, right?

If that's the case, then we could also a field

```
updated *v1.Service
``` 

to `servicePatcher` and set it to `base` (again, prior to deep-copying), thereby saving on passing through `updated` here.

timoreimann: Nit: Similarly, we could instantiate `servicePatcher` with a `kubernetes.Interface`.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -275,27 +306,91 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
...
+func (l *loadBalancers) UpdateLoadBalancer(ctx context.Context, clusterName string, service *v1.Service, nodes []*v1.Node) (err error) {
+	patcher := newServicePatcher(service)
+	defer func() {
+		if err == nil {
+			err = patcher.Patch(l.resources.kclient, service)
+			if err != nil {
+				err = fmt.Errorf("failed to patch service: %s", err)
+			}
+		}
+	}()
```
timoreimann: Since we have implemented the deferred method three times, how about attaching it to `servicePatcher` so that we can do a direct `defer.Patch(...)` call (and move the error checking / handling into the method too)?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -360,20 +453,11 @@ func (l *loadBalancers) retrieveLoadBalancer(ctx context.Context, service *v1.Se
...
-	if val := getLoadBalancerID(service); val == lbID {
-		return nil
-	}
-
-	// Make a copy so we don't mutate the shared informer cache from the cloud
-	// provider framework.
-	updated := service.DeepCopy()
-	if updated.ObjectMeta.Annotations == nil {
-		updated.ObjectMeta.Annotations = map[string]string{}
+func ensureServiceAnnotation(service *v1.Service, annotName, annotValue string) {
```
timoreimann: `ensure*` usually indicates that a piece of logic changes something or makes sure that the change is already applied. `ensureServiceAnnotation` changed to simply recording a service annotation, so I'd opt to rename it to something like `recordServiceAnnotation`.

cloud-controller-manager/do/certificates_test.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+func (f *kvCertService) Get(ctx context.Context, certID string) (*godo.Certificate, *godo.Response, error) {
+	if f.reflectorMode {
+		return &godo.Certificate{
+			ID:   certID,
+			Type: certTypeLetsEncrypt,
+		}, newFakeOKResponse(), nil
+	}
+	lb, ok := f.store[certID]
```
timoreimann: copy-pasta: `lb` should be called `certificate` or `cert`.

cloud-controller-manager/do/certificates_test.go
```diff
@@ -0,0 +1,318 @@
...
+				}
+
+				var err error
+				switch methodName {
+				case "EnsureLoadBalancer":
+					_, err = lb.EnsureLoadBalancer(context.TODO(), "test", service, nodes)
+				case "UpdateLoadBalancer":
+					err = lb.UpdateLoadBalancer(context.TODO(), "test", service, nodes)
+				default:
+					t.Errorf("unsupported loadbalancer method: %s", methodName)
```
timoreimann: Should be `t.Fatalf`.

cloud-controller-manager/do/certificates_test.go
```diff
@@ -0,0 +1,318 @@
...
+
+	"github.com/digitalocean/godo"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+type kvCertService struct {
+	store         map[string]*godo.Certificate
+	reflectorMode bool
```
timoreimann: We do not seem to set `reflectorMode` anywhere, can we remove it?

When in doubt (and FWIW), I generally prefer populating fakes with the right values over setting a special mode (except for simulating error cases, but that's not what we have here).

cloud-controller-manager/do/certificates_test.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+func newKVCertService(store map[string]*godo.Certificate, reflectorMode bool) kvCertService {
+	return kvCertService{
+		store:         store,
+		reflectorMode: reflectorMode,
+	}
+}
+
+func createServiceAndCert(lbID, certID, certType string) (*v1.Service, *godo.Certificate) {
```
timoreimann: refactor: I implemented a [service builder](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/67c685a86ca3b54897fa4f835c0e652844466a9a/cloud-controller-manager/do/resources_test.go#L42) elsewhere which could be extended and leveraged for testing certificates.

For common service builder parameters, I also [created shortcuts at times](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/67c685a86ca3b54897fa4f835c0e652844466a9a/cloud-controller-manager/do/resources_test.go#L94-L96).

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -275,27 +306,91 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
...
-// UpdateLoadBalancer will not modify service or nodes.
-func (l *loadBalancers) UpdateLoadBalancer(ctx context.Context, clusterName string, service *v1.Service, nodes []*v1.Node) error {
+func getCertificateIDFromLB(lb *godo.LoadBalancer) string {
+	var id string
+	for _, rule := range lb.ForwardingRules {
+		if rule.CertificateID != "" {
+			id = rule.CertificateID
+			// CCM does not currently support multiple certificates on the
+			// loadbalancers it manages so we only need to grab the first one we find
+			break
```
timoreimann: refactor: we do not need the `id` variable: this line can return `rule.CertificateID` directly, and 319 can return an empty string.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -275,27 +306,91 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
...
+
+func (l *loadBalancers) checkAndUpdateLBAndServiceCerts(ctx context.Context, service *v1.Service, lbCertID, serviceCertID string) error {
+	if lbCertID != "" && lbCertID != serviceCertID {
+		lbCert, _, err := l.resources.gclient.Certificates.Get(ctx, lbCertID)
+		if err != nil {
+			respErr, ok := err.(*godo.ErrorResponse)
+			if ok && respErr.Response.StatusCode == http.StatusNotFound {
+				return nil
+			}
+			return fmt.Errorf("failed to get DO certificate for lb: %s", err)
```
timoreimann: refactor: CCM should always spell out `load-balancer` in error messages.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -275,27 +306,91 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
...
+			id = rule.CertificateID
+			// CCM does not currently support multiple certificates on the
+			// loadbalancers it manages so we only need to grab the first one we find
+			break
+		}
+	}
+	return id
+}
+
+func (l *loadBalancers) checkAndUpdateLBAndServiceCerts(ctx context.Context, service *v1.Service, lbCertID, serviceCertID string) error {
```
timoreimann: Method name needs to be updated. Suggestion: `recordLetsEncryptLBCert`

(I've dropped the `check` prefix deliberately, don't feel it's necessary to encode the check-ness nature of code into names as we do those for short-circuiting all the times.)

timoreimann: Can we add a GoDoc describing what the function does and in particular why we need it (i.e., explain the case of LE certs getting updated on LBs and CCM needing to sync with that)?

cloud-controller-manager/do/loadbalancers_test.go
```diff
@@ -85,8 +86,38 @@ func (f *fakeLBService) RemoveForwardingRules(ctx context.Context, lbID string,
 	return f.removeForwardingRulesFn(ctx, lbID, rules...)
 }
 
+func newKVLBService(store map[string]*godo.LoadBalancer) fakeLBService {
```
timoreimann: Really appreciate moving towards an LB fake. 👍 

If this continues to work with the current code, then I'm all happy. FWIW, I'm working on a branch that [fakes the LB](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/b1cb741066efa35a3ae3fdeedebfec6c350f208d/cloud-controller-manager/do/fakelbsvc_test.go) as well. It's a bit more sophisticated since I'm already covering more cases. Additionally, I found out that I had to do things like copying LB structs to ensure that modifying them in production code does not impact the fake store.

Feel free to take inspiration from my WIP branch or ignore it entirely as needed by your PR.

cloud-controller-manager/do/certificates_test.go
```diff
@@ -0,0 +1,318 @@
...
+					Port:     int32(443),
+					NodePort: int32(30000),
+				},
+			},
+		},
+	}
+}
+
+func Test_LBaaSCertificateScenarios(t *testing.T) {
+	testcases := []struct {
```
timoreimann: refactor: `s/testcases/tests/` to be consistent with the naming style we employ on more recent tests.

cloud-controller-manager/do/certificates_test.go
```diff
@@ -0,0 +1,318 @@
...
+			Name: "node-3",
+		},
+	}
+	fakeDroplet := fakeDropletService{
+		listFunc: func(context.Context, *godo.ListOptions) ([]godo.Droplet, *godo.Response, error) {
+			return droplets, newFakeOKResponse(), nil
+		},
+	}
+
+	for _, tc := range testcases {
```
timoreimann: refactor: `s/tc/test/` to be consistent with the naming style we employ or recent tests.

cloud-controller-manager/do/certificates_test.go
```diff
@@ -0,0 +1,318 @@
...
+				}
+
+				serviceCertID := getCertificateID(service)
+				if tc.expectedServiceCertID != serviceCertID {
+					t.Error("unexpected service certificate ID annotation")
+					t.Logf("expected: %s", tc.expectedServiceCertID)
+					t.Logf("actual: %s", serviceCertID)
+				}
+
+				godoLoadBalancer, _, err := lbService.Get(context.TODO(), getLoadBalancerID(service))
```
timoreimann: This should use `context.Background()` instead of `context.TODO()` because the latter is supposed to be used

>when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter)

([source](https://golang.org/pkg/context/#TODO))

cloud-controller-manager/do/certificates_test.go
```diff
@@ -0,0 +1,318 @@
...
+				if err != nil {
+					t.Fatalf("failed to get loadbalancer %q from fake client: %s", getLoadBalancerID(service), err)
+				}
+				lbCertID := getCertificateIDFromLB(godoLoadBalancer)
+				if tc.expectedLBCertID != lbCertID {
+					t.Error("unexpected loadbalancer certificate ID")
```
timoreimann: refactor: I know many of our existing tests follow this three line pattern, which is overly verbose -- especially for assertions on primitive types.

```
t.Errorf("got certificate ID %s, want %s", lbCertID, tc.expectedLBCertID)
```

should be a win on all fronts.

cloud-controller-manager/do/certificates_test.go
```diff
@@ -0,0 +1,318 @@
...
+					t.Logf("expected: %s", tc.expectedServiceCertID)
+					t.Logf("actual: %s", serviceCertID)
+				}
+
+				godoLoadBalancer, _, err := lbService.Get(context.TODO(), getLoadBalancerID(service))
+				if err != nil {
+					t.Fatalf("failed to get loadbalancer %q from fake client: %s", getLoadBalancerID(service), err)
+				}
+				lbCertID := getCertificateIDFromLB(godoLoadBalancer)
+				if tc.expectedLBCertID != lbCertID {
```
timoreimann: refactor: [revert comparison order](https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures).

cloud-controller-manager/do/certificates_test.go
```diff
@@ -0,0 +1,318 @@
...
+			setupFn: func(lbService fakeLBService, certService kvCertService) *v1.Service {
+				lb := createLB()
+				lbService.store[lb.ID] = lb
+				return createService(lb.ID)
+			},
+		},
+
+		// lets_encrypt test cases
+		{
+			name: "[letsencrypt] LB cert ID and service cert ID match ",
```
timoreimann: Trailing blank at the end of the string.

cloud-controller-manager/do/loadbalancers_test.go
```diff
@@ -100,6 +131,31 @@ func createLB() *godo.LoadBalancer {
 	}
 }
 
+func createHTTPSLB(entryPort, targetPort int, lbID, certID, certType string) (*godo.LoadBalancer, *godo.Certificate) {
```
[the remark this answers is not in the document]
timoreimann: Should probably have a better name like `createHTTPSLBAndCertificate`.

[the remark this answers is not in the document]
timoreimann: refactor: What I'd prefer though is if the function does one thing only (creation of LBs), with the creation of the certificate happening outside and the certificate ID being passed into this function (i.e., follow the S in SOLID).

cloud-controller-manager/do/certificates_test.go
```diff
@@ -0,0 +1,318 @@
...
+func Test_LBaaSCertificateScenarios(t *testing.T) {
+	testcases := []struct {
+		name                  string
+		setupFn               func(fakeLBService, kvCertService) *v1.Service
+		expectedServiceCertID string
+		expectedLBCertID      string
+		err                   error
+	}{
+		{
+			name: "default test values, tls not enabled",
```
timoreimann: What is the distinct purpose of this test? Isn't it covered by the other tests below?

cloud-controller-manager/do/certificates_test.go
```diff
@@ -0,0 +1,318 @@
...
+				certService.store[cert.ID] = cert
+				return createServiceWithCert(lb.ID, cert.ID)
+			},
+			expectedServiceCertID: "test-cert-id",
+			expectedLBCertID:      "test-cert-id",
+		},
+		{
+			name: "[letsencrypt] LB cert ID and service cert ID differ and both certs exist",
+			setupFn: func(lbService fakeLBService, certService kvCertService) *v1.Service {
+				lb, cert := createHTTPSLB(443, 30000, "test-lb-id", "test-cert-id", certTypeLetsEncrypt)
```
timoreimann: nit: would name `test-cert-id` as `lb-cert-id` as "test" has fairly little meaning.

cloud-controller-manager/do/certificates_test.go
waynr: it was originally intended to validate the default LB and Cert "fake" services when I was attempting to follow the example of overriding interface functions on a per test case level. I don't think it's necessary anymore now that I've implemented the key-value store backed fakes so I'll remove it.

cloud-controller-manager/do/certificates_test.go
waynr: Ah yeah, I meant to do this eventually but forgot -- I was originally following established naming conventions I saw in the loadbalancer tests.

cloud-controller-manager/do/loadbalancers_test.go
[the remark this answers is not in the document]
waynr: I DRY'd up this up by removing the `entryPort` and `targetPort` parameters but did not refactor.

waynr:
@timoreimann I've marked conversations where I addressed your comments as resolved FYI.

@snormore @timoreimann 
Conformance tests just passed for me against clusters with the latest head of this branch deployed. I have manually validated that none of the changes here affect creation or deletion of Services. I have also validated the new behavior for Let's Encrypt certs with the following workflow:

* Create a DO cluster using a testing version of DOKS that includes the changes in this PR
* Create a cert on the command line:
```
doctl compute certificate create --type lets_encrypt --dns-names testing.waynr.net --name testing.waynr.net
```
* Create a `Service` object on the cluster with `service.beta.kubernetes.io/do-loadbalancer-certificate-id` set to the UUId of the aforementioned certificate
* Observe the creation of a load-balancer in my DO cloud account
* Observe the creation of a `testing.waynr.net` subdomain in my DNS settings
* Create another cert on the command line:
```
doctl compute certificate create --type lets_encrypt --dns-names testing2.waynr.net --name testing2.waynr.net
```
* Modify the load-balancer's forwarding rules to use the new `testing2.waynr.net` cert.
* Add a new node pool to the cluster through the cloud UI
* Observe the `Service` annotations updated with the new `testing2.waynr.net` cert UUID

PR #275, opened 2019-09-06
Add service certificate existence check

waynr:


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -335,6 +335,18 @@ func (l *loadBalancers) checkAndUpdateLBAndServiceCerts(ctx context.Context, ser
 		}
 	}
 
+serviceCheck:
```
snormore: Should this be pulled out into a function and just `return l.checkServiceCertificate(...)` in both places?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -335,6 +335,18 @@ func (l *loadBalancers) checkAndUpdateLBAndServiceCerts(ctx context.Context, ser
...
+	if serviceCertID != "" {
+		_, _, err := l.resources.gclient.Certificates.Get(ctx, getCertificateID(service))
+		if err != nil {
+			respErr, ok := err.(*godo.ErrorResponse)
+			if ok && respErr.Response.StatusCode == http.StatusNotFound {
+				return fmt.Errorf("the %q service annotation refers to nonexistent DO Certificate %q", annDOCertificateID, serviceCertID)
+			}
+			return fmt.Errorf("failed to get DO certificate for service: %s", err)
```
timoreimann: Should we mention the service that we failed to get the DO certificates for?

PR #277, opened 2019-09-09
Add service certificate existence check

waynr:


timoreimann:
I'll be closing this one for now as we are about to ship an internal improvement to the error message wording.

We can always revisit the PR if the change does not turn out to be as effective as we expect it to be.

PR #280, opened 2019-09-16
Add missing "make vendor" call

timoreimann:
CI complained about this in #279.

Not sure how it slipped through, our CI should prevent against that. Let's fix this now and keep a close eye on the next PR that touches dependencies.


timoreimann:
This belongs to the 1.13 change, so closing here.

PR #282, opened 2019-09-30
Error when HTTP to HTTPS redirect requested without HTTPS port defined

### LINKED ISSUE #192: Handle situation where HTTP-to-HTTPS redirection is enabled but HTTPS service port is missing

timoreimann:
Fixes #192

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -660,8 +660,8 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
 		httpsPorts = append(httpsPorts, defaultSecurePort)
 	}
 
-	if getRedirectHTTPToHTTPS(service) && len(httpsPorts) == 0 {
-		return nil, errors.New("redirect from HTTP to HTTPS requested but no HTTPS port defined")
+	if getRedirectHTTPToHTTPS(service) && len(httpsPorts) == 0 && len(http2Ports) == 0 {
+		return nil, errors.New("redirect from HTTP to HTTPS requested but no HTTPS/HTTP2 port defined")
 	}
 
 	httpsPortMap := map[int32]bool{}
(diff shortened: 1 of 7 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -660,6 +660,10 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
 		httpsPorts = append(httpsPorts, defaultSecurePort)
 	}
 
+	if getRedirectHTTPToHTTPS(service) && len(httpsPorts) == 0 {
```
snormore: Should this include `http2Ports`?
```go
if getRedirectHTTPToHTTPS(service) && len(httpsPorts) + len(http2Ports) == 0 {
```

timoreimann: I was under the impression that HTTP/2 wouldn't be considered by the LB product, but turns out that it is.

I updated the PR accordingly. Also made a small addendum to the service annotation docs.

timoreimann:
Had to change the way we identify the problematic configuration in question. Quoting my latest commit description:

>It turned out that redirection only works on port 443, which is why we need to be more specific.
>
>Additionally, we should not base our decision on the ports specified in the annotations because (a) we always add the default secure 443 port to the list of secure ports we check for emptiness if pass-through or a certificate are requested; and (b) it is the actually built forwarding rule that determines if HTTP-to-HTTPx redirection can apply, not the intention expressed through the annotations. For instance, a user could specify port 443 in the HTTPS ports annotation quite explicitly but still miss to have a service on port 443 defined, which the previous implementation would not mark as an error.
>
>Hence, we loop over the built forwarding rules and check that there is at least one secure entry protocol defined on port 443. As a minor benefit, this allows us to invoke `getRedirectHTTPToHTTPS` a single time only.

timoreimann:
We have moved the validation / error reporting logic into the DigitalOcean Load Balancer product by now where it's most likely better placed (since it does not benefit DOKS only).

Closing this PR therefore.

PR #284, opened 2019-10-07
Document health check port usage

timoreimann:


docs/controllers/services/annotations.md
```diff
@@ -20,6 +20,8 @@ The path used to check if a backend droplet is healthy. Defaults to "/".
 
 The health check protocol to use to check if a backend droplet is healthy. Defaults to `tcp` if not specified. Options are `tcp` and `http`.
 
+The first node port on the service will be used as health check port.
```
snormore: `s/will be used/is used`?
> The first node port on the service is used as health check port.

timoreimann: Yeah that's better. Updated the PR.

PR #287, opened 2019-10-21
Update apiVersion to apps/v1

maniis:
Updated extensions/v1beta1 to apps/v1 and added matchLabels for selector in all required manifest files.

### DIFF
--- a/docs/controllers/services/examples/http-nginx-with-healthcheck-numeric-values.yml
+++ b/docs/controllers/services/examples/http-nginx-with-healthcheck-numeric-values.yml
@@ -19,7 +19,7 @@ spec:
       targetPort: 80
 
 ---
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: nginx-example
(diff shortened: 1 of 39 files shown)


docs/controllers/services/examples/http-nginx-with-healthcheck-numeric-values.yml
```diff
@@ -11,15 +11,16 @@ metadata:
 spec:
   type: LoadBalancer
   selector:
-    app: nginx-example
+    matchLabels:
```
timoreimann: You modified the service object here and elsewhere, which I believe does not support [set-based requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements) today.

docs/controllers/services/examples/http-nginx-with-healthcheck-numeric-values.yml
```diff
@@ -11,15 +11,16 @@ metadata:
...
+      app: nginx-example
   ports:
     - name: http
       protocol: TCP
       port: 80
       targetPort: 80
 
 ---
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
```
timoreimann: I think it's an odd artefact of Github to display an update to the API version line here: the [current master does not seem to contain any further references to `extensions/v1beta1` other than for previously released manifests](https://github.com/digitalocean/digitalocean-cloud-controller-manager/search?p=1&q=extensions%2Fv1beta1&unscoped_q=extensions%2Fv1beta1) (i.e., non-examples) which I believe are best left untouched.

PR #292, opened 2019-11-10
Warn about a bug in service annotations docs

opyh:
A [kubernetes bug](https://github.com/kubernetes/kubernetes/issues/59113) caused my external-dns annotation to vanish because I had used a wrong DO service annotation with a boolean value instead of a string by accident. The load balancer annotation documentation suggests using boolean values would work here, my IDE didn't print a warning either, and using any boolean annotation causes **all** annotations on the Service object to vanish on `kubectl apply`.

This PR improves docs so other devs hopefully don't use boolean values in service annotations, possibly saving them some hours of debugging when getting bitten by similar bugs.

### DIFF
--- a/docs/controllers/services/annotations.md
+++ b/docs/controllers/services/annotations.md
@@ -60,7 +60,11 @@ Ports must not be shared between this annotation and `service.beta.kubernetes.io
 
 ## service.beta.kubernetes.io/do-loadbalancer-tls-passthrough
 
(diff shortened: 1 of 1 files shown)


opyh:
This should be handled more gracefully in kubectl - hope the fix helps & thanks for merging it! :)

timoreimann:
The upstream fix got merged in October so we should benefit from better handling in Kubernetes 1.17. 🎉

MattIPv4:
Hey, @opyh - Thanks a ton for this sweet PR! 😄

Would you please shoot me an email when you get a chance?
mcowley at digitalocean dot com 🎉

PR #293, opened 2019-11-11
Add DO specific health check

nanzhong:
We want to expose a health check endpoint that we can use for monitoring the availability of CCM running in a cluster.

The way DO's CCM is currently implemented, it's configuring a healthz handler against the default serve mux, but actually not serving it anymore.

We re-use the upstream ccm cobra command for starting up CCM, and that [cobra command actually sets up all the health checking on the secure/unsecure bind address that you can configure with the `--address` flag](https://github.com/kubernetes/kubernetes/blob/master/cmd/cloud-controller-manager/app/controllermanager.go#L145-L169), however, due to the way that is implemented, there is no way we can access and and extend it.

Digging through what health checks it actually configures, it seems like it either installs the leader election health check if leader election is configured (which we disable), or it installs the default ping health checker (the empty case).

The health check that we really want to have is to check that the DO API is functional. There are a number of approach we can take here, but the simplest way is to check if a simple request successful (I've chosen the `/v2/account` endpoint).

This PR adds an additional flag/env var that can be used to configure a debug addr that we will be serving the `/healthz` endpoint on. It's intentionally setup to serve the health information insecurely to avoid complexity for now (see future considerations).

Some things to consider in the future:
- tie the DO api health to the actual endpoints that we use
    - we could tie the "health" to a mix of the `/v2/account` endpoint (to handle the empty/initial case) + the past `n` number of droplet/load balancer requests ccm has made.
- revisit the way we start up ccm (ie. do we want to continue using the upstream command directly, or do we want to reimplement the parts of it that are relevant to us)
    - this will influence how we want to support serving the debug handler securely

### DIFF
--- a/cloud-controller-manager/do/cloud.go
+++ b/cloud-controller-manager/do/cloud.go
@@ -17,25 +17,36 @@ limitations under the License.
 package do
 
 import (
+	"context"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 
 	"github.com/digitalocean/godo"
 
 	"golang.org/x/oauth2"
 
+	"k8s.io/apiserver/pkg/server/healthz"
 	"k8s.io/client-go/informers"
 	cloudprovider "k8s.io/cloud-provider"
 )
 
 const (
 	// ProviderName specifies the name for the DigitalOcean provider
-	ProviderName        string = "digitalocean"
+	ProviderName string = "digitalocean"
+
+	// At some point we should revisit how we start up our CCM implementation.
+	// Having to look at env vars here instead of in the cmd itself is not ideal.
+	// One option is to construct our own command that's specific to us.
+	// Alibaba's ccm is an example how this is done.
+	// https://github.com/kubernetes/cloud-provider-alibaba-cloud/blob/master/cmd/cloudprovider/app/ccm.go
 	doAccessTokenEnv    string = "DO_ACCESS_TOKEN"
 	doOverrideAPIURLEnv string = "DO_OVERRIDE_URL"
 	doClusterIDEnv      string = "DO_CLUSTER_ID"
 	doClusterVPCIDEnv   string = "DO_CLUSTER_VPC_ID"
+	debugAddrEnv        string = "DEBUG_ADDR"
+	defaultDebugAddr    string = "127.0.0.1:12301"
 )
 
 var version string
@@ -58,6 +69,8 @@ type cloud struct {
 	loadbalancers cloudprovider.LoadBalancer
 
 	resources *resources
+
+	httpServer http.Server
 }
 
 func newCloud() (cloudprovider.Interface, error) {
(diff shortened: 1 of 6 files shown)


cloud-controller-manager/do/health.go
```diff
@@ -0,0 +1,46 @@
...
+func (c *godoHealthChecker) Name() string {
+	return "godo"
+}
+
+func (c *godoHealthChecker) Check(r *http.Request) error {
+	ctx, cancel := context.WithTimeout(r.Context(), godoHealthTimeout)
+	defer cancel()
+	_, _, err := c.client.Account.Get(ctx)
+	if err != nil {
+		return fmt.Errorf("checking godo health: %w", err)
```
timoreimann: Don't think we need to wrap the error here since programmatic usage seems unlikely. (See also the last paragraph of the _Whether to Wrap_ section in [this Go blog post](https://blog.golang.org/go1.13-errors).)

cloud-controller-manager/do/cloud.go
```diff
@@ -97,13 +110,24 @@ func newCloud() (cloudprovider.Interface, error) {
 	clusterVPCID := os.Getenv(doClusterVPCIDEnv)
 	resources := newResources(clusterID, clusterVPCID, doClient)
 
+	debugMux := http.NewServeMux()
+	healthz.InstallHandler(debugMux, &godoHealthChecker{client: doClient})
+	debugAddr := defaultDebugAddr
+	if os.Getenv(debugAddrEnv) != "" {
```
jcodybaker: This looks fishy.

cloud-controller-manager/do/health.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func (c *godoHealthChecker) Name() string {
+	return "godo"
+}
+
+func (c *godoHealthChecker) Check(r *http.Request) error {
+	ctx, cancel := context.WithTimeout(r.Context(), godoHealthTimeout)
+	defer cancel()
+	_, _, err := c.client.Account.Get(ctx)
```
jcodybaker: If we use the default kubernetes health-check duration (10s) we'll see 360 req/hour, which is a non-trivial portion of the 5000 req/hour limit.

cloud-controller-manager/do/cloud.go
nanzhong: 😓 Oops, fixed.

cloud-controller-manager/do/health.go
nanzhong: I think this health check won't actually get triggered by that health checking mechanism unless we (or the user) intentionally configures it right?

My thinking here is our agent would be the only thing that hits the endpoint, and that would be triggered once every few minutes by our agent monitor service in addition to once every reconcile.

The diy cluster case I'm not too sure about though; do you think clear documentation around this is sufficient?

jcodybaker: Yeah, that's fair.  I also forgot that request limit is scoped to the token, not the whole account.

PR #294, opened 2019-11-13
Feature/letsencrypt

nicktate:
### Description

This is a **WIP** enhancement to add support for letencrypt certificates on digital ocean loadbalancers. I wanted to get it up for some early review and discussion around caveats.

`annDODomain = "service.beta.kubernetes.io/do-loadbalancer-domain"` has been added as a supported annotation to specify on Kubernetes services.

**It is required**  to already have a valid root domain on digitalocean. Digital ocean only provides a name server service, it is not itself a domain registrar. I've opted to require this manual step because even though we could automatically generate the root domain entry via the apis, it would be on the user to update the nameservers for their domain to point to digitalocean nameservers.

The domain is parsed into an effective top level domain (eTLD), a root domain, and a subdomain.

> If the given domain is `mysite.example.com` the `eTLD` is `com` because this is the top level domain new domains can be issued under. As a second example `mysite.example.co.uk`, the `eTLD` is `co.uk`

The logic generally flows as follows:

1. Ensure a `root` domain exists on digital ocean for the given annotation
2. Find or create a letsencrypt certificate in digital ocean that is valid for the given domain, and then update the service `certificate-id` annotation to reflect that certificate
3. Create or update the loadbalancer with the forwarding rules using the certificate-id
  a. *Note* The digitalocean loadbalancer service also effectively creates an `A record` for the root domain to point to the load balancer IP
4. Once the loadbalancer has reached active status, create a `A record` for the subdomain pointing to the loadbalancer IP if a subdomain exists

#### TODO
* [ ] Finish updating unit tests for `EnsureLoadBalancerDeleted` and `EnsureLoadBalancer` to take into account the new functionality 
* [ ] Update documentation

#### Additional Considerations / Caveats
* The current implementation is not able to successfully cleanup generated certificates or A-records if the user modifies the service to change the `certificate-id`, or `domain`. The interface provided through the `CCM` abstracts away the previous state of the service so you lose reference to what the pre-existing values were.
  * We could attempt to solve this either via finalizers or creating our own reconciler but it seems like we would be stepping too far outside the confines of the `ccm` interface an duplicating a lot of effort
* The LB service does not cleanup generated root domain records. It seems like a fix we should push into the LB service versus handling in the ccm
* Existing certificate rotation logic will continue to work as expected because during the `ensureCertificate` step, fetch will fail to locate the existing certificate and then research for a new certificate that is valid for the given domain (which will be the rotated cert)
  * In the case of a manually specified `certificate-id` rotation will also occur as normal and defer to the `recordUpdatedLetsEncryptCert` function
* This is creating additional network requests into the DO API, we should re-evaluate to see if we can minimize or remove some requests and ensure rate limit thresholds are not approached

#### What does this pull request fix

https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/92


cloud-controller-manager/do/certificates.go
```diff
@@ -16,8 +16,199 @@ limitations under the License.
...
+	if err != nil {
+		return err
+	}
+
+	if domain == nil {
+		return nil
+	}
+
+	// the do loadbalancer service ensures the root domain associated with the
+	// certificate has an A record pointing to the LB so we do not need to ensure
```
snormore: I'm not sure we can rely on this behaviour forever, since it's not clear from conversations with the LBaaS team whether or not this is intended behaviour. We should probably check for it anyway and add the record if it's not there, so that if the currently undocumented behaviour changes we won't have to scramble to backfill.

cloud-controller-manager/do/certificates.go
```diff
@@ -16,8 +16,199 @@ limitations under the License.
...
+	if domain.sub == "" {
+		klog.V(2).Infof("domain has no subdomain, no need to ensure A records: %s", domain.full)
+		return nil
+	}
+
+	records, _, err := l.resources.gclient.Domains.Records(ctx, domain.root, &godo.ListOptions{})
```
snormore: Move to `findDomainRecordByTypeAndName`?

cloud-controller-manager/do/certificates.go
```diff
@@ -16,8 +16,199 @@ limitations under the License.
...
+	}
+
+	if domainRecord == nil {
+		domainRecordEditReq := &godo.DomainRecordEditRequest{
+			Type: "A",
+			Name: domain.sub,
+			Data: lb.IP,
+			TTL:  defaultDomainRecordTTL,
+		}
+		klog.V(2).Infof("creating new A record to loadbalance for domain: %s", domain.full)
```
snormore: Can you include the LB ID and/or IP in this log entry too?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -322,6 +346,21 @@ func getCertificateIDFromLB(lb *godo.LoadBalancer) string {
 // Load Balancer.
 func (l *loadBalancers) recordUpdatedLetsEncryptCert(ctx context.Context, service *v1.Service, lbCertID, serviceCertID string) error {
 	if lbCertID != "" && lbCertID != serviceCertID {
+		if serviceCertID != "" {
```
snormore: Not sure if we want this here because this method is used to update the service (annotation) cert ID if it doesn't match the one on the LB, so I think this addition will short circuit that check since if it doesn't match then it's likely rotated out of the account and so won't exist? 🤔

nicktate: This particular check will only short circuit out of checking for existence of the `serviceCert` if a value does not exist and then it will fall through to see if there is a valid certificate on the lb below at line 364. 

It's probably worth relooking at how this method works in conjunction with `ensureCertificate` to see if there is any room for collapsing.

cloud-controller-manager/do/certificates.go
```diff
@@ -16,8 +16,199 @@ limitations under the License.
...
+
+// ensureDomain checks to see if the service contains the annDODomain annotation
+// and if it does it verifies the domain exists on the users account
+func (l *loadBalancers) ensureDomain(ctx context.Context, service *v1.Service) error {
+	domain, err := getDomain(service)
+	if err != nil {
+		return err
+	}
+
+	if domain != nil {
```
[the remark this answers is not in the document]
nicktate: :+1: I thought I had cleaned up all of these, I also agree on short circuiting fast as a preferred style.

cloud-controller-manager/do/certificates.go
```diff
@@ -16,8 +16,199 @@ limitations under the License.
...
+	}
+
+	return nil
+}
+
+// ensureCertificate verifies the existing certificate on the service is still a valid
+// DO certificate, clearing if not. If the user also has the annDODomain annotation set,
+// it verifies the certificate is valid for the domain, creating a new one if it cannot
+// find a existing valid certificate in the account
+func (l *loadBalancers) ensureCertificate(ctx context.Context, service *v1.Service) error {
```
timoreimann: This method does a fair amount of things and feels non-trivial to follow to me.

Is it possible we split it up into two or three smaller pieces? I feel the the certificate lookup and (first) service update part would be a good scope for the first function (i.e., everything up to and including line 85). I'd return a certificate that the follow-up function (`ensureValidCertificateForDomain`?) would receive and process accordingly.

WDYT?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -339,7 +378,7 @@ func (l *loadBalancers) recordUpdatedLetsEncryptCert(ctx context.Context, servic
 	return nil
 }
 
-func (l *loadBalancers) updateLoadBalancer(ctx context.Context, lb *godo.LoadBalancer, service *v1.Service, nodes []*v1.Node) (*godo.LoadBalancer, error) {
+func (l *loadBalancers) updateLoadBalancer(ctx context.Context, service *v1.Service, lb *godo.LoadBalancer, nodes []*v1.Node) (*godo.LoadBalancer, error) {
```
timoreimann: Out of curiosity, was there a particular reason you reversed the order here?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -403,6 +442,7 @@ func (l *loadBalancers) EnsureLoadBalancerDeleted(ctx context.Context, clusterNa
 		return err
 	}
 
+	klog.V(2).Infof("deleting do loadbalancer: %s", lb.ID)
```
timoreimann: Log messages should be capitalized and wording be made consistent with the pre-existing style, i.e., `Deleting load-balancer %s`.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -411,6 +451,54 @@ func (l *loadBalancers) EnsureLoadBalancerDeleted(ctx context.Context, clusterNa
...
 	}
 
+	domain, err := getDomain(service)
+	if err != nil {
+		klog.Errorf("failed to fetch domain from service for cleanup: %s", err)
+		return nil
+	}
+
+	// attempt to clean up any managed A records or certificates related to LB
+	if domain != nil && lb != nil {
```
timoreimann: I don't think we need to check `lb` for `nil`: it should always exists given that we check for the existence of the LB first thing in this function (and bail out if it doesn't exist).

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -411,6 +451,54 @@ func (l *loadBalancers) EnsureLoadBalancerDeleted(ctx context.Context, clusterNa
...
+		certID := getCertificateIDFromLB(lb)
+		certName := getCertificateName(domain.full)
+		cert, resp, err := l.resources.gclient.Certificates.Get(ctx, certID)
+		if err != nil && resp.StatusCode != http.StatusNotFound {
+			klog.Errorf("failed to fetch certificate: %s", err)
```
timoreimann: Should this be a `return`?

timoreimann: Let's include the certificate ID in the error message (`"failed to fetch certificate by ID %s: %s", certID, err`).

Similar below.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -411,6 +451,54 @@ func (l *loadBalancers) EnsureLoadBalancerDeleted(ctx context.Context, clusterNa
...
+
+		// if subdomain existed, attempt to find and clean up A record that was created
+		if domain.sub != "" {
+			for _, record := range records {
+				// If we find the A record corresponding to this domain and LB
+				if record.Type == "A" && record.Name == domain.sub && record.Data == lb.IP {
+					klog.V(2).Infof("deleting loadbalancer A record for domain(%s): %s", domain.root, record.Name)
+					resp, err = l.resources.gclient.Domains.DeleteRecord(ctx, domain.root, record.ID)
+					if err != nil && resp.StatusCode != http.StatusNotFound {
+						klog.Errorf("failed to domain A record: %s", err)
```
timoreimann: Verb missing.

cloud-controller-manager/do/certificates.go
nicktate: Yeah completely agree. It will help lead to collapsing logic related to certificates in this function and `recordUpdatedLetsEncryptCert` https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/294/files#r346345188 as well.

cloud-controller-manager/do/certificates.go
```diff
@@ -16,8 +16,199 @@ limitations under the License.
...
+		return err
+	}
+
+	if domain != nil {
+		klog.V(2).Infof("looking up root domain: %s", domain.root)
+		_, resp, err := l.resources.gclient.Domains.Get(ctx, domain.root)
+		if err != nil && resp.StatusCode == http.StatusNotFound {
+			return fmt.Errorf("domain does not exist: %s", domain.root)
+		} else if err != nil {
+			return fmt.Errorf("failed to retrieve domain: %s", err)
```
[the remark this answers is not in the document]
nicktate: That's a good call, the API does return proper messaging.

cloud-controller-manager/do/certificates.go
```diff
@@ -16,8 +16,199 @@ limitations under the License.
...
+	serviceCertID := getCertificateID(service)
+
+	// verify the cert-id is still referencing a valid certificate
+	if serviceCertID != "" {
+		klog.V(2).Infof("looking up existing service certificate: %s", serviceCertID)
+		certificate, _, err := l.resources.gclient.Certificates.Get(ctx, serviceCertID)
+
+		if err != nil {
+			respErr, ok := err.(*godo.ErrorResponse)
+			if !ok || respErr.Response.StatusCode != http.StatusNotFound {
```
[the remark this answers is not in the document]
nicktate: I had seen a few other examples in this codebase of how I implemented error handling in this section.  I agree with you though. I will update and also update the other corresponding sections for consistency.

cloud-controller-manager/do/loadbalancers.go
nicktate: It felt like the switch was making parameter ordering more consistent across the code base... For example, these existing functions feel like they are prioritizing service.

```
(l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName string, service *v1.Service, nodes []*v1.Node) 
recordUpdatedLetsEncryptCert(ctx context.Context, service *v1.Service, lbCertID, serviceCertID string)
(l *loadBalancers) UpdateLoadBalancer(ctx context.Context, clusterName string, service *v1.Service, nodes []*v1.Node) 
```

I guess from a more abstract level, the `service` is the driving factor for all of this functionality so seems like it should be prioritized. I also could have been losing my mind while deciding to make that change 😛because it seems so subtle you could go either way so I'm fine reverting.

cloud-controller-manager/do/loadbalancers.go
nicktate: I need to take a look upstream as to what happens when you return an error response in this method. It felt like the process shouldn't as long as the loadbalancer was successfully deleted. I will have to think more about whether we want strong guarantees around cleanup of these resources.

PR #305, opened 2020-01-30
Add support for configurable health-check via annotation

nicktate:
## Description

**DO NOT MERGE**
> Need to run some integration tests to verify functionality still

### What does this pull request accomplish

* Adds `service.beta.kubernetes.io/do-loadbalancer-healthcheck-port` annotation for services to specify which port to use for the loadbalancer health check.
* Defaults to existing functionality of choosing the first port in a service

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -49,6 +49,10 @@ const (
 	// for DO load balancers. Defaults to '/'.
 	annDOHealthCheckPath = "service.beta.kubernetes.io/do-loadbalancer-healthcheck-path"
 
+	// annDOHealthCheckPort is the annotation used to specify the health check port
+	// for DO load balancers. Defaults to the first Service Port
+	annDOHealthCheckPort = "service.beta.kubernetes.io/do-loadbalancer-healthcheck-port"
+
 	// annDOHealthCheckProtocol is the annotation used to specify the health check protocol
 	// for DO load balancers. Defaults to the protocol used in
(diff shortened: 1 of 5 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -780,6 +785,30 @@ func getHostname(service *v1.Service) string {
 	return strings.ToLower(service.Annotations[annDOHostname])
 }
 
+// healthCheckPort returns the health check port specified, defaulting
+// to the first port in the service otherwise.
+func healthCheckPort(service *v1.Service) (int, error) {
+	ports, err := getPorts(service, annDOHealthCheckPort)
+	if err != nil {
+		return -1, fmt.Errorf("failed to get health check port: %v", err)
```
timoreimann: It might be more idiomatic to return the zero value for the non-error parameter, i.e., 0 instead of -1.

(Similar for other error cases below.)

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -780,6 +785,30 @@ func getHostname(service *v1.Service) string {
...
+	}
+
+	if len(ports) > 1 {
+		return -1, fmt.Errorf("annotation[%s] only supports a single port, but found multiple %v", annDOHealthCheckPort, ports)
```
timoreimann: Maybe use

```
annotation %q
```

instead of

```
annotation[%s]
```

to improve readability?

timoreimann: readability nit: colon after _multiple_?

PR #307, opened 2020-02-15
Add support for `load-balancer-name` annotation

### LINKED ISSUE #298: Support specifying custom load balancer names
Load balancer names are based on Service object UIDs which consist of hard-to-read alphanumeric characters.

Some time ago, upstream provided the ability to specify custom names. We should take advantage of this and enable users to set their own names (through, say, an annotation).

One caveat to watch out for is that load balancer names must be unique. Consequently, we either need to surface any errors returning from the DO LoadBalancer API or ensure uniqueness on our end. The latter could be enforced by appending some kind of random hash (or one based off of the Service object UID).

The final piece to take into consideration are possible name length constraints, especially when thinking about adding a hash suffix.

grzesiek:
## Description

This pull request makes it possible to use `kubernetes.digitalocean.com/load-balancer-name` to override default name that consists of hard-to-read alphanumeric characters.

Closes #298


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -245,6 +249,13 @@ func (l *loadBalancers) GetLoadBalancerName(_ context.Context, clusterName strin
 }
 
 func getDefaultLoadBalancerName(service *v1.Service) string {
```
grzesiek: TODO this is no longer a `Default` name, perhaps we should rename this function.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -245,6 +249,13 @@ func (l *loadBalancers) GetLoadBalancerName(_ context.Context, clusterName strin
...
+	name := service.Annotations[annoDOLoadBalancerName]
+
+	// TODO do we need to slugify this?
```
grzesiek: TODO, we might need to _slugify_ this

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -245,6 +249,13 @@ func (l *loadBalancers) GetLoadBalancerName(_ context.Context, clusterName strin
...
+	if len(name) > 0 {
+		return name
```
grzesiek: TODO, append a few random characters to resolve uniqueness problem

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -245,6 +249,13 @@ func (l *loadBalancers) GetLoadBalancerName(_ context.Context, clusterName strin
...
+	}
+
```
grzesiek: TODO, cover this with tests

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -245,6 +256,15 @@ func (l *loadBalancers) GetLoadBalancerName(_ context.Context, clusterName strin
 }
 
 func getDefaultLoadBalancerName(service *v1.Service) string {
```
grzesiek: We might want to rename this method, as it is no `Default` name anymore. Also it appears that it is being used in a different places too.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -245,6 +256,15 @@ func (l *loadBalancers) GetLoadBalancerName(_ context.Context, clusterName strin
...
+	name := service.Annotations[annoDOLoadBalancerName]
```
grzesiek: It might be possible to resolve the name uniqueness problem here, by introducing another annotation we would use to track the _slugified_ name with a few random chars added at the end. My understanding of the problem here is that there is no good mechanism of surfacing errors that come from creation / reconciliation of a LoadBalancer service. I might be mistaken here, of course, but this seems plausible.

If that is indeed the case, we can introduce another annotation `kubernetes.digitalocean.com/load-balancer-name-id` that would be a slugified and unique name, generated by the cloud controller from the  user-provided `kubernetes.digitalocean.com/load-balancer-name`.

Having `load-balancer-name-id` would resolve the problem of uniqueness while making it possible to specify a user-friendly load balancer name. It would make this feature a little more robust too. It adds some amount of complexity, but not too much, and it might be a good balance. WDYT @timoreimann?

timoreimann: That's an interesting idea. In general, I'm hesitant to introduce more annotations than necessary: it has become quite challenging to fully comprehend the number of annotations (and their interactions) we support, which also matches with user feedback we have received.

I think we may not need that slugified name annotation anymore though -- please see my general review comment.

grzesiek:
Thanks for the reply @timoreimann, it makes sense to me. I will try to think about it a little more, and hopefully we will find a good solution together.

I'm still excited about working on this, so once we agree on requirements here, and the direction forward, I will be happy to keep working on that.

In the meantime I will spend some more time reading this codebase and figuring it out what our options are.

My perspective is also influenced by using Terraform to deploy LoadBalancer and the cluster itself, and I fully agree that using annotations is not perfect - each time I want to deploy a change Terraform warns me about mutating annotations, and this is this part of the UX I believe we could improve too.

grzesiek:
Let's discuss the direction forward in the issue :arrow_right: https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/298#issuecomment-590341641

grzesiek:
I managed to spend some time on testing the CCM against a real cluster, and it appears that it nicely records as errors in the load balancer service events stream, so indeed surfacing errors is already handled by kubernetes correctly. What else do we need to do here @timoreimann? Is the documentation part the only thing that is missing?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -40,6 +40,10 @@ const (
 	// used to enable fast retrievals of load-balancers from the API by UUID.
 	annoDOLoadBalancerID = "kubernetes.digitalocean.com/load-balancer-id"
 
+	// annDOLoadBalancerName is the annotation used to specify a name of the
+	// load balancer that is going to be created by the controller.
+	annDOLoadBalancerName = "service.beta.kubernetes.io/do-load-balancer-name"
```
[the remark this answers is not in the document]
grzesiek: Interesting! Are the other `service.beta.kubernetes.io/do-*` official annotations @timoreimann? I couldn't find them in official docs. If there annotations are not official too, I wonder what is the good balance between consistency / confusion :thinking:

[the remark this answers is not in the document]
timoreimann: Ah you're totally right, we should stick to the annotation you have chosen in this case: I saw `kubernetes.digitalocean.com/load-balancer-id` above and only realized now that we picked a different prefix in that one case because that annotation is meant to be read-only. The LB name is to be created/modified by users for sure.

Thanks for double-checking on this one -- let's keep it the way it is. 👍

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -40,6 +40,18 @@ const (
 	// used to enable fast retrievals of load-balancers from the API by UUID.
 	annoDOLoadBalancerID = "kubernetes.digitalocean.com/load-balancer-id"
 
+	// annDOLoadBalancerName is the annotation used to specify a name of the load
+	// balancer that is going to be created by the controller.
```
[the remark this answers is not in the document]
grzesiek: Sure, I will improve the comment, and extend docs. I think it might be also useful to mention in docs that reconciliation errors can be found in the load balancer events stream.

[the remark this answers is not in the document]
timoreimann: >I think it might be also useful to mention in docs that reconciliation errors can be found in the load balancer events stream.

That's an excellent suggestion. Let's capture that in a separate PR.

grzesiek:
@timoreimann I addressed your feedback and pushed a few commits. I also the missing test case for `GetLoadBalancerName` func. Let's add more docs in a separate PR :muscle:

cloud-controller-manager/do/loadbalancers.go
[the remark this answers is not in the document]
grzesiek: Done :arrow_right: https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/310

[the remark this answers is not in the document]
timoreimann: Not be misunderstood: we should update the comment here in this PR that LB updates are covered too.

grzesiek:
> I think my general comment in [#307 (review)](https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/307#pullrequestreview-369456518) still needs to be implemented, right?

Oh yeah, I somehow missed it. Let's check that.

grzesiek:
@timoreimann I addressed your comment about looking up a load balancer by a new name (defined in annotations) and by a legacy name. It required some refactoring of how we are searching for a load balancer in general, because the code was scattered between two places - `resources.go` and `loadbalancers.go`. I moved it to a single place, and hopefully it is no easier to understand what dependencies for the load balancer search are, please take a look at ae175db3bc593064ac2059c1a61d556e125c09cb, I will be happy to explain the refactoring if it is needed.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -40,6 +40,22 @@ const (
...
+	// following naming convention:
+	// * it must start with an alphanumeric character;
+	// * it must consist of alphanumeric characters or the '.' (dot) or '-'
+	// (dash) characters; except for the final character which must not be '-'
+	// (dash). Violations to any of these rules will lead to reconciliation
+	// errors. If no custom name is given, a default name is chosen consisting of
+	// the character 'a' appended by the Service UID.
+	//
+	// Adding this annotation to an existing service will also result in renaming
+	// an existing LoadBalancer associated with it.
```
timoreimann: Can we "merge" this line with the very first sentence of the comment? Right now, the first sentence mentions creates only, and it requires the reader to go through the entire configuration description before realizing that existing LBs are affected as well. I'd prefer that to be more immediately obvious.

cloud-controller-manager/do/resources.go
```diff
@@ -53,7 +53,7 @@ type resources struct {
 }
 
 // newResources initializes a new resources instance.
-
+//
```
timoreimann: Please keep this one in -- the upper and lower part of the comment belong to one "section".

cloud-controller-manager/do/resources.go
```diff
@@ -157,25 +154,17 @@ func (r *ResourcesController) syncTags() error {
...
-			id = loadBalancerIDsByName[name]
+		id, err := r.resources.findLoadBalancerID(svc, lbs)
+		if err != nil && err != errLBNotFound {
+			return fmt.Errorf("error occurred when searching for a load balancer: %s", err)
 		}
 
-		// Renamed load-balancers that have no LB ID set yet would still be
-		// missed, so check again if we have found an ID.
-		if id != "" {
+		if len(id) > 0 {
```
timoreimann: This change doesn't seem necessary. Can we undo it?

(FWIW, I prefer to use `len()` on data structures only and `!= ""` on strings because it makes it easier to see when a string is involved.)

cloud-controller-manager/do/resources.go
```diff
@@ -229,3 +218,61 @@ func (r *ResourcesController) tagResources(res []godo.Resource) error {
 
 	return err
 }
+
+func (r *resources) retrieveLoadBalancer(ctx context.Context, service *corev1.Service) (*godo.LoadBalancer, error) {
```
timoreimann: Is there a particular reason you defined/moved the methods on the `resources` struct? AFAICS, the receiver `r` is not used other than for calling methods.

`resources` is focused on updating tags on LB objects. I believe it'd make more sense to keep the functionality related to looking up LBs on the `loadBalancers` struct instead. It could also decrease the PR change set.

grzesiek: @timoreimann the problem is that these function require access to data we hold in both, `resouces` and `loadbalancers`, but `resources` do not have access to `loadbalancers`, `loadbalancers` have access to `resources`. From what I understand `resources` represents DigtialOcean resources, and semantically it makes sense to me.

Moving this code to `loadbalancers` would result in code duplication, which might be acceptable, but my impression is that keeping this in one place makes it easier to understand the code. WDYT?

grzesiek: By the required data that we hold in `reosuces` and `loadbalancers`. I mean the DO API client which is needed to retrieve a single load balancer resource / the list of load balancers.

Keeping this code in `resources` makes it possible to avoid a significant amount of duplication - otherwise we would need to duplicate code for retrieving load balancers between `resources` and `loadbalancers`.

One solution to this problem is to extracting DO API resources related to load balancers to a separate file, but I thought that `resources` is a good place where we could keep this code implemented, because my understanding is that indeed represents DO API resources :thinking:

cloud-controller-manager/do/resources.go
grzesiek: Sure, here it is just a matter of style, let's revert this change.

cloud-controller-manager/do/resources.go
grzesiek: Do you mean that we should remove the line 56 completely or remove the change and keep the empty line?

cloud-controller-manager/do/loadbalancers.go
grzesiek: Sure, makes sense.

cloud-controller-manager/do/resources.go
timoreimann: The name _resources_ is somewhat misleading: right now, the code in that file is about ensuring that tags are set on LBs. With that context in mind, I'm uncertain if it makes sense to have half of the LB-related code in `loadbalancers.go` and the other half in `resources.go`.

It wasn't immediately clear to me where the duplication would happen, so I took a stab at moving the LB functionality that's now in `resources.go` into `loadbalancers.go`. Turns out this was fairly easy after realizing that a few helper methods added do not really need a `*loadBalancers` receiver (i.e., they can be functions). Could you please take a look at my branch at https://github.com/timoreimann/digitalocean-cloud-controller-manager/tree/feature/gb/add-load-balancer-name-annotation and the commits I added? One is about moving the logic, and two more commits are about minor improvements I noticed might be worth doing.

If you agree that it makes sense to structure things like that, could I ask you to cherry-pick my commits into your PR branch? Afterwards, we can finish the code review.

By the way, I really like the refactoring you did around putting existing logic into functions. It makes the code much more readable. 💙

grzesiek: Ahh, I somehow missed that down the road I decoupled `findLoadBalancerByName` from `gclient`. It indeed looks much better now @timoreimann. I will cherry pick commits :+1: Thanks for showing me that I was wrong :bowing_man:

grzesiek:
@timoreimann I cherry-picked and pushed your commits. Thank for help on this :pray: Sorry for so much of the back-and-forth, but I really like how this looks like after your changes :+1: Please let me know if there is anything else I can improve in this pull request / in the documentation / something somewhere else.

I just pushed created https://github.com/grzesiek/terraform-digitalocean-kubernetes-cluster and I can't wait to add the new annotation there too :smile_cat:

timoreimann:
I hope you don't mind that I submitted a small PR against your PR (such meta) consisting of a final batch of small changes. Please have a look at grzesiek#1.

Afterwards, we're good to merge. 🌟

grzesiek:
Thanks @timoreimann! I really appreciate our collaboration on this feature! Thank you for spending time on helping me and making important changes yourself :+1: :purple_heart:

timoreimann:
We have another small change lined up to be merged before we'll go for a new CCM release. The plan is to make that happen before the end of the month.

Thanks again! 👏

grzesiek:
Thank you @timoreimann! Let me know if I can help with anything else! :purple_heart:

MattIPv4:
Hey, @grzesiek - Thanks a ton for this sweet PR! 😄

Would you please shoot me an email when you get a chance?
mcowley at digitalocean dot com 🎉

PR #308, opened 2020-02-20
Support specifying a fake region by environment variable

timoreimann:
The sole purpose of this change is to enable local testing since CCM depends on the DigitalOcean metadata service to be available, which only exists on droplets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ The solution is to change the service port to a different, non-conlicting one.
 
 ## Development
 
-Requirements:
+### Basics
 
 * Go: min `v1.12.x`
 
@@ -63,16 +63,35 @@ After making your code changes, run the tests and CI checks:
 make ci
 ```
 
-If you want to test your changes, create a new image with the version set to `dev`:
+### Run Locally
+
+If you want to run `digitalocean-cloud-controller-manager` locally against a
+particular cluster, keep your kubeconfig ready and start the binary in the main
+package-hosted directory like this:
+
+```bash
+cd cloud-controller-manager/cmd/digitalocean-cloud-controller-manager
+FAKE_REGION=fra1 go run main.go --kubeconfig <path to your kubeconfig file> --leader-elect=false --v=5 --cloud-provider=digitalocean
+```
+
+The `FAKE_REGION` environment variable takes a (valid) DigitalOcean region. It
+is needed to keep `digitalocean-cloud-controller-manager` from trying to access
+the DigitalOcean metadata service which is only available on droplets. Overall,
+which region you choose should not matter a lot as long as you pick one.
+
+### Run Containerized
+
+If you want to test your changes in a containerized environment, create a new
+image with the version set to `dev`:
 
 ```bash
 VERSION=dev make publish
 ```
 
 This will create a binary with version `dev` and docker image pushed to
-`digitalocean/digitalocean-cloud-controller-manager:dev`
+`digitalocean/digitalocean-cloud-controller-manager:dev`.
 
-### Release a new version
+## Release a new version
 
 To release a new version first bump the version:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -63,16 +63,35 @@ After making your code changes, run the tests and CI checks:
...
+If you want to test your changes in a containerized environment, create a new
+image with the version set to `dev`:
 
 ```bash
 VERSION=dev make publish
 ```
 
 This will create a binary with version `dev` and docker image pushed to
-`digitalocean/digitalocean-cloud-controller-manager:dev`
+`digitalocean/digitalocean-cloud-controller-manager:dev`.
```
grzesiek: Does it assume that a person who runs `make publish` has access to `digitalocean/digitalocean-cloud-controller-manager` images repository on Docker Hub? Would it make sense to retrieve Git SHA and make `VERSION=` env optional?

grzesiek: I also see that `make build` exists. I wonder if mentioning it in README along with / instead of `make publish` would be useful for external contributors.

cloud-controller-manager/do/metadata.go
```diff
@@ -20,12 +20,17 @@ import (
...
 )
 
 const dropletRegionMetadataURL = "http://169.254.169.254/metadata/v1/region"
 
 // dropletRegion returns the region of the currently running program.
 func dropletRegion() (string, error) {
+	// FAKE_REGION can be specified for local testing.
+	if region := os.Getenv("FAKE_REGION"); region != "" {
+		return region, nil
+	}
```
grzesiek: Does it mean that if someone sets environment variable in production environment, presumably by mistake, a region will not be detected correctly? Would it be better to pass an environment identifier in a command line, like `--test-env`, so that the code execution paths do not rely on presence or absence of arbitrary envs? :thinking:

README.md
timoreimann: It is possible to push to a different Docker Hub organization by setting the `REGISTRY` environment variable (I know, the name is misleading).

I wouldn't mind a PR to document this properly. Bonus points for mimicking what we have in the csi-digitalocean repository where [a single `DOCKER_REPO` variable can be used to override both the organization and the repository part](https://github.com/digitalocean/csi-digitalocean/blob/690eca5abea5724b87a75311fe78a85cc62765b7/Makefile#L31).

timoreimann: I'm not super sure about using a git SHA by default. Personally, I think I have never specified one -- I prefer using a readable tag that clearly identifies my feature and overwrite the tag as needed. (All in my personal Docker Hub repo only of course where it's okay to do such things.) But I understand that YMMV, so if there's more demand for it I'm open to letting myself get convinced.

timoreimann: PRs to improve our docs on how to use the Makefile targets are more than welcome!

cloud-controller-manager/do/metadata.go
timoreimann: Adding a custom flag has proven to be far from trivial in CCM since it's tightly bundled to the (fairly complex) flag scaffolding that every CCM inherits from kube-controller-manager. I haven't had the bandwidth yet to look more deeply into it, but I'd be very open to switching to flags if you could make it happen somehow.

I can make a quick amendment to the PR to log when the environment flag is used as an additional warning. I also tried to document its usage, so hopefully accidental presence of the environment variable in production environments should be minimized.

timoreimann: Here's another idea I've had: instead of specifying the region, it'd be cool to have a mock HTTP server that returns a fake region. The challenge would be to make it respond on the metadata IP address though.

Just wanted to mention it in case you or anyone else feels adventurous to prototype the idea.

README.md
grzesiek: Thanks for clarification, makes sense :+1:

cloud-controller-manager/do/metadata.go
grzesiek: Thanks for adding the log!

PR #310, opened 2020-03-09
Add docs on accessing a load balancer provisioning errors

grzesiek:
## Description

This pull requests adds docs on surfacing reconciliation errors that can happen during provisioning a load balancer _via_ cloud controller manager.

### DIFF
--- a/docs/controllers/services/examples/README.md
+++ b/docs/controllers/services/examples/README.md
@@ -169,11 +169,11 @@ You can find provisioning status and all the reconciliation errors that Cloud
 Controller Manager encountered during the LoadBalancer service life-cycle in
 the service's event stream.
 
-Provided that your load balancer service name is `my-load-balancer`, you can use
-following commands to access events attached to this object:
+Provided that your kubernetes service name is `my-svc`, you can use following
+commands to access events attached to this object:
 
 ```bash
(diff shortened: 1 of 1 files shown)


grzesiek:
@timoreimann Can you please take a look? Thanks in advance! :purple_heart:

docs/controllers/services/examples/README.md
```diff
@@ -154,5 +157,30 @@ spec:
...
+
+Cloud Controller Manager is using DigitalOcean API internally to provision a
+DigitalOcean load balancer. It might happen that provisioning will be
+unsuccessful, because of various reasons.
 
+You can find provisioning status and all the reconciliation errors that Cloud
+Controller Manager encountered during the LoadBalancer service life-cycle in
+the service's event stream.
+
+Provided that your load balancer service name is `my-load-balancer`, you can use
```
timoreimann: I might be biased since we're also working on that LB naming feature, but _load balancer service name_ could be confusing as to whether it refers to the LB or the service.

How about just calling it _service name_ and use the example name `my-svc`?

grzesiek: @timoreimann sounds good! It is done, I just pushed the update in d420e930 :smile_cat:

PR #325, opened 2020-05-26
[WIP] firewalls: add controller to manage ccm worker firewall

MorrisLaw:
WIP PR for firewall controller to manage the ccm worker firewall. Still need to add tests and test logic.

@timoreimann

### DIFF
--- a/cloud-controller-manager/do/firewall_controller.go
+++ b/cloud-controller-manager/do/firewall_controller.go
@@ -39,12 +39,8 @@ import (
 const (
 	// Interval of synchronizing service status from apiserver
 	serviceSyncPeriod = 30 * time.Second
-
-	// How long to wait before retrying the processing of a service change.
-	// If this changes, the sleep in hack/jenkins/e2e.sh before downing a cluster
-	// should be changed appropriately.
-	minRetryDelay = 5 * time.Second
-	maxRetryDelay = 300 * time.Second
+	minRetryDelay     = 5 * time.Second
+	maxRetryDelay     = 300 * time.Second
 
 	// The format we should expect for ccm worker firewall names.
 	firewallWorkerCCMNameFormat = "k8s-%s-ccm"
--- a/cloud-controller-manager/do/firewall_controller.go
+++ b/cloud-controller-manager/do/firewall_controller.go
@@ -256,7 +256,7 @@ func (fc *Controller) processFirewallCreateOrUpdate(ctx context.Context, key, wo
 }
 
 func (fc *Controller) processFirewallDelete(key, workerFirewallname string, service *v1.Service) error {
-	cachedFirewall := fc.fwCache.getOrCreate(service.Name)
+	cachedFirewall := fc.fwCache.getOrCreate(key)
 	// We do not care about other firewalls
 	if cachedFirewall.state.Name != workerFirewallName {
 		return nil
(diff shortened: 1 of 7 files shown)


cloud-controller-manager/do/firewall_controller.go
```diff
@@ -0,0 +1,359 @@
...
+	fw, _, _ := fc.firewallService.Get(ctx, cachedFirewall.state.ID)
+	if fw == nil {
+		// Create a new firewall with the worker firewall name and inbound rules for NodePort range.
+		fw, err := fc.create(ctx, workerFirewallName, fr, rr)
+		if err != nil {
+			return fmt.Errorf("Failed to create firewall: %s", err)
+		}
+		cachedFirewall.state = fw
+		return nil
+	} else if fw.Name == workerFirewallName {
```
timoreimann, line 243: I think we should ignore the firewall name: if it doesn't match, then our update logic should simply overwrite it with the target name. The firewall ID we store should be enough to uniquely identify a firewall (unless it has been deleted by the user, in that case we're going to recreate it under the target name).

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+func (fc *Controller) create(ctx context.Context, workerFirewallName string, fr *godo.FirewallRequest, rr *godo.FirewallRulesRequest) (*godo.Firewall, error) {
+	// Create the firewall.
+	fw, _, err := fc.firewallService.Create(ctx, fr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create firewall: %s", err)
+	}
+	// Add the inbound rules to the firewall.
+	fc.firewallService.AddRules(ctx, fw.ID, rr)
```
timoreimann, line 279: It looks like this makes another call to the API just to add the rules. I think we can include it on the initial create/update request already?

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -0,0 +1,359 @@
...
+		// Create a new firewall with the worker firewall name and inbound rules for NodePort range.
+		fw, err := fc.create(ctx, workerFirewallName, fr, rr)
+		if err != nil {
+			return fmt.Errorf("Failed to create firewall: %s", err)
+		}
+		cachedFirewall.state = fw
+		return nil
+	} else if fw.Name == workerFirewallName {
+		// Update the existing firewall to include inbound rules for NodePort range.
+		fw, err := fc.update(ctx, fw, rr)
```
timoreimann, line 245: I think we should only send out an update request after comparing the current firewall rules to the cached state and verifying that there's a diff. We're watching for changes on Services, but not every Service change is going to be relevant for us. The extra check will make sure we won't do unnecessary API calls.

Of course it means we'll have to know the initial state of the firewall. Since we want to have another controller to watch for changes to the firewall resource in the DO API anyway, we should get the data for sure.

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -0,0 +1,359 @@
...
+
+func (fc *Controller) processFirewallDelete(key, workerFirewallname string, service *v1.Service) error {
+	cachedFirewall := fc.fwCache.getOrCreate(key)
+	// We do not care about other firewalls
+	if cachedFirewall.state.Name != workerFirewallName {
+		return nil
+	}
+	_, err := fc.firewallService.Delete(context.TODO(), cachedFirewall.state.ID)
+	if err != nil {
+		return err
```
timoreimann, line 266: This should handle the case where the firewall is already gone (404) for whatever reason. In that case, it's not an error.

MorrisLaw:
closing this for now, so I don't spam everyone's inbox with updates. have to do some re-designing of logic and add tests but want my branch to stay updated.

PR #337, opened 2020-08-20
Expose prometheus metrics

MorrisLaw:
This PR adds prometheus metrics so that we can expose metrics related to firewall API requests and firewall controller logic.

i.e.
- ~request_count~
- request_duration
- run_loop ~_count~_duration
- reconcile ~_count~_duration


cloud-controller-manager/do/cloud.go
```diff
@@ -201,6 +217,28 @@ func (c *cloud) serveDebug(stop <-chan struct{}) {
...
+	prometheus.MustRegister(successfulRequestCounter)
+	prometheus.MustRegister(failedRequestCounter)
+	prometheus.MustRegister(runLoopsStarted)
+	prometheus.MustRegister(numberOfReconciles)
+	prometheus.MustRegister(createRequestDuration)
+	prometheus.MustRegister(getRequestDuration)
+	prometheus.MustRegister(updateRequestDuration)
+	prometheus.MustRegister(eventHandlerDuration)
+
+	if err := http.ListenAndServe(c.metrics.host, nil); err != http.ErrServerClosed {
```
timoreimann, line 237: I suppose this will fail if `c.metrics.host` is empty, which AFAICS would be legitimate in case metrics are disabled. We should gate the entire metrics setup in this method on the host variable being non-empty.

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -38,9 +38,9 @@ import (
 
 const (
 	// Interval of synchronizing service status from apiserver.
-	serviceSyncPeriod = 30 * time.Second
+	serviceSyncPeriod = 1 * time.Second
```
timoreimann, line 41: Needs to be undone.

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -38,9 +38,9 @@ import (
...
 	// Frequency at which the firewall controller runs.
-	firewallReconcileFrequency = 5 * time.Minute
+	firewallReconcileFrequency = 1 * time.Second
```
timoreimann, line 43: Needs to be undone.

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -101,37 +103,51 @@ func NewFirewallController(
...
 
 	serviceInformer.Informer().AddEventHandlerWithResyncPeriod(
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: func(cur interface{}) {
 				ctx, cancel := context.WithTimeout(context.Background(), 90*time.Second)
 				defer cancel()
-				err := fc.ensureReconciledFirewall(ctx)
+				var err error
+				defer func(start time.Time) {
+					metrics.eventHandlerDuration.Observe(time.Since(start).Seconds())
```
timoreimann, line 124: I think the usage of [`ObserveDuration`](https://godoc.org/github.com/prometheus/client_golang/prometheus#Timer.ObserveDuration) could simplify this further.

cloud-controller-manager/do/metrics.go
```diff
@@ -0,0 +1,133 @@
...
+	numberOfReconciles       prometheus.Counter
+	createRequestDuration    prometheus.Histogram
+	getRequestDuration       prometheus.Histogram
+	updateRequestDuration    prometheus.Histogram
+	eventHandlerDuration     prometheus.Histogram
+}
+
+// create metrics
+var (
+	apiRequestCounter = prometheus.NewCounter(
```
timoreimann, line 39: This metric seems unused?

timoreimann (commented):
Looking pretty solid already. 👍

Left some comments. We should also extend the docs and update the change log.

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -372,3 +409,38 @@ func (fc *firewallCache) updateCache(currentFirewall *godo.Firewall) {
...
+		labels = prometheus.Labels{
+			"http_code":   strconv.FormatInt(int64(resp.StatusCode), 10),
+			"http_method": resp.Request.Method,
+			"success":     strconv.FormatBool((resp.StatusCode >= 200 && resp.StatusCode < 300)),
+		}
+	} else {
+		labels = prometheus.Labels{"http_code": "", "http_method": "", "success": "false"}
+	}
+	fm.incrementPrometheusCounter(resp, &absentResponse{expected: false})
+	t := prometheus.NewTimer(fm.metrics.apiRequestDuration.With(labels))
```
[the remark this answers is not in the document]
MorrisLaw: Nice catch. Because of the need to start the timer before the operation starts and the way I'm deriving labels from the response (which comes from performing the operation in order to get the response to pass into a helper function):
```go
func (fm *firewallManager) observeDuration(resp *godo.Response) {
	var labels prometheus.Labels
	if resp != nil && resp.Request != nil {
		labels = prometheus.Labels{
			"http_code":   strconv.FormatInt(int64(resp.StatusCode), 10),
			"http_method": resp.Request.Method,
			"success":     strconv.FormatBool((resp.StatusCode >= 200 && resp.StatusCode < 300)),
		}
	} else {
		labels = prometheus.Labels{"http_code": "", "http_method": "", "success": "false"}
	}
	fm.incrementPrometheusCounter(resp, &absentResponse{expected: false})
	t := prometheus.NewTimer(fm.metrics.apiRequestDuration.With(labels))
	t.ObserveDuration()
}
```

Because of this, it's a little trickier than just starting the timer before the operation when we get labels based off of the operation. I'll have to refactor the way I make labels.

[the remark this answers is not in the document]
MorrisLaw: @timoreimann ^

cloud-controller-manager/do/metrics.go
```diff
@@ -0,0 +1,79 @@
...
+			Namespace: "firewall",
+			Name:      "api_request_total",
+			Help:      "This is a counter for tracking firewall API requests.",
+		},
+		[]string{"http_code", "http_method", "success"},
+	)
+	apiRequestDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "firewall",
+			Name:      "api_request_duration",
```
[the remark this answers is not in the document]
MorrisLaw: I'm going to change the other one (`api_request_total`) to plural. And make this one `api_request_duration_seconds`. This better aligns with the naming guidelines. And since this is `http`, I will update `api` to be `http` instead.

Thank you for the link @timoreimann !

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -372,3 +409,38 @@ func (fc *firewallCache) updateCache(currentFirewall *godo.Firewall) {
...
+func (fm *firewallManager) observeDuration(resp *godo.Response) {
+	var labels prometheus.Labels
+	if resp != nil && resp.Request != nil {
+		labels = prometheus.Labels{
+			"http_code":   strconv.FormatInt(int64(resp.StatusCode), 10),
+			"http_method": resp.Request.Method,
+			"success":     strconv.FormatBool((resp.StatusCode >= 200 && resp.StatusCode < 300)),
+		}
+	} else {
+		labels = prometheus.Labels{"http_code": "", "http_method": "", "success": "false"}
```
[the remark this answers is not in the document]
MorrisLaw: I don't think this is allowed: ``` Observed a panic: &errors.errorString{s:"inconsistent label cardinality: expected 1 label values but got 0 in prometheus.Labels(nil)"} (inconsistent label cardinality: expected 1 label values but got 0 in prometheus.Labels(nil))```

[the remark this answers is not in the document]
timoreimann: Hmm. To be clear, I meant that you might be able to omit `http_code` and `http_method`  but still need to specify `prometheus.Labels`. But maybe I'm wrong.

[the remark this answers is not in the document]
MorrisLaw: Looks to be the case when omitting: ```panic: inconsistent label cardinality: expected 4 label values but got 1 in prometheus.Labels{"firewall_loop":"true"} [recovered]
        panic: inconsistent label cardinality: expected 4 label values but got 1 in prometheus.Labels{"firewall_loop":"true"}```

I'm just going to have to populate the values.

cloud-controller-manager/do/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+			Buckets:   prometheus.ExponentialBuckets(0.01, 2, 14),
+		},
+		[]string{"method", "code"},
+	)
+	runLoopDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "firewall",
+			Name:      "run_loop_duration_seconds",
+			Help:      "This is a histogram for tracking the duration of the run loop and whether it succeeds or not.",
+			Buckets:   prometheus.ExponentialBuckets(0.01, 2, 15),
```
waynr, line 44: In one of Timo's comments on a previous iteration of this PR he mentioned

> What I think would be more interesting to have is the duration of a loop iteration (similar to how we have eventHandlerDuration). If we then also include a label indicating success/failure, then I believe we'd get a good picture of how our Run() loop is performing. The histogram buckets should then be configured such that we can cover up to the maximum run frequency (5 minutes IIRC).

So it sounds to me like he suggested the buckets should be able to capture to to a maximum of 5 minute duration here? This current exponential bucket only creates buckets up to 163 seconds:

```
>>> for each in range(1, 15):
...   print("{}".format(0.01*2**each))
...
0.02
0.04
0.08
0.16
0.32
0.64
1.28
2.56
5.12
10.24
20.48
40.96
81.92
163.84
```

But I'm probably misunderstanding something important.

cloud-controller-manager/do/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+		[]string{"success"},
+	)
+	reconcileDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "firewall",
+			Name:      "reconcile_duration_seconds",
+			Help:      "The duration of time, in seconds, that it takes for the firewall reconcile to run and ensure that the firewall is reconciled.",
+			Buckets:   prometheus.ExponentialBuckets(0.01, 2, 14),
+		},
+		[]string{"reconcile_type"},
```
waynr, line 55: Is this the reconcile origin mentioned by Timo in one of his comments?

> The ability to break down reconciles by origin sounds useful to me as well.

cloud-controller-manager/do/metrics.go
waynr: Taking another look at [how `firewallReconcileFrequency` is used in the `actualRun` method](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/9dd809f3ea45f8cfa09826f262bc98979cf90398/cloud-controller-manager/do/firewall_controller.go#L158) confirms my suspicion that we should this such that we have meaningful buckets throughout the range of 0-300 seconds. Again I would suggest using a linear distribution with bucket boundaries that align with any SLOs we might care about.

cloud-controller-manager/do/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+		[]string{"success"},
+	)
+	reconcileDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "firewall",
+			Name:      "reconcile_duration_seconds",
+			Help:      "The duration of time, in seconds, that it takes for the firewall reconcile to run and ensure that the firewall is reconciled.",
+			Buckets:   prometheus.ExponentialBuckets(0.01, 2, 14),
+		},
+		[]string{"event_add", "event_update", "event_delete", "firewall_loop"},
```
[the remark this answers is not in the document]
MorrisLaw: Updating this to be `reconcile_type` based on Timo's suggestion earlier today.
>This should be a single label name (such as reconcile_type) with four different label values to ease filtering and aggregating.

cloud-controller-manager/do/metrics.go
MorrisLaw: No, this was an update based on his earlier comment from today.
> This should be a single label name (such as reconcile_type) with four different label values to ease filtering and aggregating.

MorrisLaw: I interpreted as there being 4 possible values (5 if you include a default `"none"` value), 'event_add', 'event_update', 'event_delete' and 'firewall_loop'.

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -372,3 +493,38 @@ func (fc *firewallCache) updateCache(currentFirewall *godo.Firewall) {
 	defer fc.mu.Unlock()
 	fc.firewall = currentFirewall
 }
+
+func (fc *FirewallController) observeReconcileDuration(ctx context.Context, origin string) error {
+	var labels prometheus.Labels
+	switch {
+	case origin == "event_add":
```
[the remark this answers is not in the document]
MorrisLaw: agreed

cloud-controller-manager/do/metrics.go
MorrisLaw: This should have been 16 in order to hit `327.68`. But yeah I think linear distribution may make more sense in this case too. I didn't imagine those first 9 or 10 numbers to be too helpful either. Good suggestion.

cloud-controller-manager/do/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// create metrics
+var (
+	apiRequestDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "firewall",
+			Name:      "http_request_duration_seconds",
+			Help:      "This is a histogram for tracking the duration of firewall API requests.",
+			Buckets:   prometheus.ExponentialBuckets(0.01, 2, 14),
```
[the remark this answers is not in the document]
MorrisLaw: I'm not opposed to using `prometheus.LinearBuckets..` in order to have more meaningful metrics. Especially with the concerns you've pointed out.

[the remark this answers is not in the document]
MorrisLaw: I think 90/20 makes sense, for the main reason that I think we'd prefer the little bit of extra granularity that, that provides.

cloud-controller-manager/do/metrics.go
```diff
@@ -0,0 +1,71 @@
...
+			Buckets:   prometheus.LinearBuckets(1, width, 21),
+		},
+		[]string{"success"},
+	)
+	reconcileDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "firewall",
+			Name:      "reconcile_duration_seconds",
+			Help:      "The duration of time, in seconds, that it takes for the firewall reconcile to run and ensure that the firewall is reconciled.",
+			Buckets:   prometheus.LinearBuckets(1, width, 21),
```
waynr, line 58: I'm not sure what we expect the range to be for `reconcile_duration_seconds` but given that `run_loop_duration_seconds` bucket definitions were copied from `http_request_duration_seconds` I would feel better if you dug in to understand what the expected range should be for that one as well. If you can then confirm that this bucket definition will be fine I'll be :+1: here.

MorrisLaw: yeah this should be different, i'll update. maybe with some comments too

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -372,3 +493,38 @@ func (fc *firewallCache) updateCache(currentFirewall *godo.Firewall) {
 	defer fc.mu.Unlock()
 	fc.firewall = currentFirewall
 }
+
+func (fc *FirewallController) observeReconcileDuration(ctx context.Context, origin string) error {
+	var labels prometheus.Labels
+	switch {
```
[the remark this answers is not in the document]
MorrisLaw: This is a great optimization @waynr ! 

I'll opt for the switch case with default, just so we have something other than `""` in those weird edge cases. Thank you!

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -372,3 +493,38 @@ func (fc *firewallCache) updateCache(currentFirewall *godo.Firewall) {
...
+}
+
+func (fc *FirewallController) observeRunLoopDuration(ctx context.Context, stopCh <-chan struct{}, fwReconcileFreq time.Duration) error {
+	labels := prometheus.Labels{"success": "false"}
+	t := prometheus.NewTimer(fc.fwManager.metrics.runLoopDuration.With(labels))
+	defer t.ObserveDuration()
+
+	err := fc.actualRun(stopCh, fwReconcileFreq)
+	if err == nil {
+		labels = prometheus.Labels{"success": "true"}
```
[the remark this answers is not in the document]
MorrisLaw: Thank you for the example, that makes sense. Setting the variable again doesn't change `labels` before ObserveDuration is run, but because map is a reference type, changing that will have the desired effect.

cloud-controller-manager/do/metrics.go
MorrisLaw: Run and Reconcile will have the same range. Because of the fwReconcileFrequency value which is `5 * time.Minute`. So, both `run_loop_duration_seconds` and `reconcile_duration_seconds` should have a width of `300/20` as pointed out in your previous comment. I've updated the code to address this.

Also, I agree. `http_request_duration_seconds` should remain `90 * time.Second`.

Thank you!

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -182,12 +194,37 @@ func (fm *firewallManager) Get(ctx context.Context) (*godo.Firewall, error) {
...
+			// The ObserverFunc gets called by the deferred ObserveDuration. The
+			// method and code values will be set before ObserveDuration is called
+			// with the value returned from the response from the Firewall API request.
+			timer := prometheus.NewTimer(prometheus.ObserverFunc(func(v float64) {
+				fm.metrics.apiRequestDuration.With(prometheus.Labels{"method": method, "code": strconv.FormatInt(int64(code), 10)}).Observe(v)
+			}))
+			defer timer.ObserveDuration()
+			fw, resp, err := fm.client.Firewalls.Get(ctx, fw.ID)
+			if resp != nil {
+				code = resp.StatusCode
```
waynr, line 215: Whoops, I just about overlooked the fact that here we are also reassigning a variable whose value has already been captured by the intended `Observer` (refer the explanation I gave in my previous review).

To fix this, instead of creating `code` and `method` variables I would do the same thing that is being done in `observeRunLoopDuration` and use a `prometheus.Labels` variable whose value you would update here (ie `labels["code"] = resp.StatusCode`). This comment applies everywhere the same timer/observer pattern is being applied, ie the changes starting at lines 239, 287, 372, and 398.

waynr: Actually, forget my previous comment I see now that you're using `prometheus.ObserverFunc` with a callback that will capture the values of `code` and `method` after they have been set and before the deferred `timer.ObserveDuration` has run.

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -372,3 +493,30 @@ func (fc *firewallCache) updateCache(currentFirewall *godo.Firewall) {
...
+}
+
+func (fc *FirewallController) observeRunLoopDuration(ctx context.Context, stopCh <-chan struct{}, fwReconcileFreq time.Duration) error {
+	labels := prometheus.Labels{"success": "false"}
+	t := prometheus.NewTimer(fc.fwManager.metrics.runLoopDuration.With(labels))
+	defer t.ObserveDuration()
+
+	err := fc.actualRun(stopCh, fwReconcileFreq)
+	if err == nil {
+		labels["success"] = "true"
```
waynr, line 518: So I dug down into the implementation of `With` just now and see that [the values are being extracted from the given `prometheus.Labels` at the time that `With` is called](https://github.com/prometheus/client_golang/blob/4675060215e9866ac3e834cfd62ba3fb36b89e12/prometheus/vec.go#L191).

Because of this I think you can wrap this `With` call with a `prometheus.ObserverFunc` the way you are doing above to properly capture the value of labels at the time the deferred `t.ObserveDuration()` call is run. Sorry again for the churn!

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -290,12 +369,54 @@ func isEqual(targetFirewall *godo.Firewall, fr *godo.FirewallRequest) (bool, []s
...
+		}
+		return fw, err
+	}()
+
 	return currentFirewall, err
 }
 
 func (fm *firewallManager) updateFirewall(ctx context.Context, fwID string, fr *godo.FirewallRequest) (*godo.Firewall, error) {
-	currentFirewall, _, err := fm.client.Firewalls.Update(ctx, fwID, fr)
+	currentFirewall, err := func() (*godo.Firewall, error) {
```
waynr, line 398: nit: I think the reason you're using a callback approach in `firewallManager.Get` is that more is happening in that function than simply calling out to the firewalls API whereas here in `firewallManager.updateFirewall` we are simply wrapping a single call to the firewalls API; removing the callback would lead to a somewhat cleaner implementation here.

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -290,12 +369,54 @@ func isEqual(targetFirewall *godo.Firewall, fr *godo.FirewallRequest) (bool, []s
 }
 
 func (fm *firewallManager) createFirewall(ctx context.Context, fr *godo.FirewallRequest) (*godo.Firewall, error) {
-	currentFirewall, _, err := fm.client.Firewalls.Create(ctx, fr)
+	currentFirewall, err := func() (*godo.Firewall, error) {
```
waynr, line 372: nit: I think the reason you're using a callback approach in `firewallManager.Get` is that more is happening in that function than simply calling out to the firewalls API whereas here in `firewallManager.createFirewall` we are simply wrapping a single call to the firewalls API; removing the callback would lead to a somewhat cleaner implementation here.

cloud-controller-manager/do/firewall_controller.go
MorrisLaw: nice catch!

README.md
```diff
@@ -90,6 +90,41 @@ Please note that if you use a kubernetes cluster created on DigitalOcean, there
...
+passing it in as the `METRICS_ADDR`.
+
+```bash
+cd cloud-controller-manager/cmd/digitalocean-cloud-controller-manager
+FAKE_REGION=fra1 DO_ACCESS_TOKEN=your_access_token METRICS_ADDR=<host:port> go run main.go      \
+  --kubeconfig <path to your kubeconfig file>                                                   \
+  --leader-elect=false --v=5 --cloud-provider=digitalocean
+```
+
+After you have started up the CCM, run the following curl command if you want to view the prometheus metrics output. In order to illustrate the example, we'll assume you've set `METRICS_ADDR` to `localhost:8080`.
```
waynr, line 119: s/if you want//

Also you should consider sticking with the same placeholder value (`<host:port>`) for `METRICS_ADDR` as what you used in the `go run main.go` example above; then you don't need to say "In order to illustrate the example, we'll assume you've set `METRICS_ADDR` to `localhost:8080`." at all

README.md
```diff
@@ -90,6 +90,41 @@ Please note that if you use a kubernetes cluster created on DigitalOcean, there
...
+
+```bash
+cd cloud-controller-manager/cmd/digitalocean-cloud-controller-manager
+FAKE_REGION=fra1 DO_ACCESS_TOKEN=your_access_token PUBLIC_ACCESS_FIREWALL_NAME=firewall_name       \
+PUBLIC_ACCESS_FIREWALL_TAGS=k8s,k8s:8847f3c3-387f-4795-b794-b2f7d9f7d22f,k8s:worker go run main.go \
+  --kubeconfig <path to your kubeconfig file>                                                      \
+  --leader-elect=false --v=5 --cloud-provider=digitalocean
+```
+
+If you want to expose prometheus metrics, be sure to include the endpoint by
```
waynr, line 109: I would create a subsection here named `Expose Prometheus Metrics` or something to that effect. 

Then you might write:

To expose a prometheus metrics endpoint:

```
... example ...
```

The `METRICS_ADDR` environment variable takes a colon-separated host:port address pair; both `host` and `port` parts are required.

README.md
```diff
@@ -90,6 +90,41 @@ Please note that if you use a kubernetes cluster created on DigitalOcean, there
 will be a cloud controller manager running in the cluster already, so you local
 one will compete for API access with it.
 
+### Run Locally (optional parameters)
+
+In addition to the standard way of running `digitalocean-cloud-controller-manager`
+locally against a particular cluster, you can add some additional parameters that
+enable you to run the cluster with some extra features.
```
waynr, line 97: If we call this entire section `Optional Features` I think we can get rid of this paragraph has a kind of filler feel to it.

README.md
```diff
@@ -90,6 +90,58 @@ Please note that if you use a kubernetes cluster created on DigitalOcean, there
...
+If you want to add an additional firewall, that allows public access to your
+cluster, you can run a command like this:
+
+```bash
+cd cloud-controller-manager/cmd/digitalocean-cloud-controller-manager
+FAKE_REGION=fra1 DO_ACCESS_TOKEN=your_access_token            \
+PUBLIC_ACCESS_FIREWALL_NAME=firewall_name                     \
+PUBLIC_ACCESS_FIREWALL_TAGS=k8s,k8s:<cluster-uuid>,k8s:worker \
+go run main.go                                                \
+  --kubeconfig <path to your kubeconfig file>
```
waynr, line 106: nit: while I recognize this is the established way of writing commands, I think it's actually harder to read to try to fit everything on a single command line call using line continuations like this, but as long as we're doing it we should also add one on this line. 

I will submit a PR after yours is merged to propose an alternate way of writing this that is easier on the eyes.

MorrisLaw:
thank you @waynr @timoreimann. will rebase, squash then merge 👍

PR #338, opened 2020-08-21
firewalls: add exponential backoff to service informer

MorrisLaw:
This PR adds exponential backoff as the retry mechanism for the firewall controller. Specifically, the service informer when handling event notifications for:
- `Add` service
- `Update` service
- `Delete` service

### DIFF
--- a/cloud-controller-manager/do/firewall_controller.go
+++ b/cloud-controller-manager/do/firewall_controller.go
@@ -34,6 +34,7 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	corelisters "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 )
 
@@ -42,10 +43,12 @@ const (
 	serviceSyncPeriod = 30 * time.Second
 	// Frequency at which the firewall controller runs.
 	firewallReconcileFrequency = 5 * time.Minute
-	originEventAdd             = "event_add"
-	originEventUpdate          = "event_update"
-	originEventDelete          = "event_delete"
+	originEvent                = "event"
 	originFirewallLoop         = "firewall_loop"
+
+	// How long to wait before retrying the processing of a firewall change.
+	minRetryDelay = 1 * time.Second
+	maxRetryDelay = 5 * time.Minute
 )
 
 var (
@@ -97,6 +100,7 @@ type FirewallController struct {
 	serviceLister      corelisters.ServiceLister
 	fwManager          firewallManager
 	metrics            metrics
+	queue              workqueue.RateLimitingInterface
 }
 
 // NewFirewallController returns a new firewall controller to reconcile public access firewall state.
@@ -117,36 +121,19 @@ func NewFirewallController(
 		workerFirewallName: workerFirewallName,
 		fwManager:          fwManager,
 		metrics:            metrics,
+		queue:              workqueue.NewNamedRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(minRetryDelay, maxRetryDelay), "firewall"),
 	}
 
 	serviceInformer.Informer().AddEventHandlerWithResyncPeriod(
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: func(cur interface{}) {
-				ctx, cancel := context.WithTimeout(context.Background(), 90*time.Second)
-				defer cancel()
-
-				err := fc.observeReconcileDuration(ctx, originEventAdd)
-				if err != nil {
-					klog.Errorf("failed to reconcile worker firewall after adding a resource object: %s", err)
-				}
+				fc.queue.Add("service")
 			},
 			UpdateFunc: func(old, cur interface{}) {
-				ctx, cancel := context.WithTimeout(context.Background(), 90*time.Second)
-				defer cancel()
-
-				err := fc.observeReconcileDuration(ctx, originEventUpdate)
-				if err != nil {
-					klog.Errorf("failed to reconcile worker firewall after updating a resource object: %s", err)
-				}
+				fc.queue.Add("service")
 			},
-			DeleteFunc: func(obj interface{}) {
-				ctx, cancel := context.WithTimeout(context.Background(), 90*time.Second)
-				defer cancel()
-
-				err := fc.observeReconcileDuration(ctx, originEventDelete)
-				if err != nil {
-					klog.Errorf("failed to reconcile worker firewall after deleting a resource object: %s", err)
-				}
+			DeleteFunc: func(cur interface{}) {
+				fc.queue.Add("service")
 			},
 		},
 		serviceSyncPeriod,
@@ -159,17 +146,42 @@ func NewFirewallController(
 // Run starts the firewall controller loop.
 func (fc *FirewallController) Run(ctx context.Context, stopCh <-chan struct{}, fwReconcileFrequency time.Duration) {
 	wait.Until(func() {
-		err := fc.observeRunLoopDuration(ctx, stopCh, fwReconcileFrequency)
+		err := fc.observeRunLoopDuration(ctx)
 		if err != nil {
 			klog.Errorf("failed to run firewall controller loop: %v", err)
 		}
 	}, fwReconcileFrequency, stopCh)
+	fc.queue.ShutDown()
 }
 
-func (fc *FirewallController) actualRun(stopCh <-chan struct{}, fwReconcileFrequency time.Duration) error {
-	ctx, cancel := context.WithTimeout(context.Background(), fwReconcileFrequency)
-	defer cancel()
+func (fc *FirewallController) runWorker() {
+	for fc.processNextItem() {
+	}
+}
 
+func (fc *FirewallController) processNextItem() bool {
+	ctx, cancel := context.WithTimeout(context.Background(), firewallReconcileFrequency)
+	defer cancel()
+	// Wait until there is a new item in the working queue
+	key, quit := fc.queue.Get()
+	if quit {
+		return false
+	}
+	// Tell the queue that we are done with processing this key. This unblocks the key for other workers
+	// This allows safe parallel processing because items with the same key are never processed in
+	// parallel.
+	defer fc.queue.Done(key)
+
+	err := fc.observeReconcileDuration(ctx, originEvent)
+	if err != nil {
+		klog.Errorf("failed to run firewall controller loop: %v", err)
+		fc.queue.AddRateLimited(key)
+	}
+	fc.queue.Forget(key)
+	return true
+}
+
+func (fc *FirewallController) reconcileCloudFirewallChanges(ctx context.Context) error {
 	currentFirewall, err := fc.fwManager.Get(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to get worker firewall: %s", err)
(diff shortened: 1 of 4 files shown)


timoreimann (changes requested):
The logic still seems to be missing a few things that you can see in https://github.com/kubernetes/client-go/blob/master/examples/workqueue/main.go, like forgetting / done-ing keys and setting up a worker goroutine.

waynr (dismissed):
So there are a couple problems I see with the implementation of the workqueue:

* You are adding items to the queue using `fc.enqueueService`, but never grabbing them off of it
* You are following the `fc.enqueueService` calls with calls to `fc.ensureReconciledFirewall` in the main goroutine which defeats the purpose of the queue

So the use of the queue here assumes that you have goroutines created that will asynchronously process items that have been pushed onto the queue in the `AddFunc`, `UpdateFunc`, and `DeleteFunc` handlers. To see what I mean consider the client-go example:

* here is an example of [a handler func adding an item to the queue](https://github.com/kubernetes/client-go/blob/20923fd71b14d068ee301eaf1c5320178b9ba585/examples/workqueue/main.go#L175-L180)
* here is [where the goroutine that spins up worker goroutines is created](https://github.com/kubernetes/client-go/blob/20923fd71b14d068ee301eaf1c5320178b9ba585/examples/workqueue/main.go#L213)
* here is [where each work queue item processing goroutine is created](https://github.com/kubernetes/client-go/blob/20923fd71b14d068ee301eaf1c5320178b9ba585/examples/workqueue/main.go#L132)
* here is [where each worker goroutine grabs an item off the queue](https://github.com/kubernetes/client-go/blob/20923fd71b14d068ee301eaf1c5320178b9ba585/examples/workqueue/main.go#L53)
* here is [the business logic of the example](https://github.com/kubernetes/client-go/blob/20923fd71b14d068ee301eaf1c5320178b9ba585/examples/workqueue/main.go#L63) (in our case this would be `fc.ensureReconciledFirewall`

waynr:
Oh whoops, I didn't refresh to see Timo's comment before submitting mine.

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -41,6 +43,10 @@ const (
 	serviceSyncPeriod = 30 * time.Second
 	// Frequency at which the firewall controller runs.
 	firewallReconcileFrequency = 5 * time.Minute
+
+	// How long to wait before retrying the processing of a firewall change.
+	minRetryDelay = 5 * time.Second
```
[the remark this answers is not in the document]
MorrisLaw: That's a reasonable suggestion.

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -117,35 +125,42 @@ func NewFirewallController(
...
 			},
-			DeleteFunc: func(obj interface{}) {
-				ctx, cancel := context.WithTimeout(context.Background(), 90*time.Second)
-				defer cancel()
-
-				err := fc.observeReconcileDuration(ctx, originEventDelete)
-				if err != nil {
-					klog.Errorf("failed to reconcile worker firewall after deleting a resource object: %s", err)
+			DeleteFunc: func(cur interface{}) {
+				key, err := cache.MetaNamespaceKeyFunc(cur)
```
timoreimann, line 156: I think we should use `DeletionHandlingMetaNamespaceKeyFunc` as in the [official example](https://github.com/kubernetes/client-go/blob/ba26c4506d39dece49d00e969d03794e8089b8dc/examples/workqueue/main.go#L188-L190).

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -156,20 +171,52 @@ func NewFirewallController(
 	return fc
 }
 
+// obj could be an *v1.Service, or a DeletionFinalStateUnknown marker item.
+func (fc *FirewallController) enqueueService(obj interface{}) {
```
timoreimann, line 175: This doesn't seem to be called from anywhere. Could we possibly drop it?

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -156,20 +171,52 @@ func NewFirewallController(
...
-		err := fc.observeRunLoopDuration(ctx, stopCh, fwReconcileFrequency)
-		if err != nil {
-			klog.Errorf("failed to run firewall controller loop: %v", err)
-		}
-	}, fwReconcileFrequency, stopCh)
+func (fc *FirewallController) Run(threadiness int, stopCh <-chan struct{}, fwReconcileFrequency time.Duration) {
+	for i := 0; i < threadiness; i++ {
+		go wait.Until(fc.runWorker, fwReconcileFrequency, stopCh)
+	}
+
```
timoreimann, line 189: We should [make sure that the queue is shut down](https://github.com/kubernetes/client-go/blob/ba26c4506d39dece49d00e969d03794e8089b8dc/examples/workqueue/main.go#L119-L120).

cloud-controller-manager/do/firewall_controller_test.go
```diff
@@ -491,7 +491,7 @@ func TestFirewallController_NoDataRace(t *testing.T) {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		fc.Run(context.TODO(), ctx.Done(), time.Duration(0))
+		fc.Run(1, ctx.Done(), time.Duration(0))
```
timoreimann, line 494: Currently I don't see us using more than one thread/goroutine for handling Service events in the context of the worker firewall: by design, we always look at all (NodePort) Services on each run.

Overall, I'd remove the `threadiness` variable and instantiate a single worker goroutine in the controller only.

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -97,6 +103,8 @@ type FirewallController struct {
 	serviceLister      corelisters.ServiceLister
 	fwManager          firewallManager
 	metrics            metrics
+	// nodeport services that need to be synced
```
timoreimann, line 106: Not sure I understand this comment. 🤔

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -156,20 +171,52 @@ func NewFirewallController(
...
+		return false
+	}
+	// Tell the queue that we are done with processing this key. This unblocks the key for other workers
+	// This allows safe parallel processing because items with the same key are never processed in
+	// parallel.
+	defer fc.queue.Done(key)
+
+	err := fc.observeRunLoopDuration(ctx)
+	if err != nil {
+		klog.Errorf("failed to run firewall controller loop: %v", err)
```
timoreimann, line 214: At this point we need to [add the key back into the queue](https://github.com/kubernetes/client-go/blob/ba26c4506d39dece49d00e969d03794e8089b8dc/examples/workqueue/main.go#L106).

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -156,20 +171,52 @@ func NewFirewallController(
...
+	}
+	return true
```
timoreimann, line 216: Before returning, we need to [make the queue forget the key](https://github.com/kubernetes/client-go/blob/ba26c4506d39dece49d00e969d03794e8089b8dc/examples/workqueue/main.go#L110).

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -156,20 +171,52 @@ func NewFirewallController(
...
+}
 
+func (fc *FirewallController) actualRun(ctx context.Context) error {
```
timoreimann, line 219: This method now lives fairly disconnected from `Run()`, so I think we should rename it. How about something like `reconcileCloudFirewallChanges`?

cloud-controller-manager/do/firewall_controller.go
MorrisLaw: oops, I think that was copy pasta'd from somewhere

MorrisLaw: will remove that

cloud-controller-manager/do/firewall_controller.go
MorrisLaw: Nice catch!

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -156,20 +171,52 @@ func NewFirewallController(
...
+	key, quit := fc.queue.Get()
+	if quit {
+		return false
+	}
+	// Tell the queue that we are done with processing this key. This unblocks the key for other workers
+	// This allows safe parallel processing because items with the same key are never processed in
+	// parallel.
+	defer fc.queue.Done(key)
+
+	err := fc.observeRunLoopDuration(ctx)
```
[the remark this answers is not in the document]
MorrisLaw: I added a commit to move things around @timoreimann , I hope it makes sense. It ended up being a little weird because of my reliance on adding the key when checking that error.

cloud-controller-manager/do/firewall_controller_test.go
MorrisLaw: > instantiate a single worker goroutine in the controller only

Are you suggesting to take out the threadiness variable and the loop that references it, and leave the `go wait.Until(func)...`.

_or_

Add a wrapping `go func()` so that this is included? @timoreimann

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -117,35 +123,42 @@ func NewFirewallController(
...
-				defer cancel()
-
-				err := fc.observeReconcileDuration(ctx, originEventAdd)
-				if err != nil {
-					klog.Errorf("failed to reconcile worker firewall after adding a resource object: %s", err)
+				key, err := cache.MetaNamespaceKeyFunc(cur)
+				if err == nil {
+					fc.queue.Add(key)
+
+					reconcileErr := fc.observeReconcileDuration(ctx, originEventAdd)
```
timoreimann, line 136: If you look at [how the official example works](https://github.com/kubernetes/client-go/blob/ba26c4506d39dece49d00e969d03794e8089b8dc/examples/workqueue/main.go#L175-L194), you'll notice that the `*Func` implementations don't do much anymore other than adding an item to the queue. It is the [`runWorker` loop](https://github.com/kubernetes/client-go/blob/ba26c4506d39dece49d00e969d03794e8089b8dc/examples/workqueue/main.go#L139-L142) which keeps taking items off of that queue and processes them.

(Speaking of, I think we used to have a `runWorker` method at some point but it looks like we lost it. I believe we need to get it back for things to work correctly.)

This means that we need to drop the `observeReconcileDuration` calls from our handlers. The one downside though  is that we lose the information as to which event we processed. I'm not exactly sure if there's a way to pass along context to an item to put onto the queue, we'll need to dig.

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -158,18 +171,37 @@ func NewFirewallController(
...
 	}, fwReconcileFrequency, stopCh)
+
+	<-stopCh
+	klog.Info("Stopping firewall controller")
 }
 
-func (fc *FirewallController) actualRun(stopCh <-chan struct{}, fwReconcileFrequency time.Duration) error {
-	ctx, cancel := context.WithTimeout(context.Background(), fwReconcileFrequency)
-	defer cancel()
+func (fc *FirewallController) processNextItemAndCheckError(observeRunLoopDurationErr error) bool {
```
timoreimann, line 185: I'm not sure the implementation we have here makes sense. The idea is to pick an item off of the queue and process it, but we do not seem to work with the `key` at all.

timoreimann (commented):
This looks great to me. 👍 

Could we do a quick manual test to verify that things work as expected? I'm thinking of making a local tweak to let queue-driven reconciliation loops fail (e.g., by modifying [this line](https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/338/files#diff-92ac4af525e02d85b1ee8eb33e17ad3bR184) temporarily) and trigger a Service change afterwards. If all things go as expected, we should see the retry intervals increase steadily until we hit the 5 min max limit.

MorrisLaw:
Confirmed the retry logic _is_ exponentially retrying by running CCM locally with a stage2 cluster and monitoring the log output for an exponential increase in time between retries @timoreimann

timoreimann (approved):
LGTM, thanks Jeremy. 🎉 

Please amend the change log real quick, and then merge at will. 👍

PR #339, opened 2020-09-10
Update Kubernetes dependency to 1.19.1

adamwg:
Kubernetes 1.19 introduces the `InstancesV2()` method for providers. Add a dummy implementation for now - we could implement it, but there's no need to do it immediately.


timoreimann (approved):
👏 LGTM.

Could you do a quick update to the change log file as well?

PR #340, opened 2020-09-17
Add addresses instead of tags for inbound rule sources

MorrisLaw:
Discovered a bug where we write tags instead of addresses to the firewall inbound rules when creating the inbound rules. So allow all access to that nodeport did not work.

This PR:
- adds addresses to inbound rule sources, instead of tags
- updates the tests to check for addresses instead of tags when testing inbound rule objects


MorrisLaw:
Confirmed this works by creating an nginx deployment on a cluster I have running, added a nodeport service. Then ran curl against that port.

timoreimann (approved):
:+1:

PR #342, opened 2020-10-01
Remove unused mutex

gottwald:
This mutex seems to be a leftover from a former cache implementation
that no longer exists in the code.


timoreimann (approved):
Good catch -- thanks Ingo 🦈

PR #344, opened 2020-10-02
Expose health check failures

timoreimann:
We use the `healthz` package from `k8s.io/apiserver` to implement our health check. However, that check does not expose the failure reason in the logs. Example output:

```
I1002 08:05:39.882128       1 healthz.go:200] [-]godo failed: reason withheld
```

As described in [an upstream code comment](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/f3ea5608898f253f7bfd85de35f166dde36a278a/vendor/k8s.io/apiserver/pkg/server/healthz/healthz.go#L226-L227), the rationale here is that the endpoint is public and not supposed to give away sensitive data. Instead, Kubernetes users are [expected to inspect failure details through separate, authenticated access](https://stackoverflow.com/a/53861547).

This makes the package a bad fit for our use case because (1) we do not have an alternative way to inspect errors and (2) we do not need to: users concerned about the information that is being returned from the health check implementation (currently the respone from the `/v2/account` endpoint of the DigitalOcean API) can configure CCM's debug endpoint to run on localhost.

Consequently, we re-implement the health check as a custom, plain net/http handler and log health check failures explicitly.

### DIFF
--- a/cloud-controller-manager/do/health_test.go
+++ b/cloud-controller-manager/do/health_test.go
@@ -17,67 +17,74 @@ limitations under the License.
 package do
 
 import (
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
 	"github.com/digitalocean/godo"
 )
 
-func TestGodoHealthChecker_Name(t *testing.T) {
-	c := godoHealthChecker{}
-	if want, got := "godo", c.Name(); want != got {
-		t.Errorf("incorrect name\nwant: %#v \n got: %#v", want, got)
+func TestGodoHealthChecker(t *testing.T) {
+	tests := []struct {
+		name        string
+		stubHandler func(http.ResponseWriter, *http.Request)
+		wantCode    int
+		wantBody    string
+	}{
+		{
+			name: "healthy",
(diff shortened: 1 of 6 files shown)


cloud-controller-manager/do/health_test.go
```diff
@@ -17,67 +17,74 @@ limitations under the License.
...
 	"testing"
 
 	"github.com/digitalocean/godo"
 )
 
-func TestGodoHealthChecker_Name(t *testing.T) {
-	c := godoHealthChecker{}
-	if want, got := "godo", c.Name(); want != got {
-		t.Errorf("incorrect name\nwant: %#v \n got: %#v", want, got)
+func TestGodoHealthChecker(t *testing.T) {
```
timoreimann, line 29: Note to reviewers: the pre-existing two tests were merged into a single test with two test cases.

MorrisLaw (approved):
👍🏽

waynr (approved):
Curious if this would have any security implications in a future where we end up attaching the master kubelet to the apiserver. I suppose probably not beyond what the existing stdout logs are already emitting.

PR #345, opened 2020-10-02
Fix klog log level

timoreimann:
Currently, `klog` seems to be pinned to level 0 even when specifying `--v`, meaning that logs higher than `V(0)` are never emitted. After investigating, it turned out that our update script did not fully vendor all k8s.io dependencies causing our copy of `klog` (v0.4.0) to be significantly behind what Kubernetes 1.19 actually uses (v2.2.0). See the related commit message for details.

This change fixes the `update-k8s.sh` script, updates the Kubernetes dependencies to 1.19.2, and updates the code to use `klog/v2`. I verified that `klog` works correctly again afterwards.


go.mod
```diff
@@ -14,57 +14,67 @@ require (
...
+replace k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.19.2
 
-replace k8s.io/kubectl => k8s.io/kubectl v0.19.1
+replace k8s.io/kubectl => k8s.io/kubectl v0.19.2
 
-replace k8s.io/klog => k8s.io/klog v0.4.0
+replace k8s.io/kubernetes => k8s.io/kubernetes v1.19.2
 
-replace k8s.io/kubernetes => k8s.io/kubernetes v1.19.1
+replace k8s.io/gengo => k8s.io/gengo v0.0.0-20200428234225-8167cfdcfc14
```
timoreimann, line 70: Note to reviewers: the replacements below are the ones we were missing.

MorrisLaw (approved):
🚢

waynr (approved):

PR #346, opened 2020-10-02
load balancers: add new size slug field

anitgandhi:
## Changes

This adds a new annotation for DO LBs for setting the size slug, which is a new feature currently not yet generally not available. 

It defaults to the slug `lb-small`. 

## Notes

Internal ticket number: LBAAS-1582

This depends on https://github.com/digitalocean/godo/pull/373

## TODO

- [x] After the godo PR is merged and a release is cut, update go.mod here.


docs/controllers/services/annotations.md
```diff
@@ -101,6 +101,10 @@ Specifies the hostname used for the Service `status.Hostname` instead of assigni
 
 Specifies which algorithm the Load Balancer should use. Options are `round_robin`, `least_connections`. Defaults to `round_robin`.
 
+## service.beta.kubernetes.io/do-loadbalancer-size-slug
+
+Specifies which size the Load Balancer should be created with. Options are `lb-small`, `lb-medium`, and `lb-large`. Defaults to `lb-small`.
```
timoreimann, line 106: If we plan to release the next CCM by the team the slug sizes will already be available backend-wise, then I think this update to the annotation docs suffices. For the case where it might still take us some time to honor the new parameter, however, I'd suggest to merge this change with a small comment saying that the feature might not be available yet.

The context of this comment here is that many DOKS customers rely on the annotation docs to reflect the current state, so we should make sure there's no confusion for customers that try it out with the next CCM release we ship but don't immediately see that the feature won't be effective yet.

Let's clarify on the timeline internally.

timoreimann (commented):
Two quick comments, but otherwise this looks great to me. 👍 

We'll keep the PR open until godo has been updated and released.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -1093,6 +1104,21 @@ func getAlgorithm(service *v1.Service) string {
...
+
+	if !ok || sizeSlug == "" {
+		sizeSlug = "lb-small"
+	}
+
+	switch sizeSlug {
+	case "lb-small", "lb-medium", "lb-large":
+		return sizeSlug, nil
+	default:
+		return "", fmt.Errorf("invalid LB size slug provided")
```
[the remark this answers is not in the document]
anitgandhi: Fixed 👍

docs/controllers/services/annotations.md
anitgandhi: Yup that makes sense; I have no issue with keeping this blocked until the time comes 👍

anitgandhi: Resolving this for now, I've added a note to the annotation that this feature is in closed beta.

timoreimann (commented):
LGTM 👍

timoreimann (approved):

PR #349, opened 2020-10-06
Create context after retrieving item from worker queue

MorrisLaw:
Getting `context deadline exceeded` errors within the Firewall Controller. Looks to be because of how we currently initialize the context in a particular part of the worker item logic.

This PR:
- updates the logic to create the context after an item is retrieved from the queue
- adds a timeout value to be used within the context that's created within the process worker item logic
- adds a better error message, relating to the fact we failed to retrieve an item from the worker queue

### DIFF
--- a/cloud-controller-manager/do/firewall_controller.go
+++ b/cloud-controller-manager/do/firewall_controller.go
@@ -43,8 +43,10 @@ const (
 	serviceSyncPeriod = 30 * time.Second
 	// Frequency at which the firewall controller runs.
 	firewallReconcileFrequency = 5 * time.Minute
-	originEvent                = "event"
-	originFirewallLoop         = "firewall_loop"
+	// Timeout value for processing worker items taken from the queue.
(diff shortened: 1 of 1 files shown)


timoreimann (approved):
LGTM 👍

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -43,6 +43,8 @@ const (
 	serviceSyncPeriod = 30 * time.Second
 	// Frequency at which the firewall controller runs.
 	firewallReconcileFrequency = 5 * time.Minute
+	// Frequency at which we process worker items taken from the queue.
+	processWorkerItemFrequency = 30 * time.Second
```
adamwg, line 47: Nit: As far as I can tell, this isn't actually a frequency but a timeout for processing an operation. Perhaps `processWorkerItemTimeout`?

adamwg (approved):

PR #352, opened 2020-10-12
Fix firewall cache usage

timoreimann:
The firewall cache used by the firewall controller uses the `go-cmp` library to decide if the cache is up-to-date or not by comparing the target firewall with the current firewall. However, the approach of simply invoking `cmp.Equal()` is insufficient because of several observed API characteristics:

- The order of the inbound and outbound rules as returned from the API is not deterministic.
- The port ranges are specified as `"all"` on create and update requests but are returned as `"0"` on the response.
- Except for `Addresses`, none of the `Sources`/`Destinations` fields are relevant. (This does not seem to be a current problem but could potentially become one going forward.)

This change addresses the described problems by adjusting the comparison functionality accordingly.

We also apply a couple of other improvements that futher optimize our cache usage and reduce the overall complexity of the implementation. Specifically, we:

- Update the cache on successful `Get()` and `Set()`, including when the firewall is missing on the API.
- Do not start `Run()` immediately since the Service event handler will be triggered right form the start.
- Have `Run()` issue an event into the worker queue so that we do not run concurrently to the Service event handler.
  As a consequence, remove the origin label from the metrics since all reconcile requests are now funneled into a single method.
- Significantly simplify `Set()` by removing all reading operations, now that the cache is retrieved right before the method call.
- Add metrics labels to distinguish between skipped and non-skipped reconciles due to cache hits.

Finally, we include a few semi-related refactorings:

- Eliminate the informer synchronization period and rely on the backoff queue solely.
- Pass the `firewallCache` by pointer so that we can embed a non-pointerized `RWMutex` with default initialization.
- Remove the firewall name, firewall tags, and metrics from the `FirewallController` struct and reference them on the `firewallManager` only.
- Move caching and test fixture logic into dedicated files.
- Improve logging.
- Deep-copy test data to rule out unintended, mutation-induced side effects.


cloud-controller-manager/do/firewall_controller.go
```diff
@@ -78,16 +73,10 @@ var (
 	}
 )
 
-// firewallCache stores a cached firewall and mutex to handle concurrent access.
-type firewallCache struct {
```
timoreimann, line 82: Note to reviewers: the cache-related logic was moved into a dedicated file.

cloud-controller-manager/do/firewall_controller_test.go
```diff
@@ -19,221 +19,26 @@ package do
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/informers"
-	"k8s.io/client-go/kubernetes"
+	k8sfake "k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/utils/pointer"
 )
 
-var (
```
timoreimann, line 34: Note to reviewers: the test helpers have been moved into a separate file.

cloud-controller-manager/do/firewall_controller_test.go
```diff
@@ -310,407 +115,178 @@ func TestFirewallController_Get(t *testing.T) {
...
-}
+	eg.Go(loopCreator(func() error {
+		_, err := fc.ensureReconciledFirewall(ctx)
+		if err != nil {
+			return fmt.Errorf("ensureReconciledFirewall failed: %s", err)
+		}
+		return nil
+	}))
 
-func TestFirewallController_isEqualCheckForFirewallRequest(t *testing.T) {
```
timoreimann, line 570: Note to reviewers: the method and test have been superseded by the new improved isEqual / diff functionality located in `firewall_controller_diff{_test}.go`.

cloud-controller-manager/do/firewall_controller_cache_test.go
```diff
@@ -0,0 +1,63 @@
...
+			cache := firewallCache{
+				firewall: test.cachedFirewall,
+			}
+
+			cache.updateCache(test.curFirewall)
+			if !cache.isSet {
+				t.Error("cache is not set")
+			}
+			if cache.firewall != test.curFirewall {
+				t.Errorf("got firewall %s, want %s", printRelevantFirewallParts(cache.firewall), printRelevantFirewallParts(test.curFirewall))
```
gottwald, line 59: Why not re-use the diff output of `compFirewallsEqual`?
Would maybe make the print helper functions obsolete.

cloud-controller-manager/do/firewall_controller_diff.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		return false
+	}, cmp.Ignore())
+
+	diff := cmp.Diff(cf1, cf2, sorterInboundRules, sorterOutboundRules, portRangeMapper, ruleSourceDestFilter, cmpopts.EquateEmpty())
+	return diff == "", diff
+}
+
+func compFirewallFromFirewall(fw *godo.Firewall) *comparableFirewall {
```
gottwald, line 103: This function and the next one deserve a comment of why this auxiliary struct is being built. Alternatively I'd move them up to the struct which has an explanation for that.
An even cleaner solution would be to take those two functions and convert them into transformers for the `cmp.Diff` call.

cloud-controller-manager/do/firewall_fixture_test.go
```diff
@@ -0,0 +1,294 @@
...
+	"sync"
+
+	"github.com/digitalocean/godo"
+	"github.com/mitchellh/copystructure"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+)
+
+var (
+	ctx = context.TODO()
```
gottwald, line 31: AFAIK that's not what `context.TODO` is for. I believe we should use `context.Background` here.

cloud-controller-manager/do/firewall_controller_print.go
```diff
@@ -0,0 +1,96 @@
...
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/digitalocean/godo"
+)
+
+func printRelevantFirewallParts(fw *godo.Firewall) string {
+	if fw == nil {
+		return "<n/a>"
```
MorrisLaw, line 29: nit: do you think we can make `"<n/a>"` a constant, so that the reader could have more context regarding the relevancy of this string. Or a comment as to why it has to be `"<n/a>"`.

cloud-controller-manager/do/firewall_controller_print.go
```diff
@@ -0,0 +1,96 @@
...
+	if inboundRule.PortRange == "" || inboundRule.PortRange == "all" {
+		portRange = "0"
+	}
+	rule := fmt.Sprintf("<Proto:%s PortRange:%s", inboundRule.Protocol, portRange)
+
+	if inboundRule.Sources != nil {
+		rule += fmt.Sprintf(" AddrSources:%s", inboundRule.Sources.Addresses)
+	}
+
+	rule += ">"
```
MorrisLaw, line 70: nit: can we also make this a constant value. I'm assuming that this is the end/closing character for the rule output, but I'm not sure. Maybe we can add a descriptive constant or a comment that could make it more obvious. Wdyt?

cloud-controller-manager/do/firewall_controller_print.go
```diff
@@ -0,0 +1,96 @@
...
+	if outboundRule.PortRange == "" || outboundRule.PortRange == "all" {
+		portRange = "0"
+	}
+	rule := fmt.Sprintf("<Proto:%s PortRange:%s", outboundRule.Protocol, portRange)
+
+	if outboundRule.Destinations != nil {
+		rule += fmt.Sprintf(" AddrDestinations:%s", outboundRule.Destinations.Addresses)
+	}
+
+	rule += ">"
```
MorrisLaw, line 94: same comment as previous

cloud-controller-manager/do/firewall_controller_print.go
```diff
@@ -0,0 +1,96 @@
...
+func printRelevantFirewallParts(fw *godo.Firewall) string {
+	if fw == nil {
+		return "<n/a>"
+	}
+	return printFirewallParts(fw.Name, fw.InboundRules, fw.OutboundRules, fw.Tags)
+}
+
+func printRelevantFirewallRequestParts(fr *godo.FirewallRequest) string {
+	if fr == nil {
+		return "<n/a>"
```
MorrisLaw, line 36: same comment as previous

cloud-controller-manager/do/firewall_controller_diff.go
```diff
@@ -0,0 +1,127 @@
...
+	if fw == nil {
+		return nil
+	}
+
+	return &comparableFirewall{
+		Name:          fw.Name,
+		InboundRules:  fw.InboundRules,
+		OutboundRules: fw.OutboundRules,
+		Tags:          fw.Tags,
```
MorrisLaw, line 112: Is it possible for `fm.Tags` to return a nil slice and if so, could this be problematic if we don't check?

cloud-controller-manager/do/firewall_controller_diff.go
```diff
@@ -0,0 +1,127 @@
...
+func compFirewallFromFirewallRequest(fr *godo.FirewallRequest) *comparableFirewall {
+	if fr == nil {
+		return nil
+	}
+
+	return &comparableFirewall{
+		Name:          fr.Name,
+		InboundRules:  fr.InboundRules,
+		OutboundRules: fr.OutboundRules,
+		Tags:          fr.Tags,
```
MorrisLaw, line 125: same question regarding a nil slice

MorrisLaw (commented):
I like this refactoring! Left some comments

cloud-controller-manager/do/firewall_controller_cache.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/digitalocean/godo"
+	"k8s.io/klog/v2"
+)
+
+// firewallCache stores a cached firewall and mutex to handle concurrent access.
+type firewallCache struct {
+	sync.RWMutex
+	firewall *godo.Firewall
+	isSet    bool
```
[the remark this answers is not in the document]
timoreimann: Unfortunately not because the firewall we periodically try to fetch from the API could be absent, and we still need to be able to signal to cache users that the cache has been refreshed. If we relied on `firewall` being nil only, we could run into a scenario where reach out to the API on every Service reconcile, leading to a fair number of extra requests. For instance, if we keep failing to create the firewall continuously, the cache would always be considered empty (because `firewall` never becomes non-nil), so we'd try to fetch it again and again until the firewall can eventually be created.

Admittedly it's a bit of an edge case (though I ran into it during tests when the implementation had a subtle bug). I think it's still good to minimize the number of requests we do.

I'll push a small update to comment the field.

cloud-controller-manager/do/firewall_controller_cache_test.go
timoreimann: Will include the diff output from `firewallsEqual` since I personally found it useful to see both the print helper representation and the diff during testing.

cloud-controller-manager/do/firewall_controller_diff.go
timoreimann: Not 100% sure about what would all need to change to make the two functions produce transformers. I'll move them further up for now, but curious to hear details about the transformer suggestion (either here or via DM).

cloud-controller-manager/do/firewall_fixture_test.go
timoreimann: Yup you're right. Will change.

cloud-controller-manager/do/firewall_controller_diff.go
```diff
@@ -0,0 +1,127 @@
...
+	// value for comparing.
+	portRangeMapper := cmp.FilterPath(func(p cmp.Path) bool {
+		switch p.String() {
+		case "InboundRules.PortRange", "OutboundRules.PortRange":
+			return true
+		}
+		return false
+	}, cmp.Comparer(func(pr1, pr2 string) bool {
+		sanitizePortRange := func(pr string) string {
+			if pr == "" || pr == "0" || pr == "all" {
```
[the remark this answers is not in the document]
timoreimann: You're completely right; I'll remove it.

cloud-controller-manager/do/firewall_controller_diff.go
timoreimann: With regards to accessing the value, it should not cause issues since we always read the value in a way that's unproblematic (e.g., use the `%s` placeholder with `fmt.Printf`).

With regards to the result of our comparison logic, it should also not cause issues since we pass `cmpopts.EquateEmpty()` to `cmp.Diff`, making sure that nil slices are considered equal to empty ones.

cloud-controller-manager/do/firewall_fixture_test.go
```diff
@@ -0,0 +1,294 @@
...
+
+	kclient kubernetes.Interface
+
+	inf = informers.NewSharedInformerFactory(kclient, 0)
```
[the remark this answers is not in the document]
timoreimann: It's used by two tests: `TestFirewallController_ConcurrentUsage` and `TestFirewallController_Set`.

timoreimann:
@gottwald @MorrisLaw thanks for the feedback, I addressed your comments.

cloud-controller-manager/do/firewall_controller_cache.go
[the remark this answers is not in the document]
gottwald: Gotcha, didn't see that case.

cloud-controller-manager/do/firewall_controller_cache_test.go
gottwald: SGTM

cloud-controller-manager/do/firewall_controller_print.go
```diff
@@ -24,7 +24,11 @@ import (
 	"github.com/digitalocean/godo"
 )
 
-const missingFirewallLabel = "<n/a>"
+const (
```
MorrisLaw, line 27: nice 👍

MorrisLaw (approved):
Thanks for the updates 👍

gottwald (approved):
LGTM

PR #354, opened 2020-10-21
Use vendored dependencies for compile Makefile target

timoreimann:


Makefile
```diff
@@ -72,7 +72,7 @@ compile:
 	@echo "==> Building the project"
 	@docker run -v $(PWD):/go/src/github.com/digitalocean/digitalocean-cloud-controller-manager \
 	  -w /go/src/github.com/digitalocean/digitalocean-cloud-controller-manager \
-	  -e GOOS=linux -e GOARCH=amd64 -e CGO_ENABLED=0 -e GOFLAGS=-mod=vendor golang:$(GO_VERSION) \
+	  -e GOOS=linux -e GOARCH=amd64 -e CGO_ENABLED=0 -e GO111MODULE=on -e GOFLAGS=-mod=vendor golang:$(GO_VERSION) \
```
gottwald, line 75: From the go module reference:

>If GO111MODULE=auto or is unset, the go command runs in module-aware mode if a go.mod file is present in the current directory or any parent directory (the default behavior).

So we shouldn't need it. Also true for the hunk below.
Rest LGTM

gottwald (approved):

Makefile
timoreimann: Removed here. As discussed via DM, we will keep it for the `test` target since we cannot guarantee for sure that users don't set `GO111MODULE=off` (which no one _should_ do anymore at least for this repo, but you never know).

PR #355, opened 2020-10-21
Target proper tags when updating Kubernetes dependencies

timoreimann:
The latest Kubernetes dependencies moved from using the kubernetes-X.Y.Z tag to vX.Y.Z, which our upgrade script now accounts for.

We also remove the Makefile variable indicating which Kubernetes version we are using since that information can now be easily retrieved from the `go.mod` file.

A follow-up PR will update the Kubernetes dependencies to 1.19.3.


gottwald (approved):
LGTM

PR #357, opened 2020-10-21
Improve firewall metrics design

timoreimann:
This change updates our firewall-related metrics design in the following aspects:

- Rename `api_request_duration_seconds` to `api_operation_duration_seconds` and capture API operations (get by ID, get by list, create, update) instead of HTTP requests. This allows us to distinguish between requests using the same method (namely GETs by ID and by list) while also making it easier for the user to understand what these metrics are about (telling what a PUT or POST means is less clear than a "create" or "update").
- Rename `run_loop_duration_seconds` to `resource_sync_duration_seconds` to (presumably) ease understanding: the former name did not easily imply what kind of functionality happened during a run loop.
- Add a "result" label to be two before-mentioned metrics to easily parse for succeeded and failed operations in Prometheus since we may not always have an HTTP code available (e.g., when a network error occurred).
- Make metrics help descriptions more concise and align with Prometheus convention in terms of wording/phrasing.
- Add counter metrics for existing histograms to allow inexpensive, basic observability.

We also fix `filterFirewallList` by returning a response so that we can fill in the HTTP code Prometheus label.

Finally, we DRY out the existing observing functionality into a set of helper methods.

### DIFF
--- a/cloud-controller-manager/do/firewall_fixture_test.go
+++ b/cloud-controller-manager/do/firewall_fixture_test.go
@@ -273,12 +273,7 @@ func newFakeFirewallManager(client *godo.Client, cache *firewallCache) *firewall
 		fwCache:            cache,
 		workerFirewallName: testWorkerFWName,
 		workerFirewallTags: testWorkerFWTags,
-		metrics: metrics{
-			host:               "localhost:8080",
-			apiRequestDuration: apiRequestDuration,
-			runLoopDuration:    runLoopDuration,
-			reconcileDuration:  reconcileDuration,
-		},
+		metrics:            newMetrics("localhost:8080"),
 	}
 }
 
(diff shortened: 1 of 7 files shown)


gottwald (approved):
LGTM

cloud-controller-manager/do/firewall_fixture_test.go
```diff
@@ -273,12 +273,7 @@ func newFakeFirewallManager(client *godo.Client, cache *firewallCache) *firewall
 		fwCache:            cache,
 		workerFirewallName: testWorkerFWName,
 		workerFirewallTags: testWorkerFWTags,
-		metrics: metrics{
-			host:               "localhost:8080",
-			apiRequestDuration: apiRequestDuration,
-			runLoopDuration:    runLoopDuration,
-			reconcileDuration:  reconcileDuration,
-		},
+		metrics:            newMetrics("localhost:8080"),
```
MorrisLaw, line 276: I like this 👍

MorrisLaw (approved):
Nice refactoring Timo! :shipit:

PR #362, opened 2020-11-12
Fix broken firewall counter metrics by incrementing

timoreimann:
The counter metrics would never increase because we forgot to call the `Inc()` method.


gottwald (approved):
LGTM

PR #365, opened 2020-12-01
Include HTTPS in LB health check protocol docs

timoreimann:
HTTPS as health check protocol for LBs has been supported in quite a while but was not documented.


MorrisLaw (approved):
:ship:

gottwald (approved):

PR #366, opened 2020-12-08
Remove beta note from LB size annotation

timoreimann:
As announced today: https://www.digitalocean.com/blog/introducing-new-digitalocean-load-balancers-plans/


anitgandhi (approved):
Thanks!

PR #369, opened 2021-01-15
vendor: Bump kubernetes dependencies to 1.20.2

adamwg:
As part of this change, we have to update our main to use k8s.io/cloud-provider/app, since the old version of it has now been removed from the Kubernetes tree.


cloud-controller-manager/cmd/digitalocean-cloud-controller-manager/main.go
```diff
@@ -23,14 +23,21 @@ import (
...
 
 	"github.com/digitalocean/digitalocean-cloud-controller-manager/cloud-controller-manager/do"
 	"github.com/spf13/pflag"
 )
 
 func main() {
-	command := app.NewCloudControllerManagerCommand()
+	opts, err := options.NewCloudControllerManagerOptions()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to construct options: %v\n", err)
```
timoreimann, line 37: Is it correct that we emit an error message but do not terminate the program?

timoreimann (commented):
One quick question, and maybe the ask to append the change log briefly. Otherwise all good. 👍 

Glad to see we can finally stop vendoring k/k. 🎉

timoreimann (approved):
(LGTM'ing to save one round trip.)

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager/main.go
adamwg: Oops, good catch. Fixed.

timoreimann (approved):

adamwg:
This doesn't quite work properly due to https://github.com/kubernetes/cloud-provider/issues/44. Let's revisit this once the upstream situation is clearer.

PR #372, opened 2021-02-08
Do not forget work item on firewall controller error

timoreimann:
This is how controllers are supposed to be used.

See also the [canonical sample controller implementation](https://github.com/kubernetes/sample-controller/blob/bc98db7fe443244690abef147d4da7bdff956476/controller.go#L185-L238).


gottwald (approved):
LGTM

PR #376, opened 2021-03-28
Document label to exclude specific nodes from the LB target set

### LINKED ISSUE #375: Expose LoadBalancer only on worker nodes
Is there a way to tell the controller to load balance services only on worker nodes?

timoreimann:
This is an upstream label implemented in the CCM core part, so we automatically support it.

Fixes #375

/cc @braghettos

### DIFF
--- a/docs/controllers/services/examples/README.md
+++ b/docs/controllers/services/examples/README.md
@@ -253,7 +253,7 @@ You're done!
 
 ## Excluding specific nodes as targets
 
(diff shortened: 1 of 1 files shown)


braghettos:
@timoreimann I think that you should write the label in the doc with the '=true' value.

timoreimann:
@braghettos great call. The value is actually not relevant, it suffices to have the key set. I updated the docs to make this more clear.

gottwald (approved):
LGTM

PR #379, opened 2021-06-22
Update Kubernetes dependencies to v1.21.2

adamwg:
With this change, we need to switch to using `k8s.io/cloud-provider/app` instead of `k8s.io/kubernetes/cmd/cloud-controller-manager/app`, which introduces some changes in our setup code. This also means we no longer need to vendor k/k.


adamwg:
Currently testing this on 1.19, 1.20, and 1.21. CCM starts up correctly (after a bit of configuration finessing), so I don't anticipate any problems, but I'll leave the WIP label until those tests pass.

adamwg:
Tests passed, this is ready for review.

scripts/update-k8s.sh
```diff
@@ -38,8 +38,6 @@ while read -ra LINE; do
 done < <(curl -fsSL "https://raw.githubusercontent.com/kubernetes/kubernetes/$KUBERNETES_VERSION/go.mod" \
   | grep -E '^\s*k8s.io/\S+ v\S+$')
 
-deps+=("-replace k8s.io/kubernetes=k8s.io/kubernetes@$KUBERNETES_VERSION")
```
timoreimann, line 41: Did you still use the (updated) script to bump the dependencies? I'm curious to know if we can just do `go get ...` directly and rely on Go modules / proper semver in Kubernetes repositories other than k/k to resolve the (transitive) dependencies correctly.

If that works, then I'd think it's preferable because we ideally shouldn't set `-replace` directives explicitly (as we still do in line 37); in that case, we could leave the script as-is in case we do need to revendor k/k again.

If the modified script is still needed and/or makes things easier, then let's keep it in the updated form by all means.

timoreimann (approved):
One question around the usage/value of the `update-k8s.sh` script. I assume it's still needed/useful, so approving already for that case.

MorrisLaw (approved):
this lgtm

MorrisLaw (approved):
this lgtm

scripts/update-k8s.sh
adamwg: I did use the script, but removing the `replace`s works fine, and without those there's really no need for the script. Removed the replaces and the script in additional commits.

adamwg: I also pondered writing a new script that would `go get` the right versions of things, but since `k8s.io/klog` and `k8s.io/utils` don't follow the same versioning pattern that's a little dicey. If I do this update again in the future I would just edit the go.mod file by hand and then `go mod tidy; go mod vendor`.

timoreimann:
Still LGTM.

adamwg:
Travis has abandoned us. `make ci` passes locally; added a changelog update and I'll merge now.

PR #382, opened 2021-08-24
replace travis with github actions

varshavaradarajan:
Since Travis is no longer free, we're going with GH actions. Additionally, this PR adds a release workflow to automate releasing a new ccm version.


.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
+      - name: run make bump-version
+        run: make bump-version
+        env:
+          NEW_VERSION: ${{ github.event.inputs.tag }}
+      - name: commit manifests
+        uses: EndBug/add-and-commit@a988073222b8bd50f3ecfca9c0ab7dfbf0d08ceb
+        with:
+          message: "create release manifests for ${{ github.event.inputs.tag }}"
+          add: "releases/"
```
timoreimann, line 26: The `bump-version` Makefile target [modifies a lot of files across the repo](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/a843046121562eec208cc12c9637ada7894e4ef7/Makefile#L51-L58) in an attempt to increase all mentioned "latest version" specifiers. This was always pretty gross IMO, and has occasionally targeted files that shouldn't have been updated. If we automate the commit, then I think we should first have a PR that removes these latest version specifiers so that we only commit the `releases/` directory and the `VERSION` file.

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # target commit should be the commit with the new release manifest, not the commit when workflow was triggered
+      - name: find target commit to tag
+        run: |
+          sha=$(git rev-parse HEAD)
+          echo "TARGET_SHA=$sha" >> $GITHUB_ENV
+      - name: Create GitHub Release
+        uses: actions/create-release@1e92f6fc665e271a5435631ba00711fdd83d9d27
```
timoreimann, line 35: Is this the same Action referenced by https://github.com/actions/create-release? That repo is archived, is that a reason to be concerned?

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
+name: ci
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
timoreimann, line 6: This means we'd only trigger the CI on PRs against the master branch, correct?

If so, could we relax this restriction so that we can "chain" PRs if needed and still run CI?

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+        env:
+          NEW_VERSION: ${{ github.event.inputs.tag }}
+      - name: commit manifests
+        uses: EndBug/add-and-commit@a988073222b8bd50f3ecfca9c0ab7dfbf0d08ceb
+        with:
+          message: "create release manifests for ${{ github.event.inputs.tag }}"
+          add: "releases/"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # target commit should be the commit with the new release manifest, not the commit when workflow was triggered
```
timoreimann, line 29: I'm not sure I understand this one. 🤔  What's the "new release manifest"?

README.md
```diff
@@ -173,7 +171,20 @@ This will create a binary with version `dev` and docker image pushed to
...
+
+The workflow does the following:
+
+Runs make bump-version with the specified tag
+Creates the ccm related manifests file as <tag>.yaml
+Commits the manifest file under releases/ directory in the repo
+Creates release and tags the new commit with the input tag specified when workflow is triggered
+Login with dockerhub credentials specified as secrets
+Builds the docker image digitalocean/digitalocean-cloud-controller-manager:<tag>
+Pushes digitalocean/digitalocean-cloud-controller-manager:<tag> to dockerhub
```
timoreimann, line 185: nit: move that above _"Builds the docker image [...]"_ since we first push the tag and then build/push the container image.

timoreimann, line 185:
```suggestion
Pushes digitalocean/digitalocean-cloud-controller-manager:<tag> to git
```

timoreimann (commented):
I like the degree of automation we will get to once this PR is merged. 👍 

Left a few comments, suggestions, and questions mostly with regards to the release part. We could also split out the CI part into a separate PR so that we can merge that earlier and work on the release Action separately. Just an idea though, we can also work on both features in this PR.

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
+name: release
+on:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
varshavaradarajan: Screen Shot 2021-08-25 at 9 31 32 AM

Allows you to select branch. But, I can only test restricting the branch after merging the release workflow. It doesn't show up on the Actions tab right now.

.github/workflows/release.yml
varshavaradarajan: bump version alters the version in these files:
- VERSION
- README.md
- docs/getting-started.md
- docs/example-manifests/cloud-controller-manager.yml
- releases/${NEW_VERSION}.yml

These are valid changes IMO and we should commit them. I'll modify this task to add them as well.

.github/workflows/release.yml
varshavaradarajan: Ah, okay. Didn't verify that. Will switch to https://github.com/softprops/action-gh-release

README.md
varshavaradarajan: This has 
Creates release and tags the new commit with the input tag
and builds docker image digitalocean/digitalocean-cloud-controller-manager:<tag> to dockerhub in that order. So, I don't understand what you mean..

.github/workflows/release.yml
varshavaradarajan: Let's say that XXX is the HEAD on master when the workflow was triggered. After bumping the versions, we now have YYY as the HEAD commit. We need to tag the new version at YYY. So, we need to find that target commit sha to pass to the release action (for both softprops/action-gh-release and actions/create-release)

.github/workflows/release.yml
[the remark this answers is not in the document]
timoreimann: Ah okay makes sense. We can refine once this is merged. 👍

.github/workflows/release.yml
timoreimann: What I was trying to say is that I don't like how we have been doing these mass replacements across several documentation files just to have the latest released version mentioned at several convenient places. This was somewhat acceptable when we did the release manually, but we shouldn't let automation do this. As I wrote, there have been situations in the past where those replacements changed portions of the repo that shouldn't have been changed, it was only due to the interactive release process that we could undo them before they were pushed to the git server.

Customers/users can figure out what the latest version is by looking at the releases page, so my strong preference would be to do a quick PR that removes the latest references from the docs before merging this one.

.github/workflows/release.yml
timoreimann: Thanks, the explanation made it clear for me now. 👍

.github/workflows/release.yml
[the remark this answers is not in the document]
timoreimann: That's a fair compromise, agreed.

.github/workflows/release.yml
varshavaradarajan: Only added releases/ and VERSION file here. Will follow up with the bump versions change in the next PR

PR #385, opened 2021-09-07
update gh release workflow version

varshavaradarajan:
The new version of gh release workflow supports tag_name and tag_committish which we use in the release workflow


MorrisLaw (approved):
:rocket:

PR #388, opened 2021-10-15
Support disabling automatic DNS records for load balancers.

andrewsomething:
The DO API now [supports disabling automatic DNS record creation](https://docs.digitalocean.com/release-notes/#october-12) for load balancers using Let's Encrypt certificates. This PR extends that support to digitalocean-cloud-controller-manager.


dikshant:
I have updated #390 to align with the same behavior as that one. We wanted to get that PR out the door as soon as possible on Monday so I'd like to move forward with #390 and address any feedback there. Thanks for doing this @andrewsomething I wasn't familiar with the workings of DOKS before and missed it.

PR #390, opened 2021-10-16
cloud-controller-manager/do/load-balancers: add new annotation for disabling automatic DNS record creation for let's encrypt cert

dikshant:
By default when adding a cert to a load balancer it automatically creates a A record pointing to the root domain in the cert. Setting this new annotation of `"service.beta.kubernetes.io/do-loadbalancer-disable-lets-encrypt-dns-records"` to `"true"` will disable this behavior and not automatically create the A record.


timoreimann:
Just realized that #388 by @andrewsomething  adds the same functionality, just with slightly different semantics around when/if a `nil` value is passed to the API. It's probably a good idea to sync up on what approach we want to follow.

PR #399, opened 2022-01-10
bump k8s dependencies to 1.22.5

CollinShoop:
CON-4801

```
GO111MODULE=on go get k8s.io/client-go@v0.22.5
GO111MODULE=on go get k8s.io/cloud-provider@v0.22.5
GO111MODULE=on go mod tidy
GO111MODULE=on go mod vendor
```

todo
- [ ] compatibility testing?

follow-up
- [ ] https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/400

[1.22.x csi update](https://github.com/digitalocean/csi-digitalocean/pull/421), dependencies ended up used the old `update-k8s` script and ended up pulling `v0.22.6-rc.0`. Just want to confirm which version is best to use


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v0.1.35 (beta) - Jan 10 2022
```
timoreimann, line 3: I think this needs to be v0.1.36 since v0.1.35 was released already.

IMHO we should defer to amend the change log until we submit the "release PR" like #389 (which happened to miss the change log update). That way, the release entry does not show up right before we are actually making the release, which ensures that the date is correct.

timoreimann (commented):
LGTM with one small comment.

CHANGELOG.md
CollinShoop: Ah, I see. I was wondering about that. I'll remove it from this PR and write a note to follow up.

PR #408, opened 2022-02-09
ensure only one droplet with given name exists

varshavaradarajan:
This PR ensures that we throw an error if more than one droplet with the same name exists in the user's account. In doing so, we ensure that  node.spec.ProviderID is always set to the right droplet id or is empty. The impact is that this will cause an error when creating or updating load balancers resulting in the `provider id cannot be empty` error in the ccm log and records a warning on the service object. But, we would have logged the non-unique droplets in the ccm log making it easier to identify cases and fix it. In the long run, we plan to fix this automatically in our backend and prevent droplets with the same name coming up as far as possible.

### DIFF
--- a/cloud-controller-manager/do/droplets.go
+++ b/cloud-controller-manager/do/droplets.go
@@ -186,24 +186,51 @@ func dropletByID(ctx context.Context, client *godo.Client, id int) (*godo.Drople
 // considered.
 func dropletByName(ctx context.Context, client *godo.Client, nodeName types.NodeName) (*godo.Droplet, error) {
 	// TODO (andrewsykim): list by tag once a tagging format is determined
+	needed := []*godo.Droplet{}
 	droplets, err := allDropletList(ctx, client)
 	if err != nil {
 		return nil, err
 	}
 
 	for _, droplet := range droplets {
+		droplet := droplet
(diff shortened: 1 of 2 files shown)


varshavaradarajan:
@timoreimann - if the impact (i.e unable to create or update lb) is a huge concern, then we can merge this after we fix the non unique droplet name issue internally

CollinShoop (approved):

timoreimann (commented):
I'm inclined to propose we shouldn't have this for two reasons:

1. `dropletByName` is only invoked when the provider/droplet ID is missing. That's never the case in DOKS where we configure the kubelet with the droplet ID, and that's also what many other systems like Cluster API / CAPDO do. The reason is that lookup by name is very inefficient for bigger clusters/accounts.
2. There's still an edge case where an equally named node could slip in while we iterate over the paged results, so we can't rule out we won't run into the problem again.

IMHO the problem needs to be resolved at the backend/client side, i.e., whoever creates the droplets.

mo-rieger:
Should we close this PR? @varshavaradarajan @timoreimann

timoreimann:
Yes, I agree we should close it. Thanks for raising @mo-rieger.

PR #409, opened 2022-02-11
ccm/do/load-balancers: add udp protocol support

dikshant:
This adds UDP protocol to the load balancers through CCM. We also handle the case where we use the same port between {http2, http, https, tcp} and udp.


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -793,6 +805,8 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
 		return nil, err
 	}
 
+	// we don't pass UDP ports here since we can allow duplicate
```
dikshant, line 808: Ugh, while LB supports listening on the same port for both udp and tcp looks like Kubernetes does not:
https://github.com/kubernetes/kubernetes/issues/39188
https://github.com/kubernetes/kubernetes/issues/2995#issuecomment-267024643

I guess I'll rework this to be consistent with the other ports to avoid confusion for users.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -85,6 +85,11 @@ const (
 	// value must be between 2 and 10. Defaults to 5.
 	annDOHealthCheckHealthyThreshold = "service.beta.kubernetes.io/do-loadbalancer-healthcheck-healthy-threshold"
 
+	// annDOUDPPorts is the annotation used to specify which ports of the load balancer
+	// should use the UDP protocol. This is a comma separated list of ports
+	// (e.g., 53,67).
+	annDOUDPPorts = "service.beta.kubernetes.io/do-loadbalancer-udp-ports"
```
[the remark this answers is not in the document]
dikshant: Ahhh I see. In that case yeah saving service ports will be enough. I will remove this annotation.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -823,14 +841,23 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
...
 		// are found.
+		if udpPortMap[port.Port] {
+			protocol = protocolUDP
+		}
+
+		// we have to delete the ports after we add them to prevent
+		// them from overriding a port that is found in the udp port map
 		if httpPortMap[port.Port] {
 			protocol = protocolHTTP
+			delete(httpPortMap, port.Port)
```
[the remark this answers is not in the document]
dikshant: > Assuming we want to allow same-port-different-protocol usage: do we know if Kubernetes supports that cleanly? Looking the issue you linked and other issues linked from that, I didn't get a super clear picture on that second point.
Sorry about the confusion on this. I had initially thought that Kubernetes didn't support having the same port for two protocols {TCP, UDP}. But I was a bit wrong about this. The discussion for the actual issue is here:
https://github.com/kubernetes/kubernetes/issues/39188
It seems the TLDR; is that as long as the ports are named, setting the service up for the first time shouldn't be an issue. But it appears due to an issue on K8s side, if the ports are updated, it can end up updating the wrong port/protocol pair.

> And do we (want to) allow reusing the same port for TCP and UDP protocols at the same time from a DO LB perspective? Removing the port here makes me think we don't but a comment further above explains why we don't consider UDP ports in duplicate detection for reuse purposes, so I'm somewhat confused. (Chances are I'm misunderstanding something.)
On the LB side we have support for allowing the same port for TCP and UDP. The deleting ports from the other maps was a way of achieving this behavior. I ran into a problem in the tests where the protocols would end up getting overridden. Since we have this pattern, where if we find a particular port in a particular map (where each map has ports for a particular protocol) we determine that port to be of that protocol. If I don't delete a port once it's found from the maps excluding UDP and that port is found again we end up overwriting its protocol. 
This results in test failure like this:
```
FAIL: Test_buildForwardingRules (0.00s)
    --- FAIL: Test_buildForwardingRules/UDP_and_HTTP2_ports_shared (0.00s)
        loadbalancers_test.go:2203: unexpected forwarding rules
        loadbalancers_test.go:2204: expected: [godo.ForwardingRule{EntryProtocol:"http2", EntryPort:4443, TargetProtocol:"http2", TargetPort:10443, CertificateID:"", TlsPassthrough:true} godo.ForwardingRule{EntryProtocol:"http2", EntryPort:443, TargetProtocol:"http2", TargetPort:10444, CertificateID:"", TlsPassthrough:true} godo.ForwardingRule{EntryProtocol:"udp", EntryPort:4443, TargetProtocol:"udp", TargetPort:10445, CertificateID:"", TlsPassthrough:false} godo.ForwardingRule{EntryProtocol:"udp", EntryPort:443, TargetProtocol:"udp", TargetPort:10446, CertificateID:"", TlsPassthrough:false}]
        loadbalancers_test.go:2205: actual: [godo.ForwardingRule{EntryProtocol:"http2", EntryPort:4443, TargetProtocol:"http2", TargetPort:10443, CertificateID:"", TlsPassthrough:true} godo.ForwardingRule{EntryProtocol:"http2", EntryPort:443, TargetProtocol:"http2", TargetPort:10444, CertificateID:"", TlsPassthrough:true} godo.ForwardingRule{EntryProtocol:"http2", EntryPort:4443, TargetProtocol:"http2", TargetPort:10445, CertificateID:"", TlsPassthrough:true} godo.ForwardingRule{EntryProtocol:"http2", EntryPort:443, TargetProtocol:"http2", TargetPort:10446, CertificateID:"", TlsPassthrough:true}]
```
 I'm happy to change this to a better way if there is one, but that deleting from the map trick is what I came up with.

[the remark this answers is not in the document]
dikshant: Looks like removing `annDOUDPPorts` like suggessted below would get rid of this change, but first I'll have to rework a large portion of the PR which I am doing now

cloud-controller-manager/do/loadbalancers.go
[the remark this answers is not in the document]
dikshant: Fixed!

cloud-controller-manager/do/loadbalancers.go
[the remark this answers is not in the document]
dikshant: Okay since the annotations are removed, we are just using the Protocol from the service definition and there is no need for this terrible map deletion logic. Thanks @timoreimann!

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -923,7 +927,7 @@ func getProtocol(service *v1.Service) (string, error) {
 	}
 
 	switch protocol {
-	case protocolTCP, protocolHTTP, protocolHTTPS, protocolHTTP2:
+	case protocolTCP, protocolUDP, protocolHTTP, protocolHTTPS, protocolHTTP2:
```
timoreimann, line 930: `getProtocol` returns the default protocol that is to be used when no explicit protocol was specified for a particular port. (The function would probably be better named `getDefaultProtocol`.) This is needed because we support TCP and various HTTP* flavors.

I'm thinking that we do not need to support this for UDP since selection of the UDP protocol is determined by the protocol value on the Service port definition (i.e., customers need to specify it explicitly anyway). WDYT?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -846,8 +846,12 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
 func buildForwardingRule(service *v1.Service, port *v1.ServicePort, protocol, certificateID string, tlsPassThrough bool) (*godo.ForwardingRule, error) {
 	var forwardingRule godo.ForwardingRule
 
-	if port.Protocol != portProtocolTCP {
-		return nil, fmt.Errorf("only TCP protocol is supported, got: %q", port.Protocol)
+	switch port.Protocol {
+	case portProtocolTCP:
+	case portProtocolUDP:
+		protocol = protocolUDP
```
timoreimann, line 852: I wonder if we should do this assignment in the calling `buildForwardingRules()` function where we also assign the other protocol values. This would keep the logic that sets the protocol in one place.

It's a small bit but could make the code slightly easier to understand.

timoreimann (commented):
Could we also extend the [findDups()](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/0681058bc53f86ed6ef1a2cfcabee927f6847bce/cloud-controller-manager/do/loadbalancers.go#L1251) function to detect if a UDP port was accidentally specified on a HTTP*-based annotation? Users should never do this, and our validation logic can help identify this invalid configuration case early on.

docs/controllers/services/annotations.md
```diff
@@ -17,7 +17,7 @@ If no custom name is specified, a default name is chosen consisting of the chara
 
 ## service.beta.kubernetes.io/do-loadbalancer-protocol
 
-The default protocol for DigitalOcean Load Balancers. Options are `tcp`, `http`, `https`, and `http2`. Defaults to `tcp`.
+The default protocol for DigitalOcean Load Balancers. Options are `tcp`, `udp`, `http`, `https`, and `http2`. Defaults to `tcp`. Note that if UDP is supported a healtcheck configuration configured against a TCP/HTTP/HTTPS service is still required. In order to use UDP protocol with a Load Balancer, please reach out to support with a ticket to enable it.
```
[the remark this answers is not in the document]
dikshant: Fixed.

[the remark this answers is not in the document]
dikshant: > With a TCP health check port now required when UDP is used
It doesn't have to be TCP healthcheck. We support HTTP/HTTPS for the healthcheck with a path as well. We just don't support UDP as a healthcheck protocol.

cloud-controller-manager/do/loadbalancers.go
dikshant: Moved it!

docs/controllers/services/annotations.md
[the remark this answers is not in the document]
timoreimann: Sorry, I meant TCP as in the Kubernetes Service protocol, not as in the protocols that DO LBs support.

We currently pick the first Service (node) port for health checking, so my suggestion is to pick the first where the Service protocol is TCP. Otherwise, if no health check port is defined explicitly, I think our logic could theoretically pick a port from a UDP Service which I suppose would lead to an API error.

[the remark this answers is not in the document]
dikshant: Ah I see. I'd add some checks

timoreimann:
@dikshant might have gotten lost in the previous review round -- did you see my comment/suggestion re: extending `findDups()` at https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/409#pullrequestreview-930000932?

dikshant:
> Could we also extend the [findDups()](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/0681058bc53f86ed6ef1a2cfcabee927f6847bce/cloud-controller-manager/do/loadbalancers.go#L1251) function to detect if a UDP port was accidentally specified on a HTTP*-based annotation? Users should never do this, and our validation logic can help identify this invalid configuration case early on.

Missed this earlier but just saw it. Not sure how we could extend that to findDups 🤔. I assume you mean we would want to error out in configs like these?
```
                                {
					Name: "test",
					UID:  "abc123",
					Annotations: map[string]string{
						annDOHTTPPorts: "8888",
					},
				},
				Spec: v1.ServiceSpec{
					Ports: []v1.ServicePort{
						{
							Name:     "test-udp",
							Protocol: "UDP",
							Port:     int32(8888),
							NodePort: int32(18443),
						},
					},
				},
```

I am trying to find a way to catch this currently.

docs/controllers/services/annotations.md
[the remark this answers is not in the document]
dikshant: Fixed this!

timoreimann:
Re: dups: as clarified via DM, I had misunderstood that we wouldn't support reusing the same port for TCP and UDP Service specs, which is what I thought we should try to identify. Turns out that specifying TCP/HTTP* ports on annotations without a corresponding Service port is something we don't check for today, which isn't great but also not something this PR needs to address.

All good then, and sorry about the confusion.

timoreimann (approved):
Thanks @dikshant. 💙

cpanato (approved):
looks cool!

cpanato:
defer to @timoreimann to merge

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -833,6 +844,13 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
 			protocol = protocolHTTP2
 		}
 
+		if port.Protocol == v1.ProtocolUDP {
+			if tcpPortMap[port.Port] {
+				return nil, errors.New("cannot share a port between TCP and UDP")
```
timoreimann, line 849: Can we include the shared port in the returned error message?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -950,14 +970,22 @@ func healthCheckPort(service *v1.Service) (int, error) {
 
 	if len(ports) == 1 {
 		for _, servicePort := range service.Spec.Ports {
-			if int(servicePort.Port) == ports[0] {
+			if int(servicePort.Port) == ports[0] && servicePort.Protocol != v1.ProtocolUDP {
```
timoreimann, line 973: If the protocol is UDP, then I think we'd fall through to the error returned in line 977 saying that the specified health check port was not found. This could make it harder than necessary for a user to debug what they did wrong.

I'd suggest we return a more specific, dedicated error when the protocol is UDP.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -950,14 +970,22 @@ func healthCheckPort(service *v1.Service) (int, error) {
...
-	return int(service.Spec.Ports[0].NodePort), nil
+	// return the first TCP port found as the healthcheck port
+	for _, p := range service.Spec.Ports {
+		if p.Protocol == v1.ProtocolTCP {
+			return int(p.NodePort), nil
+		}
+	}
+
+	// Otherwise no TCP ports were found
+	return 0, errors.New("healthcheck port has to be of type protocol TCP, HTTP or HTTPS")
```
timoreimann, line 988: I think HTTP* is not relevant in this context, only whether the Service port is using the TCP protocol.

What do you think about simplifying the message to something like

```
no healthcheck port of type TCP found
```

?

docs/controllers/services/annotations.md
```diff
@@ -17,7 +17,7 @@ If no custom name is specified, a default name is chosen consisting of the chara
 
 ## service.beta.kubernetes.io/do-loadbalancer-protocol
 
-The default protocol for DigitalOcean Load Balancers. Options are `tcp`, `http`, `https`, and `http2`. Defaults to `tcp`.
+The default protocol for DigitalOcean Load Balancers. Options are `tcp`, `udp`, `http`, `https`, and `http2`. Defaults to `tcp`. Note that if UDP is supported a health check configuration configured against a TCP/HTTP/HTTPS service is still required. In order to use UDP protocol with a Load Balancer, please reach out to support with a ticket to enable it. Note: currently, a port cannot be shared between TCP and UDP due to a [bug in Kubernetes](https://github.com/kubernetes/kubernetes/issues/39188).
```
timoreimann, line 20: I think we've moved away from supporting/expecting the `udp` string value in the `service.beta.kubernetes.io/do-loadbalancer-protocol` annotation, meaning we should drop it here.

timoreimann (commented):
This looks good. Left a last round of suggestions, mostly around improving our error reporting.

I remembered that we also [maintain a number of examples](https://github.com/digitalocean/digitalocean-cloud-controller-manager/tree/2209522dcf6429bddce77ea5cacb7ae6c28d15b8/docs/controllers/services/examples). Would you mind adding one (or more) for UDP as well?

cloud-controller-manager/do/loadbalancers.go
dikshant: Added!

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -793,7 +795,18 @@ func buildForwardingRules(service *v1.Service) ([]godo.ForwardingRule, error) {
...
+	for _, port := range service.Spec.Ports {
+		switch port.Protocol {
+		case v1.ProtocolTCP:
+			tcpPortMap[port.Port] = true
+		case v1.ProtocolUDP:
+			udpPorts = append(udpPorts, int(port.Port))
+		}
+	}
+
+	portDups := findDups(httpPorts, httpsPorts, http2Ports, udpPorts)
```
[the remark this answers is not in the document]
dikshant: Fixed!

cloud-controller-manager/do/loadbalancers.go
dikshant: Added!

cloud-controller-manager/do/loadbalancers.go
dikshant: Changed!

docs/controllers/services/annotations.md
dikshant: I removed udp references from here. Left the health check bit.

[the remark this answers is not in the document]
dikshant: Added!

cloud-controller-manager/do/loadbalancers.go
dikshant: Added!

docs/controllers/services/examples/udp-with-healthcheck.yml
```diff
@@ -0,0 +1,50 @@
...
+        app: tcp-udp-server
+    spec:
+      containers:
+      - name: tcp-udp-server
+        image: example.com/example/test-tcp-udp-server:latest
+        imagePullPolicy: Always
+        ports:
+        - containerPort: 10001
+          protocol: TCP
+        - containerPort: 10001
```
timoreimann, line 49: Shouldn't this say 10002 to match the `targetPort`?

docs/controllers/services/examples/udp-with-healthcheck.yml
```diff
@@ -0,0 +1,50 @@
...
+          protocol: UDP
```
timoreimann, line 50: nit: newline missing

timoreimann (commented):
Final comment/question, then we should be good. :)

docs/controllers/services/examples/udp-with-healthcheck.yml
dikshant: Oops. Fixed!

docs/controllers/services/examples/udp-with-healthcheck.yml
dikshant: Ah whoops. Great catch! Fixed!

PR #410, opened 2022-02-12
dockerfile: update alpine base image to the latest to fix any security issues

cpanato:
using an old alpine version can contain several security issues and be very outdated
so bumping to the latest one available

also if use `trivy` we got

```
2022-02-12T16:28:10.982+0100    WARN    This OS version is no longer supported by the distribution: alpine 3.6.5
```

/assign @timoreimann


timoreimann (approved):
This is great, thanks.

PR #492, opened 2022-07-26
fix: don't delete attached load balancers

### LINKED ISSUE #454: Protect load balancer from being deleted
Currently load balancers are always deleted when the service is deleted unless the `disown` annotation is specified.
The `disown` annotation however disables all other features as well.

So what I would like to have is an annotation `protect: "true"` that leaves all features intact but prevents deletion such that the ip address of that load balancer can be reused if someone manages to delete the service.

rawkode:
When a Load Balancer is attached via the load-balancer-id annotation,
 thee CCM should not be responsible for deleting it as it was likely
 provisioned by another means.

Signed-off-by: David Flanagan <david@rawkode.dev>

Closes #454

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -162,6 +162,10 @@ const (
 	// disowned. Defaults to false.
 	annDODisownLB = "service.kubernetes.io/do-loadbalancer-disown"
 
+	// annDOProtect is the annotation specifying that the load-balancer should not be deleted
+	// Defaults to false.
+	annDOProtectLB = "service.kubernetes.io/do-loadbalancer-protect"
+
 	// defaultActiveTimeout is the number of seconds to wait for a load balancer to
 	// reach the active state.
 	defaultActiveTimeout = 90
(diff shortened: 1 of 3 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -1275,6 +1286,16 @@ func getDisownLB(service *v1.Service) (bool, error) {
 	return disownLB, nil
 }
 
+// getAttachedLB returns whether the load-balancer was attached to the service..
+// False is returned if not specified.
+func getAttachedLB(service *v1.Service) (bool, error) {
+	attachedLB, _, err := getBool(service.Annotations, annoDOLoadBalancerID)
```
timoreimann, line 1292: `annoDOLoadBalancerID` is managed by our CCM to more easily and directly look up LBs in the DO API. I'm not sure I understand why we should stop deleting LBs for which CCM added the annotation in the first place?

timoreimann (commented):
Hey @rawkode, thanks for your contribution!

I'm not sure I follow the rationale for your change though, more context in my comment.

If #454 is to be implemented, shouldn't we add a new annotation to support preserving LBs?

cloud-controller-manager/do/loadbalancers.go
timoreimann: I also feel we should have a new annotation to preserve LBs.

If it's about separating LBs managed by different CCM implementations, then [the LB class](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class) introduced in 1.24 might be worth considering adding support for in our CCM.

rawkode: It's not different CCM implementations. My Pulumi code provisions a DO LB and adds domain names to it for TLS, and then in my LB service I provide the load-balancer-id annotation to hook them up.

timoreimann: I see. A separate, new annotation would seem appropriate to me then.

rawkode: Pushed as new annotation

timoreimann (commented):
Thanks, this looks really good.

May I ask you to extend our annotation documentation at https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/docs/controllers/services/annotations.md as well?

rawkode:
Yeah. I'll get that updated

rawkode:
@timoreimann Should be good to go

cpanato (approved):
lgtm

timoreimann:
I'll have another look soon, thanks.

timoreimann:
Apologies for having dropped the ball on this one. By now, I think there are out-of-band alternatives to generally protect resources from deletion, e.g., kyverno. I'm going to close out this PR as I think the involved complexity is better managed by such tools.

Issue #511, opened 2022-08-30
Allow the region to be explicitly specified instead of using the Region metadata API


xmudrii:
This is a feature request to allow the region to be explicitly specified instead of using the Region metadata API. The reason for this is to allow setups where the CCM is not running on a DigitalOcean Droplet. For example, the CCM might be running in some management cluster (which might not be a cluster running on DigitalOcean), along with the other control plane compoenents, while the worker nodes are running on DigitalOcean.

[The function used the determine the region](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/ba6aef8dbc3a36a75999f33c368c30551be07c32/cloud-controller-manager/do/metadata.go#L30-L38) supports overriding the region with the `FAKE_REGION` environment variable. However, it's explicitly mentioned that it is only for testing, so I doesn't feel safe to use it in production.

As a resolution, it would be nice if `FAKE_REGION` could be renamed to something that's not only for testing, e.g. that it feels safer to use it in production. Alternatively, this might be a CLI flag.

shatoboar:
Hey, i would like to work on this :)
/assign

timoreimann:
Marko and I talked about this one on the Kubernetes Slack some time ago. This request does seem legitimate to me, especially given that we have had a similar need with our CSI driver and implemented it accordingly.

My suggestion would be to rename `FAKE_REGION` to `REGION` and let it take precedence if specified. The docs should be updated accordingly.

Our CCM is still at 0.x, so we shouldn't need to worry about backwards compatibility for a development-only feature.

timoreimann:
Thanks for taking over this one @shatoboar.

almereyda:
Now that #513 is merged, can this be closed?

timoreimann:
Release is yet to be done; but yeah, we can close the issue out already.

Thanks for reminding me @almereyda.

PR #513, opened 2022-09-09
Change FAKE_REGION to REGION

shatoboar:
This change addresses #511, and is needed for running the Digitalocean-CCM outside of Digitalocean.

### DIFF
--- a/cloud-controller-manager/do/cloud.go
+++ b/cloud-controller-manager/do/cloud.go
@@ -54,6 +54,7 @@ const (
 	metricsAddrEnv              string = "METRICS_ADDR"
 	publicAccessFirewallNameEnv string = "PUBLIC_ACCESS_FIREWALL_NAME"
 	publicAccessFirewallTagsEnv string = "PUBLIC_ACCESS_FIREWALL_TAGS"
+	regionEnv                   string = "REGION"
 )
 
 var version string
--- a/cloud-controller-manager/do/cloud.go
+++ b/cloud-controller-manager/do/cloud.go
@@ -109,9 +109,9 @@ func newCloud() (cloudprovider.Interface, error) {
 		return nil, fmt.Errorf("failed to create godo client: %s", err)
 	}
 
-	region, err := dropletRegion()
+	region, err := dropletRegion(doClient.Regions)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get region from droplet metadata: %s", err)
+		return nil, fmt.Errorf("failed to determine region; %v", err)
 	}
 
 	clusterID := os.Getenv(doClusterIDEnv)
--- a/cloud-controller-manager/do/cloud.go
+++ b/cloud-controller-manager/do/cloud.go
@@ -111,7 +111,7 @@ func newCloud() (cloudprovider.Interface, error) {
 
 	region, err := dropletRegion(doClient.Regions)
 	if err != nil {
-		return nil, fmt.Errorf("failed to determine region; %v", err)
+		return nil, fmt.Errorf("failed to determine region: %v", err)
 	}
 
 	clusterID := os.Getenv(doClusterIDEnv)
--- a/cloud-controller-manager/do/metadata.go
+++ b/cloud-controller-manager/do/metadata.go
@@ -46,7 +46,7 @@ func dropletRegion(regionsService godo.RegionsService) (string, error) {
 		return "", fmt.Errorf("failed to determine if region is valid: %s", err)
 	}
 	if !validRegion {
-		return "", errors.New(fmt.Sprintf("invalid region specified: %s", region))
+		return "", fmt.Errorf("invalid region specified: %s", region)
 	}
 
 	klog.Infof("Using region %q from environment variable", region)
@@ -74,7 +74,7 @@ func listAllRegions(regionsService godo.RegionsService) ([]godo.Region, error) {
 	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
 	defer cancel()
 
-	list := make([]godo.Region, 0)
+	var list []godo.Region
 
 	listOptions := &godo.ListOptions{
 		Page:    1,
@@ -84,7 +84,7 @@ func listAllRegions(regionsService godo.RegionsService) ([]godo.Region, error) {
 	for {
 		regions, resp, err := regionsService.List(ctx, listOptions)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("regions list request failed: %w", err)
 		}
 
 		if resp == nil {
@@ -100,7 +100,7 @@ func listAllRegions(regionsService godo.RegionsService) ([]godo.Region, error) {
 
 		page, err := resp.Links.CurrentPage()
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to get current page number in pagination: %w", err)
 		}
 
 		listOptions.Page = page + 1
--- a/cloud-controller-manager/do/metadata.go
+++ b/cloud-controller-manager/do/metadata.go
@@ -61,7 +61,7 @@ func isValidRegion(region string, regionsService godo.RegionsService) (bool, err
 	}
 
 	for _, reg := range regions {
-		if reg.Name == region {
+		if reg.Slug == region {
 			return true, nil
 		}
 	}
--- a/cloud-controller-manager/do/metadata.go
+++ b/cloud-controller-manager/do/metadata.go
@@ -17,11 +17,14 @@ limitations under the License.
 package do
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	"os"
 
+	"github.com/digitalocean/godo"
+	"github.com/pkg/errors"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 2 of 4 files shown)


cloud-controller-manager/do/metadata.go
```diff
@@ -17,24 +17,49 @@ limitations under the License.
...
-		klog.Warningf("Using fake region %q from environment variable FAKE_REGION -- this should only be set for testing purposes!", region)
-		return region, nil
+	// REGION can be specified for local testing.
+	region := os.Getenv("REGION")
+	if region == "" {
+		return httpGet(dropletRegionMetadataURL)
 	}
-	return httpGet(dropletRegionMetadataURL)
+
+	if !validRegion(region) {
```
timoreimann, line 41: Could we also document in the README that the DO Regions service will be used to validate an explicitly specified region? This may be helpful for users that tighten network access for security reasons.

timoreimann (commented):
Thanks for the PR! I left a few line-specific comments.

Could we also change the error message [here](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/699bfbc352602ce9c1c51b12abba16c4b5904387/cloud-controller-manager/do/cloud.go#L114) from `"failed to get region from droplet metadata"` to `"failed to determine region"` since we're now not always using the metadata service?

cloud-controller-manager/do/metadata.go
```diff
@@ -17,24 +17,49 @@ limitations under the License.
...
+		return "", errors.New("invalid region specified")
+	}
+
+	klog.Warningf("Using region %q from environment variable", region)
+	return region, nil
+}
+
+// checks whether the given region is a valid DO region
```
timoreimann, line 49: GoDocs for functions/methods should always start with the function/method name, i.e., `// validRegion checks whether [...]`.

shatoboar:
Hi, thanks for the comments :) I fixed your suggestions, could you take a look again?

cloud-controller-manager/do/cloud.go
```diff
@@ -109,9 +109,9 @@ func newCloud() (cloudprovider.Interface, error) {
 		return nil, fmt.Errorf("failed to create godo client: %s", err)
 	}
 
-	region, err := dropletRegion()
+	region, err := dropletRegion(doClient.Regions)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get region from droplet metadata: %s", err)
+		return nil, fmt.Errorf("failed to determine region; %v", err)
```
timoreimann, line 114: Can we use a colon instead of a semicolon?

cloud-controller-manager/do/metadata.go
```diff
@@ -17,27 +17,63 @@ limitations under the License.
...
-		return region, nil
+func dropletRegion(regionsService godo.RegionsService) (string, error) {
+	region := os.Getenv("REGION")
+	if region == "" {
+		return httpGet(dropletRegionMetadataURL)
+	}
+
+	validRegion, err := isValidRegion(region, regionsService)
+	if err != nil {
+		return "", err
```
timoreimann, line 45: Could you annotate/wrap the error like

```golang
return "", fmt.Errorf("failed to determine if region is valid: %s", err)
```

cloud-controller-manager/do/metadata.go
```diff
@@ -17,27 +17,63 @@ limitations under the License.
...
+	}
+	if !validRegion {
+		return "", errors.New("invalid region specified")
```
timoreimann, line 48: Let's include the invalid region in the output so it's easier for the user to debug.

cloud-controller-manager/do/metadata.go
```diff
@@ -17,27 +17,63 @@ limitations under the License.
...
+	}
+
+	klog.Warningf("Using region %q from environment variable", region)
```
timoreimann, line 51: I think it's okay to use `klog.Info` here now that setting `REGION` becomes a legitimate thing not limited to testing only.

cloud-controller-manager/do/metadata.go
```diff
@@ -17,27 +17,63 @@ limitations under the License.
...
+}
+
+// isValidRegion checks whether the given region is a valid DO region
+func isValidRegion(region string, regionsService godo.RegionsService) (bool, error) {
+	listOptions := &godo.ListOptions{
+		Page:    1,
+		PerPage: resultsPerPage,
+	}
+
+	regions, _, err := regionsService.List(context.Background(), listOptions)
```
timoreimann, line 62: I suggest we set a timeout-based context on each `List()` invocation to ensure the request doesn't hang indefinitely in case of weird networking problems. I think 30 seconds should be enough.

cloud-controller-manager/do/metadata.go
```diff
@@ -17,27 +17,63 @@ limitations under the License.
...
+	}
+
+	klog.Warningf("Using region %q from environment variable", region)
+	return region, nil
+}
+
+// isValidRegion checks whether the given region is a valid DO region
+func isValidRegion(region string, regionsService godo.RegionsService) (bool, error) {
+	listOptions := &godo.ListOptions{
+		Page:    1,
```
timoreimann, line 58: We're never bumping the page. Please see how we do it with the droplet service that I referred to in my previous review.

shatoboar: Ah sorry, fixed it now

cloud-controller-manager/do/metadata.go
```diff
@@ -17,27 +17,99 @@ limitations under the License.
...
+	if region == "" {
+		return httpGet(dropletRegionMetadataURL)
+	}
+
+	validRegion, err := isValidRegion(region, regionsService)
+	if err != nil {
+		return "", fmt.Errorf("failed to determine if region is valid: %s", err)
+	}
+	if !validRegion {
+		return "", errors.New(fmt.Sprintf("invalid region specified: %s", region))
```
timoreimann, line 49: We can use `fmt.Errorf` here.

cloud-controller-manager/do/metadata.go
```diff
@@ -17,27 +17,99 @@ limitations under the License.
...
+
+	return false, nil
+}
+
+func listAllRegions(regionsService godo.RegionsService) ([]godo.Region, error) {
+	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	list := make([]godo.Region, 0)
```
timoreimann, line 77: I don't think the zero is needed. We can just do `make([]godo.Region)` or `var list []godo.Region`.

cloud-controller-manager/do/metadata.go
```diff
@@ -17,27 +17,99 @@ limitations under the License.
...
+	listOptions := &godo.ListOptions{
+		Page:    1,
+		PerPage: resultsPerPage,
 	}
-	return httpGet(dropletRegionMetadataURL)
+
+	for {
+		regions, resp, err := regionsService.List(ctx, listOptions)
+		if err != nil {
+			return nil, err
```
timoreimann, line 87: Since we now have multiple occasions in this function where we return errors, let's wrap/annotate this one with context through `fmt.Errorf()`.

Similar in line 103.

timoreimann (commented):
A few more small items remaining, then we should be good.

shatoboar:
Hi @timoreimann , could you take one more look at it? Thanks :)

cloud-controller-manager/do/metadata.go
```diff
@@ -17,27 +17,99 @@ limitations under the License.
...
+	listOptions := &godo.ListOptions{
+		Page:    1,
+		PerPage: resultsPerPage,
 	}
-	return httpGet(dropletRegionMetadataURL)
+
+	for {
+		regions, resp, err := regionsService.List(ctx, listOptions)
+		if err != nil {
+			return nil, fmt.Errorf("regions list request failed: %w", err)
```
timoreimann, line 87: I think we should use `%s` here and in line 103 to avoid exposing the API surface (in terms of error types we make accessible) unnecessarily.

timoreimann (commented):
One final minor thing :)

timoreimann:
I just tried this out locally by setting `REGION` to `fra1` but getting an error:

```
1 main.go:67] Cloud provider could not be initialized: could not init cloud provider \"digitalocean\": failed to determine region: invalid region specified: fra1\n","stream":"stderr","time":"2022-10-04T07:55:19.141267853Z"}
```

Something is still off here. 🤔

cloud-controller-manager/do/metadata.go
```diff
@@ -17,27 +17,99 @@ limitations under the License.
...
+
+// isValidRegion checks whether the given region is a valid DO region
+func isValidRegion(region string, regionsService godo.RegionsService) (bool, error) {
+	regions, err := listAllRegions(regionsService)
+	if err != nil {
+		return false, err
+	}
+
+	for _, reg := range regions {
+		if reg.Name == region {
```
timoreimann, line 64: This must use [reg.Slug](https://github.com/digitalocean/godo/blob/dd71fd5fc8df50446e6a853669d17d6fa3bf01af/regions.go#L25).

timoreimann:
Got it: https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/513/files#r986547512

@shatoboar could you please address this one?

shatoboar:
> Got it: https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/513/files#r986547512
> 
> @shatoboar could you please address this one?

Done :D

timoreimann:
@shatoboar looks like we need to run `go mod tidy && go mod vendor` as well.

cloud-controller-manager/do/metadata.go
```diff
@@ -17,27 +17,99 @@ limitations under the License.
...
+)
 
 // dropletRegion returns the region of the currently running program.
-func dropletRegion() (string, error) {
-	// FAKE_REGION can be specified for local testing.
-	if region := os.Getenv("FAKE_REGION"); region != "" {
-		klog.Warningf("Using fake region %q from environment variable FAKE_REGION -- this should only be set for testing purposes!", region)
-		return region, nil
+func dropletRegion(regionsService godo.RegionsService) (string, error) {
+	region := os.Getenv("REGION")
```
timoreimann, line 39: We have a number of environment variables names defined at the top of `cloud.go` already. Could we extract a constant and reference it here?

timoreimann:
Doing some final testing tomorrow and will merge afterwards if everything goes well. 🤞

timoreimann:
Working as expected. Thanks for the contribution @shatoboar.

PR #540, opened 2022-10-28
update k8s dependencies for 1.25

mo-rieger:
Unfortunately, dependabot updates every k8s dependency on its own, which causes failures. So I created this PR to group all k8s dependencies updates. 

CLOSES: CON-7650


timoreimann (approved):
LGTM.

We may want to disable dependabot for Kubernetes-related dependencies.

varshavaradarajan (approved):

mo-rieger:
There is an open issue for grouping dependencies from 2019 https://github.com/dependabot/dependabot-core/issues/1296 .  May we should consider moving to [renovate](https://www.mend.io/free-developer-tools/renovate/) since there seems to be no movement from dependabot in this direction.

PR #545, opened 2022-11-11
LBAAS-2427 Add project id annotation for load balancers CCM

StephenVarela:


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -1261,6 +1268,16 @@ func getEnableBackendKeepalive(service *v1.Service) (bool, error) {
 	return enableBackendKeepalive, nil
 }
 
+func getProjectID(service *v1.Service) string {
+	projectID, ok := service.Annotations[annDOLBProjectID]
+
+	if !ok {
+		return ""
```
StephenVarela, line 1275: Not sure currently if there is a way to obtain the cluster project id and use that as the default. 

As it stands this loadbalancer will be assigned to the user's default project

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -1261,6 +1268,16 @@ func getEnableBackendKeepalive(service *v1.Service) (bool, error) {
...
+	}
+
+	return projectID
```
anitgandhi, line 1278: nit: strictly speaking, since the fallback case is to just return an empty string anyway, this could be simplified to just `return service.Annotations[annDOLBProjectID]`

docs/controllers/services/annotations.md
```diff
@@ -167,3 +167,8 @@ See also the [elaborate example](/docs/controllers/services/examples/README.md#c
 **Note**
 
 You have to supply the value as string (ex. `"true"`, not `true`), otherwise you might run into a [k8s bug that throws away all annotations on your `Service` resource](https://github.com/kubernetes/kubernetes/issues/59113).
+
+
+## service.beta.kubernetes.io/do-loadbalancer-project-id
+
+Indicates which project to associate the loadbalancer to. This defaults to the users default project
```
anitgandhi, line 174:
```suggestion
Indicates which project to associate the Load Balancer with. If not specified, the Load Balancer will be placed in your default project.
```

cloud-controller-manager/do/loadbalancers_test.go
```diff
@@ -557,6 +557,60 @@ func Test_getCertificateID(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range testcases {
+		t.Run(test.name, func(t *testing.T) {
+			projectID := getProjectID(test.service)
+			if projectID != test.projectID {
+				t.Error("unexpected project ID")
+				t.Logf("expected %q", test.projectID)
+				t.Logf("actual: %q", projectID)
```
anitgandhi, line 608: nit, log messages are kinda unnecessary/annoying here imo

```suggestion
				t.Errorf("project ID did not match. expected %v, got %v", test.projectID, projectID)
```

anitgandhi (approved):

StephenVarela:
Just learned that there is a process which will sync the project_id of the load balancer with the cluster. Closing this PR as it doesnt seem necessary to add a separate annotation for project_id

Issue #555, opened 2022-12-07
IPv6 address missing in nodes status


sachintiptur:
We integrated DO CCM into a dual-stack(IPv4 + IPv6) cluster. 
Nodes are up with both IPv4 and IPv6 addresses but ipv6 addresses are not updated in node status(`kubectl describe nodes`).

DO CCM flags used:

`digitalocean-cloud-controller-manager --kubeconfig=/etc/kubernetes/kubeconfig/kubeconfig --leader-elect=false`

Can you let us know whether IPv6 support is available in the version `v0.1.40` or is there any specific configuration that we are missing.

timoreimann:
👋 

the CCM implementation [fetches v4 addresses only at this point](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/306f43abeaf47eaa47b847ecd26ce61f3c751101/cloud-controller-manager/do/common.go#L127-L145). It should be fairly easy to include v6 ones from the DO API / the droplet struct we're working on already as well here.

The one part I'm not absolutely sure about is how those addresses would need to be reflected in the `[]v1.NodeAddress` slice. Does it suffice to add another set of items of types `v1.NodeInternalIP` and `v1.NodeExternalIP`, respectively, to put the v6 addresses in? That is, is Kubernetes able to distinguish the address protocols without any additional tagging in the struct (for which I currently don't see a way to do it, i.e., there's no `v1.NodeInternalIPv6` type or something equivalent).

This may need some discovery.

timoreimann:
(And yeah, additional binary flags might be needed too.)

timoreimann:
I might be wrong and we don't necessarily need a new flag. (CCMs like ours mostly inherit flags from kube-controller-manager, so we could check there for any IPv6-ish options.)

I think we'd want a flag or environment variable though to control which IP addresses should be collected (IPv4 as today or IPv4 and IPv6) to make sure that clients expecting the node address fields to be IPv4-populated only do not break.

timoreimann:
@sachintiptur CCM does not create VM / droplets but reads IP information from droplets retrieved via the public API. As such, we'd probably want to look at [the `PublicIPv6()` method](https://github.com/digitalocean/godo/blob/d8d6ec42f6f218bd3704477c2fc078b20650af44/droplets.go#L98).

timoreimann:
Oh gotcha @sachintiptur.

[...]

PR #561, opened 2022-12-12
IPv6 address support in node addresses

sachintiptur:
Signed-off-by: Sachin Tiptur <sachin@kubermatic.com>

This PR adds a new env `DO_IP_ADDR_FAMILIES` to configure IP families required to add IP addresses in node's status.
This also populates the nodes address in the same order as `DO_IP_ADDR_FAMILIES` is configured.
This fixes the [issue #555](https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/555)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ package-hosted directory like this:
 
 ```bash
 cd cloud-controller-manager/cmd/digitalocean-cloud-controller-manager
-REGION=fra1 DO_ACCESS_TOKEN=your_access_token DO_IP_ADDR_FAMILIES=ipv4 go run main.go \
+REGION=fra1 DO_ACCESS_TOKEN=your_access_token go run main.go \
   --kubeconfig <path to your kubeconfig file>                     \
   --leader-elect=false --v=5 --cloud-provider=digitalocean
 ```
@@ -79,10 +79,10 @@ You might also need to provide your DigitalOcean access token in
 the cloud controller to start, but in that case, you will not be able to
 validate integration with DigitalOcean API.
 
-The `DO_IP_ADDR_FAMILIES` is used to configure the required IP familes and the
-order in which address should be populated in nodes status. The accepted values
-are one of the `{"", "ipv4", "ipv6", "ipv4,ipv6", "ipv6,ipv4"}`.IPv4 is the 
-default, if not set or empty.
+The `DO_IP_ADDR_FAMILIES` is used to configure the required IP families and the
+order in which the addresses should be populated in nodes status. The accepted values
+are one of the `"ipv4", "ipv6"` or a comma-separated list of multiple IP address
+families. IPv4 is the default, if not set or empty.
 
 Please note that if you use a Kubernetes cluster created on DigitalOcean, there
 will be a cloud controller manager running in the cluster already, so your local
--- a/cloud-controller-manager/do/cloud.go
+++ b/cloud-controller-manager/do/cloud.go
@@ -62,9 +62,6 @@ const (
 
 var version string
 
-var ipFamilyOpts = []string{"", "ipv4", "ipv6", "ipv4,ipv6", "ipv6,ipv4"}
-var ipFamilies string
-
 type tokenSource struct {
 	AccessToken string
 }
@@ -162,10 +159,8 @@ func newCloud() (cloudprovider.Interface, error) {
 		addr = fmt.Sprintf("%s:%s", addrHost, addrPort)
 	}
 
-	// var ipFamilies string
 	ipf, set := os.LookupEnv(doIPAddrFamiliesEnv)
-	ipFamilies = ipf
-	if set && !containsString(ipFamilyOpts, ipFamilies) {
+	if set && !validateAndSetIPFamilies(ipf) {
 		return nil, fmt.Errorf("invalid value set for environment variable %q", doIPAddrFamiliesEnv)
 	}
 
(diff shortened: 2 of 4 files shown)


cloud-controller-manager/do/common.go
```diff
@@ -128,18 +130,28 @@ func allLoadBalancerList(ctx context.Context, client *godo.Client) ([]godo.LoadB
 func nodeAddresses(droplet *godo.Droplet) ([]v1.NodeAddress, error) {
 	var addresses []v1.NodeAddress
 	addresses = append(addresses, v1.NodeAddress{Type: v1.NodeHostName, Address: droplet.Name})
+	ipFamilies := os.Getenv(doIPAddrFamiliesEnv)
```
timoreimann, line 133: I think it'd be great if we parsed the env var during the startup phase only (and use the opportunity to error out if something unsupported is passed in). The logic processing the parameters (i.e., this function) can then work off of more strongly typed data (say, a typed string with two constant values).

cloud-controller-manager/do/common.go
```diff
@@ -128,18 +130,28 @@ func allLoadBalancerList(ctx context.Context, client *godo.Client) ([]godo.LoadB
 func nodeAddresses(droplet *godo.Droplet) ([]v1.NodeAddress, error) {
```
timoreimann, line 130: Can we extend our unit tests for this new functionality?

timoreimann (commented):
Thanks for the contribution, appreciated.

It'd be great if we could extend the README / documentation as well.

README.md
```diff
@@ -62,7 +62,7 @@ package-hosted directory like this:
 
 ```bash
 cd cloud-controller-manager/cmd/digitalocean-cloud-controller-manager
-REGION=fra1 DO_ACCESS_TOKEN=your_access_token go run main.go \
+REGION=fra1 DO_ACCESS_TOKEN=your_access_token DO_IP_ADDR_FAMILIES=ipv4 go run main.go \
```
timoreimann, line 65: I'd omit the usage from this example since it's not strictly needed when running locally.

README.md
```diff
@@ -79,6 +79,11 @@ You might also need to provide your DigitalOcean access token in
 the cloud controller to start, but in that case, you will not be able to
 validate integration with DigitalOcean API.
 
+The `DO_IP_ADDR_FAMILIES` is used to configure the required IP familes and the
```
timoreimann, line 82: Typo: families

README.md
```diff
@@ -79,6 +79,11 @@ You might also need to provide your DigitalOcean access token in
...
+order in which address should be populated in nodes status. The accepted values
```
timoreimann, line 83:
```suggestion
order in which the addresses should be populated in nodes status. The accepted values
```

cloud-controller-manager/do/cloud.go
```diff
@@ -158,6 +162,13 @@ func newCloud() (cloudprovider.Interface, error) {
 		addr = fmt.Sprintf("%s:%s", addrHost, addrPort)
 	}
 
+	// var ipFamilies string
+	ipf, set := os.LookupEnv(doIPAddrFamiliesEnv)
+	ipFamilies = ipf
+	if set && !containsString(ipFamilyOpts, ipFamilies) {
```
timoreimann, line 168: I'd rather prefer pulling the string split we do further down the stack up here. This allows us to work with a slice of tokenized family IDs in the business logic and do all the parsing in the bootstrap logic.

Let's also trim white spaces from split items.

cloud-controller-manager/do/cloud.go
```diff
@@ -57,10 +57,14 @@ const (
 	publicAccessFirewallTagsEnv string = "PUBLIC_ACCESS_FIREWALL_TAGS"
 	regionEnv                   string = "REGION"
 	doAPIRateLimitQPSEnv        string = "DO_API_RATE_LIMIT_QPS"
+	doIPAddrFamiliesEnv         string = "DO_IP_ADDR_FAMILIES"
 )
 
 var version string
 
+var ipFamilyOpts = []string{"", "ipv4", "ipv6", "ipv4,ipv6", "ipv6,ipv4"}
```
timoreimann, line 65: Having to specify all orderings seems a bit unclean and doesn't allow for white spaces to be used. How about introducing a custom string type with two constants instead, one for each family?

We can then also reference the family in both the bootstrap and the business logic.

README.md
```diff
@@ -79,6 +79,11 @@ You might also need to provide your DigitalOcean access token in
...
+are one of the `{"", "ipv4", "ipv6", "ipv4,ipv6", "ipv6,ipv4"}`.IPv4 is the
```
timoreimann, line 84: Why do we have curly braces here? It almost looks like JSON but we don't require JSON, do we?

Can we rephrase this in a more user friendly manner, e.g., say `one of "ipv4", "ipv6", or a comma-separated list of multiple IP address families`?

cloud-controller-manager/do/common.go
```diff
@@ -129,17 +130,41 @@ func nodeAddresses(droplet *godo.Droplet) ([]v1.NodeAddress, error) {
...
 
-	publicIP, err := droplet.PublicIPv4()
-	if err != nil || publicIP == "" {
-		return nil, fmt.Errorf("could not get public ip: %v", err)
+	return addresses, nil
+}
+
+func discoverAddress(droplet *godo.Droplet, family string) ([]v1.NodeAddress, error) {
+	var addresses []v1.NodeAddress
+	if family == "ipv4" || family == "" {
```
timoreimann, line 146: Using a switch statement could make things slightly more readable here I imagine.

cloud-controller-manager/do/cloud.go
sachintiptur: Sure, makes sense, fixed it.

README.md
sachintiptur: Fixed

README.md
sachintiptur: fixed

README.md
sachintiptur: fixed

README.md
sachintiptur: fixed

PR #562, opened 2022-12-14
Add firewall support for Loadbalancers

jrolheiser:


StephenVarela (approved):
LGTM

docs/controllers/services/annotations.md
```diff
@@ -180,4 +180,20 @@ You have to supply the value as string (ex. `"true"`, not `true`), otherwise you
 
 ## service.beta.kubernetes.io/do-loadbalancer-http-idle-timeout-seconds
 
-Specifies the HTTP idle timeout configuration in seconds. If not specified, the default is 60 seconds.
\ No newline at end of file
+Specifies the HTTP idle timeout configuration in seconds. If not specified, the default is 60 seconds.
+
+## service.beta.kubernetes.io/do-loadbalancer-deny-rules
+
+Specifies the comma seperated DENY firewall rules for the load-balancer, supports IP and CIDR based rules
```
timoreimann, line 187: We could make documentation slightly more maintainable by omitting this part about which types are exactly supported (and possibly reference some product documentation page instead) since we transparently pass the (comma-separated) annotation value to the LB API. That should allow us to automatically and always support the deny/allow rules that the backend supports without having to update the CCM.

Similar for the deny rule.

timoreimann (commented):
This looks great, thanks.

One comment about whether we should document which types we maintain vs referring to authoritative sources. This may sound picky but based on experience, customers and users will rely on the information we put here, so there's a chance to minimize risk for outdating.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -1194,6 +1220,16 @@ func getPorts(service *v1.Service, anno string) ([]int, error) {
...
 }
 
+// getStrings returns the strings for the given service and annotation.
+func getStrings(service *v1.Service, anno string) []string {
+	vals, ok := service.Annotations[anno]
+	if !ok || len(vals) == 0 {
+		return nil
+	}
+
+	return strings.Split(vals, ",")
```
bbassingthwaite, line 1230: Either here or on the backend, we should `Trim` each entry. This in case someone were to add a space after the comma: `ip:1.2.3.4, ip:2.3.4.5`

jrolheiser: I'd prefer to add trimming on the backend as that's a centralized point.

jrolheiser:
@timoreimann Should be good for another look now. I addressed comments and rebased to master to resolve dep conflicts.

timoreimann:
This looks good to me. @bbassingthwaite okay from your end as well?

bbassingthwaite (approved):

timoreimann:
Thanks @jrolheiser

PR #566, opened 2022-12-29
load balancers: multiple fixes and updates

anitgandhi:


timoreimann (commented):
Good cleanup. Thanks Anit.

Re: the algorithm deprecation notice: I can see we still use the corresponding annotation in multiple example files. Should we remove those everywhere as well?

waynr:
I just opened #567 after receiving a report from a customer that annotation set that looks like
```
    service.beta.kubernetes.io/do-loadbalancer-certificate-id: <certificate-id>
    service.beta.kubernetes.io/do-loadbalancer-http-ports: '80'
    service.beta.kubernetes.io/do-loadbalancer-http3-port: '443'
    service.beta.kubernetes.io/do-loadbalancer-protocol: http3
```
was failing with an error that looked like
```
Error syncing load balancer: failed to ensure load balancer: failed to build load-balancer request: invalid protocol "http3" specified in annotation "service.beta.kubernetes.io/do-loadbalancer-protocol"
```
But then I noticed this PR, which could potentially be a better fit for that fix. Just FYI, happy to keep my PR open and follow it through.

anitgandhi:
@waynr your PR is fine as-is , I think it's largely orthogonal to mine 👍

anitgandhi:
@timoreimann good catch; fixed!

varshavaradarajan (approved):

PR #568, opened 2023-01-04
docs: minor wording fix

jkpe:
annotation should be 'service.beta.kubernetes.io/do-loadbalancer-allow-rules'


timoreimann (approved):
Thanks, good catch.

PR #569, opened 2023-01-05
CON-7849 / Extract annotation parsing on service objects into the DO package

elijahomolo:
https://jira.internal.digitalocean.com/browse/CON-7849


timoreimann (changes requested):
The annotations were moved to a different file only without adding a new package. Go-wise, this requires creating a new directory (like `common`) that should probably live in the root so that it can be referenced by both the CCM implementation and the future web hook.

I'd also appreciate if we could add a proper PR description. (I think it's fine to share publicly what we are aiming for.) Bonus points for adding a public, tracking Github issue that outlines the goal / feature.

cloud-controller-manager/do/lb_annotations.go
```diff
@@ -0,0 +1,167 @@
+/*
+Copyright 2020 DigitalOcean
```
timoreimann, line 2: Could use the opportunity to update to 2022.

timoreimann (approved):
LGTM with one optional suggestion.

cloud-controller-manager/do/lb_annotations.go
cpanato: Would say 2023 :)

timoreimann: Haha yeah good catch. I'm still stuck in the past it seems. 😀

PR #572, opened 2023-01-20
Allow specifying certificate by name

dverbeir:
Using the 'service.beta.kubernetes.io/do-loadbalancer-certificate-name' service annotation, the certificate for an HTTPS load balancer can bow be specified by name. The corresponding certificate ID is then retrieved using the DO API at load balancer creation/update time and is used from then on.

So far, the certificate had to be specified using the '.../do-loadbalancer-certificate-id' annotation and this was automatically updated to reflect an ID change resulting from a certificate renewal. However if, after such a renewal, the service was re-created without re-fecthing the latest certificate ID, the service remained stuck waiting for its external IP.

This is for example very useful when deploying with a CD system such as Flux, where the deployment specification comes from a source repository. In that case, one would not expect having to update the source repository as a result of an automatic certificate renewal.

With this change, the source repository can specify the certificate by name so it won't have to be updated after the certificate is renewed. And a redeployment of the service will automatically fetch the current ID of the referenced certificate.

### DIFF
--- a/cloud-controller-manager/do/certificates.go
+++ b/cloud-controller-manager/do/certificates.go
@@ -16,8 +16,29 @@ limitations under the License.
 
 package do
 
+import (
+	"context"
+
+	"github.com/digitalocean/godo"
+)
+
 const (
 	// DO Certificate types
 	certTypeLetsEncrypt = "lets_encrypt"
 	certTypeCustom      = "custom"
 )
+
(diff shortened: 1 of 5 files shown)


timoreimann:
Apologies for not getting to your PR in time, @dverbeir -- we pretty much dropped the ball on this one. :(

On the upside, the feature was now implemented via #722 and is going to get shipped to all supported clusters in the near term.

PR #584, opened 2023-02-09
loadbalancers: support loadBalancerSourceRanges

J0sh0nat0r:
This PR introduces support for `loadBalancerSourceRanges` as a configuration source for the recently added firewall support.

I understand this may be somewhat redundant given that it's actually _less_ powerful than the annotations, however, given that it's standardised and easy to support I think it's nice-to-have.

I wasn't quite sure where to document this so I left a note under the existing annotations documentation, but if there's somewhere better suited (the examples README perhaps?) let me know

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -796,7 +796,7 @@ func buildForwardingRule(service *v1.Service, port *v1.ServicePort, protocol, ce
 
 func buildFirewall(service *v1.Service) *godo.LBFirewall {
 	denyRules := getStrings(service, annDODenyRules)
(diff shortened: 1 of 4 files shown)


llDrLove:
Hi @J0sh0nat0r , this indeed looks like a nice addition. Could you possibly fix the conflicting files in this PR ? We also want this feature supported on our end. Would you be able to continue working on this PR ? I'm asking because the last changes are going back quite some time and we wouldn't want to duplicate the work if there is already a PR in progress. Thanks !

gottwald:
Closing this PR in favor of #731 (rebased and slightly adapted version of this PR)

Thank you @J0sh0nat0r for your contribution!

PR #585, opened 2023-02-09
set -buildvcs=false

elijahomolo:
CI builds fail with the following error: 

```
Status: Downloaded newer image for golang:1.19
error obtaining VCS status: exit status 1[28](https://github.com/digitalocean/digitalocean-cloud-controller-manager/actions/runs/4136491568/jobs/7150424330#step:4:29)
	Use -buildvcs=false to disable VCS stamping.
make: *** [Makefile:69: compile] Error 1
```

This PR adds the -buildvcs=false flag to the CCM binary build step


MorrisLaw (commented):
How does this work if setting `-buildvcs=auto` ?

MorrisLaw (approved):
You know what, I read some more (https://go-review.googlesource.com/c/go/+/400454) and I think that `-buildvcs=false` is ok. It should be explicitly false until we don't want it to be. But for now we seem to have not known about this and it wasn't explicitly set to true so it's safe to assume it's not needed at the moment (the stamping of VCS metadata). It would just require removing this field and defaulting to true or making it `-buildvcs=true`.

Approving this PR, thank you @elijahomolo !

PR #586, opened 2023-02-09
CON-7851\Register Lb validation handler

elijahomolo:
https://jira.internal.digitalocean.com/browse/CON-7851

This PR addresses CON-7851 by adding a new validation handler to handle LB type requests. This PR only adds the functionality to parse the LB service type and pass Create and Update requests to the appropriate LBAAS DO API endpoints. A followup PR will add validation for individual configurations.


elijahomolo:
PR to fix CI: https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/585

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+	clusterID := os.Getenv(doClusterIDEnv)
+	if clusterID == "" {
+		fmt.Println("missing cluster id")
+	}
+	region := "nyc1"
+	if region == "" {
+		fmt.Println("missing region id")
+	}
+
+	nodePools, _, err := v.gClient.Kubernetes.ListNodePools(ctx, clusterID, &godo.ListOptions{})
```
timoreimann, line 119: This part is problematic for two reasons:

1. It couples the implementation to DOKS again, which we should not do. The upstream portion of CCM fetches the droplet IDs from the Node objects.
2. It ignores filters that the upstream part of the CCM implementation applies, which we don't. Re-implementing them here would be quite tedious and brittle, but not having them means we would have to iterate through possibly hundreds of worker nodes on large clusters. That puts much more stress on the validator and makes it more susceptible to cluster failures.

From our (CCM) perspective, we ideally wouldn't want to include any droplet IDs at all since we only care about validating the non-node customer configuration (i.e., the annotations). Perhaps we can chat with the LB team to see if excluding those (either unconditionally or when a specific flag on the validation request is set) would be possible.

The second best option is to be minimalistic and fetch a single droplet ID through the available Nodes just to satisfy the node portion of the validation request. I personally don't prefer this option since it is more complex (we would need to page through Nodes and stop as soon as we have one with a valid droplet ID / provider ID) and brittle (it does not work when no Nodes with droplet IDs exist).

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+	err = v.validateUpdate(ctx, svc, lbRequest, v.gClient)
+	if err != nil {
+		fmt.Printf("failed to update load balancer")
+		v.Log.Error(err, "failed to update load balancer")
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+	return admission.Allowed("validating if request type is UPDATE")
+}
+
+func (v *DOKSLBServiceValidator) validateCreate(ctx context.Context, lbRequest *godo.LoadBalancerRequest, doClient *godo.Client) error {
```
timoreimann, line 173: If we move `doClient` to `v`, then we don't have to pass it in.

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+
+	lbRequest, err := v.buildRequest(svc.Name, region, dropletIDs, forwardingRules)
+	if err != nil {
+		v.Log.Error(err, "failed to build lb request")
+		//fmt.Printf("failed to build lb request", err)
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	err = v.decoder.DecodeRaw(req.OldObject, svc)
+	if err != nil {
```
timoreimann, line 155: Is there a way to distinguish between genuine recoding errors and the old object not existing? For instance, is `req.OldObject` possibly nil when the Service is brand new?

Besides, we cannot assume that absence of the old object alone implies that a create request is needed: for instance, users trying to create an LB with an erroneous configuration predating the webhook would have a Service object but the LB might have never been created. For such a case, we would want to send out a create validation request.

 I think it's more reliable to look at the status field of the old Service object (if it exists) and determine if we have a field set indicating that an LB create has started. An easy way to discover these fields is to create a managed LB and check the backing Service object right after that.

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+			TargetProtocol: "http",
+			TargetPort:     80,
+			CertificateID:  "",
+			TlsPassthrough: false,
+		}}
+
+	lbRequest, err := v.buildRequest(svc.Name, region, dropletIDs, forwardingRules)
+	if err != nil {
+		v.Log.Error(err, "failed to build lb request")
+		//fmt.Printf("failed to build lb request", err)
```
timoreimann, line 150: Can be removed.

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+			dropletID, err := strconv.Atoi(node.DropletID)
+			if err != nil {
+				v.Log.Error(err, "failed to retrieve droplet ids")
+				return admission.Errored(http.StatusConflict, err)
+			}
+			dropletIDs = append(dropletIDs, dropletID)
+		}
+	}
+
+	forwardingRules := []godo.ForwardingRule{
```
timoreimann, line 137: Let's have a comment here describing the placeholder nature of the rules and that further development is required.

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+		if err != nil {
+			v.Log.Error(err, "failed to validate lb creation")
+			return admission.Denied("failed to validate load balancer request")
+		}
+		return admission.Allowed("valid lb create request")
+	}
+
+	err = v.validateUpdate(ctx, svc, lbRequest, v.gClient)
+	if err != nil {
+		fmt.Printf("failed to update load balancer")
```
timoreimann, line 166: I think this can be dropped.

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+		//fmt.Printf("failed to build lb request", err)
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	err = v.decoder.DecodeRaw(req.OldObject, svc)
+	if err != nil {
+		err = v.validateCreate(ctx, lbRequest, v.gClient)
+		if err != nil {
+			v.Log.Error(err, "failed to validate lb creation")
+			return admission.Denied("failed to validate load balancer request")
```
timoreimann, line 159: Can we reuse the same message as in the log?

We should also include the actual error in the response.

Similar comments for the update portion.

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		v.Log.Error(err, "failed to create load balancer")
+		return err
+	}
+
+	return nil
+}
+
+func (v *DOKSLBServiceValidator) validateUpdate(ctx context.Context, svc *v1.Service, lbRequest *godo.LoadBalancerRequest, doClient *godo.Client) error {
+	currentLBID := svc.Annotations["kubernetes.digitalocean.com/load-balancer-id"]
```
timoreimann, line 184: In some circumstances, the LB ID may be missing. One example is other (user) controllers fighting with ours and removing the annotation.

It's probably best if we handle this gracefully by logging and error if the LB ID is absent but proceeding to allow the webhook request in order to not break LB management in such cases.

I'd also move the LB ID extraction and validation logic to the call site so that this method can be focused on executing the actual validation request.

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+	_, _, err := doClient.LoadBalancers.Update(ctx, currentLBID, lbRequest)
+	if err != nil {
+		fmt.Printf("THIS IS AN ERROR: %v", err)
```
timoreimann, line 187: No need to log here since we do it at the call site already.

The whole block can be simplified to 

```go
_, _, err := doClient.LoadBalancers.Update(ctx, currentLBID, lbRequest)
return err
```

Similar for `validateCreate`.

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+		return err
+	}
+	return nil
+}
+
+func (v *DOKSLBServiceValidator) buildRequest(name string, region string, dropletIDs []int, forwardingRules []godo.ForwardingRule) (*godo.LoadBalancerRequest, error) {
```
timoreimann, line 193: This method does not need to have an `error` return type given we never return a non-nil error.

timoreimann, line 193: At its current form, this method could be function. However, as I suggested elsewhere, we should initialize the region at startup time once and store it on a field in `v`, at which point we want this to be a method.

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+	return &godo.LoadBalancerRequest{
+		Name:            name,
+		DropletIDs:      dropletIDs,
+		Region:          region,
+		ForwardingRules: forwardingRules,
+		ValidateOnly:    true,
+	}, nil
+}
+
+func (v *DOKSLBServiceValidator) InjectDecoder(d *admission.Decoder) error {
```
timoreimann, line 203: I don't see this being invoked. Is it part of some interface, or could the method be removed?

tmp/webhook_manifests/cluster_role_binding.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: dokswebhooks-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+  - kind: ServiceAccount
+    name: default
+    namespace: default
```
timoreimann, line 12: nit: newline missing.

Same for other manifest files.

tmp/webhook_manifests/cluster_role_binding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: dokswebhooks-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
```
timoreimann, line 8: Can we have a dedicated (cluster) role right away? I wouldn't want anyone to use `cluster-admin` for this, even when the webhook is still in the development stage.

tmp/webhook_manifests/cluster_role_binding.yaml
```diff
@@ -0,0 +1,12 @@
...
+subjects:
+  - kind: ServiceAccount
+    name: default
```
timoreimann, line 11: Dedicated ServiceAccount would be great here too.

cloud-controller-manager/do/doks_lb_validator_test.go
```diff
@@ -0,0 +1,371 @@
...
+
+	//"github.com/go-logr/logr"
+	v1 "k8s.io/api/admission/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	//"sigs.k8s.io/controller-runtime/pkg/client"
```
timoreimann, line 36: Let's drop the commented imports.

tmp/webhook_manifests/manifest.yaml
```diff
@@ -0,0 +1,152 @@
...
+  annotations:
+    cert-manager.io/inject-ca-from: default/dokswebhooks-serving-cert
+  name: dokswebhooks-validating-webhook-configuration
+webhooks:
+  - admissionReviewVersions:
+      - v1
+    clientConfig:
+      # Because cert-manager lives in the control plane and the webhook lives in
+      # the data plane, we cannot use cert-manager directly to manage the CA
+      # bundle. We use C3 instead to populate the field during reconciliation.
```
timoreimann, line 132: Please drop this.

tmp/webhook_manifests/manifest.yaml
```diff
@@ -0,0 +1,152 @@
...
+      caBundle: ""
+      # The webhook service lives in the control plane, so use a service *URL*
```
timoreimann, line 134: Drop this.

tmp/webhook_manifests/manifest.yaml
```diff
@@ -0,0 +1,152 @@
...
+kind: Service
+metadata:
+  name: dokswebhooks-service
+  namespace: default
+spec:
+  ports:
+    - port: 443
+      protocol: TCP
+      targetPort: 9443
+#    - port: 40001
```
timoreimann, line 37: Can we drop this?

tmp/webhook_manifests/webhook-rolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
timoreimann, line 2: Why do we need two ClusterRoleBindings? 🤔

timoreimann (commented):
I'd appreciate if we could introduce a flag / environment variable at this point already that causes the webhook to not do anything (and log clearly about it) unless it is enabled specifically.

This is to guard against users eventually discovered our (unfinished) work and trying it out, possibly creating more tickets due to confusion.

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+		return admission.Allowed("")
+	}
+
+	// check if request is for LB deletion
+	if svc.Spec.Type == v1.ServiceTypeLoadBalancer && svc.DeletionTimestamp != nil {
+		return admission.Allowed("the loadbalancer is deleting")
+	}
+
+	// initialize DO Client if not already initialized
+	if v.gClient == nil {
```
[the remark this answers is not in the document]
MorrisLaw: I suggested this change since we do _not_ initialize the doclient in main and any call to Handle would overwrite the dummy client we initialized in our test. The only thing initialized in main, iirc, is the Log.

If we prefer to have doclient initialization in main, then this would remove the need for the check in it's entirety and we can just use v.gClient.

P.S.
> One missing bit seems to be that we do not store our initialized client in the DOKSLBServiceValidator struct.

Isn't it stored down below in the bottom of this condition via `v.gClient = doClient`?

cloud-controller-manager/do/doks_lb_validator.go
MorrisLaw: In it's current state, `v.gClient` will give you the `doClient` you need. So passing in the `doClient` is redundant.

The only thing needed now, as Timo suggested, is to initialize `v.gClient` in main instead of in this file and also remove the `doClient` value from this function signature and just call `v.gClient` as needed.

go.mod
```diff
@@ -4,7 +4,8 @@ go 1.19
 
 require (
 	github.com/davecgh/go-spew v1.1.1
-	github.com/digitalocean/godo v1.97.0
+	github.com/digitalocean/godo v1.97.1-0.20230308191258-ca2138d84446
```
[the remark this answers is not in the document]
MorrisLaw: This should be 1.98. Change might've got lost in the rebase here. Pretty sure he had it at one point.

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	v.Log.V(10).Info("checking received request")
+	if svc.Spec.Type != v1.ServiceTypeLoadBalancer {
+		return admission.Allowed("")
+	}
+
+	// check if request is for LB deletion
+	if svc.Spec.Type == v1.ServiceTypeLoadBalancer && svc.DeletionTimestamp != nil {
```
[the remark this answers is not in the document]
elijahomolo: @timoreimann Decoding the request to extract the service needs to happen first prior to checking the LB type however, I have combined the service type validation  and deletion timestamp check to remove an extra request:

```
	v.Log.V(6).Info("checking received request")
	if svc.Spec.Type != v1.ServiceTypeLoadBalancer || svc.DeletionTimestamp != nil {
		return admission.Allowed("request is either not a load balancer or delete")
	}
```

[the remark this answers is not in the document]
timoreimann: 👍 
 
We don't need to check the Service type here anymore since we already return early in line 91 if it's not about a load-balancer. (Alternatively, we eliminate the `if` statement from that line above.)

cloud-controller-manager/do/doks_lb_validator.go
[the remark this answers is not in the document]
timoreimann: Yes, but it should be stored on the struct at bootstrapping time.

cloud-controller-manager/do/doks_lb_validator.go
elijahomolo: Removed

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+	return doClient, nil
+}
+
+// DOKSLBServiceValidator ...
+func (v *DOKSLBServiceValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	svc := &v1.Service{}
+	v.Log.V(10).Info("decoding received request")
+	err := v.decoder.Decode(req, svc)
+	if err != nil {
+		v.Log.Error(err, "bad request")
```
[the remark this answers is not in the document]
elijahomolo: @timoreimann re: 

> Can we also include this context in the returned HTTP response right below?

 Do you mean using a different status code or including the same error message in admission response?

cloud-controller-manager/cmd/digitalocean-webhook-server/main.go
```diff
@@ -0,0 +1,61 @@
...
+	ctrlruntimelog.SetLogger(ll)
+
+	ll.Info("getting config")
+
+	server := webhook.Server{}
+
+	server.Register("/validate-doks-lb-service", &webhook.Admission{Handler: &do.DOKSLBServiceValidator{Log: ll}})
+	ll.Info("Registering Webhook server handlers")
+	if err := server.StartStandalone(ctrl.SetupSignalHandler(), scheme); err != nil {
+		return err
```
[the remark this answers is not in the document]
elijahomolo: @timoreimann I'm not entirely sure about this 

> SetupSignalHandler registered for SIGTERM and SIGINT. A stop channel is returned which is closed on one of these signals. If a second signal is caught, the program is terminated with exit code 1.

Is there any way I can validate the how a cancelled context would be handled?

cloud-controller-manager/do/doks_lb_validator.go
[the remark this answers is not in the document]
timoreimann: I think the comment is outdated now and might be addressed, but just in case: I meant that we do not return a naked error as in

```golang
return err
```

but something like

```golang
return fmt.Errorf("failed to decode request: %s", err)
```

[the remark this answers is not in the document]
elijahomolo: @timoreimann the implementation requires that an admission response be required. I can add the print statement and then return an admission denial to satisfy the required return type, something like this although it feels a bit redundant: 

```
	v.Log.V(6).Info("decoding received request")
	err := v.decoder.Decode(req, svc)
	if err != nil {
		v.Log.Error(err, "failed to decode request")
		fmt.Errorf("failed to decode request: %s", err)
		return admission.Denied("failed to decode request")
	}
```

cloud-controller-manager/do/doks_lb_validator.go
```diff
@@ -0,0 +1,206 @@
...
+	if v.gClient == nil {
+		doClient, err := initDOClient()
+		if err != nil {
+			v.Log.Error(err, "failed to initialize DO client")
+			return admission.Errored(http.StatusConflict, err)
+		}
+		v.gClient = doClient
+	}
+
+	clusterID := os.Getenv(doClusterIDEnv)
```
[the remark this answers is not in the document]
elijahomolo: We're going to leverage the cloud initialization of these values. @MorrisLaw this is something we probably should look into how these fields are implemented by CCM

cloud-controller-manager/do/doks_lb_validator.go
elijahomolo: @timoreimann in a case where the annotation is removed, would it be an empty string or nil?

cloud-controller-manager/do/doks_lb_validator.go
elijahomolo: @time req.OldObject will error if empty:

> It errors out if rawObj is empty i.e. containing 0 raw bytes.

From an existing LB, the only field under the status is the loadbalancer IP:
```
status:
  loadBalancer:
    ingress:
    - ip: 144.126.249.45
```
There is a creationTimestamp in the metadata, perhaps we can use that instead?

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,185 @@
...
+	clusterID := os.Getenv(doClusterIDEnv)
+	if clusterID == "" {
+		fmt.Println("missing cluster id")
+	}
+	region := "nyc1"
+	if region == "" {
+		fmt.Println("missing region id")
+	}
+
+	nodePools, _, err := v.gClient.Kubernetes.ListNodePools(ctx, clusterID, &godo.ListOptions{})
```
MorrisLaw, line 116: After speaking a bit about this PR and it's use of the DOKS client w/ Timo and Varsha. They had some good points as to why we shouldn't use it and how we could address it.

Let's consider making LB create requests w/ tags and not Droplet IDs (https://docs.digitalocean.com/reference/api/kafka-beta-api-reference/#operation/loadBalancers_create)

That way... we could just use droplet tags (not droplet IDs) to make that request since droplet tags could simply be the 3 we use for droplets associated w/ a cluster:
- `k8s`
- `k8s:<cluster-id>`
- `k8s:worker`

In other words, you can construct these tags in-memory in your code, don't need to pull them from anywhere since we already have in-memory access to the cluster ID.

cloud-controller-manager/do/doks_lb_validator.go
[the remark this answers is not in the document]
elijahomolo: Nevermind, this is resolved

cloud-controller-manager/do/doks_lb_validator.go
timoreimann: The IP would only be available once an LB is fully provisioned, so I'm not sure we can use that.

Can you check if `status.loadBalancer.ingress` exists right after a managed LB is created? If it does even though the IP address is still missing, then we could use that as a trigger.

The other possibility is to check the LB ID annotation, though using something on the status should be a bit more reliable.

timoreimann: Can you answer this question of mine:

>Is req.OldObject possibly nil when the Service is brand new?

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,138 @@
...
+		if err != nil {
+			v.Log.Error(err, "failed to validate lb creation")
+			return admission.Denied("failed to validate lb creation")
+		}
+		return admission.Allowed("valid lb create request")
+	}
+
+	currentLBID := svc.Annotations["kubernetes.digitalocean.com/load-balancer-id"]
+	if currentLBID == "" {
+		v.Log.Error(fmt.Errorf("missing loadbalancer id in annotation"), "missing kubernetes.digitalocean.com/load-balancer-id annotation")
```
timoreimann, line 103: If the annotation is missing, then it means we need to do a create request as well.

cloud-controller-manager/do/doks_lb_validator.go
elijahomolo: I confirmed that oldObject is empty on a new service and added a condition to ensure that it is on create requests

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,138 @@
...
+
+	v.Log.V(6).Info("checking received request")
+	if svc.Spec.Type != v1.ServiceTypeLoadBalancer || svc.DeletionTimestamp != nil {
+		return admission.Allowed("the service object is either not a load balancer or being deleted")
+	}
+
+	tags := []string{
+		"k8s",
+		"k8s:worker",
+		fmt.Sprintf("k8s: %v", v.ClusterID),
```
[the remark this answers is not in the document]
elijahomolo: @timoreimann I've added a check for a 'droplet-tag' that a user can specify using an annotation and if none is provided, construct one using the in-memory cluster-ID. Is there a conversation that I have missed regarding an alternative that doesn't require a tag?

cloud-controller-manager/cmd/digitalocean-webhook-server/main.go
```diff
@@ -32,6 +36,64 @@ var (
...
+		return nil, fmt.Errorf("environment variable %q is required", doAccessTokenEnv)
+	}
+
+	tokenSource := &tokenSource{
+		AccessToken: token,
+	}
+
+	var opts []godo.ClientOpt
+
+	opts = append(opts, godo.SetUserAgent("digitalocean-webhook-server/"+version))
```
varshavaradarajan, line 77: The name is too generic, we have a bunch of webhooks already.

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,152 @@
...
+		}}
+
+	lbRequest := v.buildRequest(svc.Name, v.Region, tag, forwardingRules)
+
+	if req.OldObject.Object == nil {
+		if len(svc.Status.LoadBalancer.Ingress) == 0 || svc.Status.LoadBalancer.Ingress[0].IP == "" {
+			err = v.validateCreate(ctx, lbRequest)
+			if err != nil {
+				v.Log.Error(err, "failed to validate lb creation")
+				return admission.Denied("failed to validate lb creation")
```
varshavaradarajan, line 98: Surface the err since the whole point of this is to show to the user why the request is denied and why if the req was allowed, it wouldn't create the LB.

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,152 @@
...
+			}
+			return admission.Allowed("valid lb create request")
+		}
+	}
+
+	err = v.decoder.DecodeRaw(req.OldObject, svc)
+	if err != nil {
+		v.Log.Error(err, "failed to decode request")
```
varshavaradarajan, line 106: failed to decode existing obj

cloud-controller-manager/do/kubernetes_lb_validator.go
varshavaradarajan: This applies wherever you return Denied.

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,365 @@
...
+			name: "Allow Update happy path",
+			req: admission.Request{AdmissionRequest: fakeAdmissionRequest(
+				fakeService("test2"), fakeService("old-service"))},
+			expectedAllowed: true,
+		},
+	}
+
+	for _, test := range testcases {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
```
varshavaradarajan, line 224: run in parallel

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,365 @@
...
+			// setup client
+			var testKubernetesService godo.KubernetesService
+			testKubernetesService = &fakeKubernetesService{}
+
+			gClient := godo.NewFromToken("test-token")
+			gClient.Kubernetes = testKubernetesService
```
varshavaradarajan, line 230: refactor to be gClient.Kubernetes = &fakeKubernetesService{}

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,365 @@
...
+}
+
+func (f *fakeKubernetesService) ListNodePools(ctx context.Context, clusterID string, opts *godo.ListOptions) ([]*godo.KubernetesNodePool, *godo.Response, error) {
+	cluster := fakeCluster()
+	return cluster.NodePools, newFakeOKResponse(), nil
+}
+
+func Test_Handle(t *testing.T) {
+	//cluster := fakeCluster()
+	testcases := []struct {
```
varshavaradarajan, line 152: add tests for denied.

varshavaradarajan (changes requested):

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,155 @@
...
+		}}
+
+	lbRequest := v.buildRequest(svc.Name, v.Region, tag, forwardingRules)
+
+	if req.OldObject.Object == nil {
+		if len(svc.Status.LoadBalancer.Ingress) == 0 || svc.Status.LoadBalancer.Ingress[0].IP == "" {
+			err = v.validateCreate(ctx, lbRequest)
+			if err != nil {
+				v.Log.Error(err, "failed to validate lb creation")
+				fmt.Printf("failed to validate lb creation: %v", err)
```
varshavaradarajan, line 98: It can't be a prrintf statement. Should be:
return admission.Denied(fmt.Sprintf("failed to validate lb creation: %v", err))

varshavaradarajan (changes requested):

cloud-controller-manager/cmd/digitalocean-webhook-server/main.go
```diff
@@ -40,25 +102,40 @@ func main() {
...
-		fmt.Fprintf(os.Stderr, "failed to construct cr client %v\n", err)
+		ll.Error(err, "failed to initialize DO client")
+		os.Exit(1)
 	}
 
-	server.Register("/validate-doks-lb-service", &webhook.Admission{Handler: &do.DOKSLBServiceValidator{Client: c, Log: ll}})
-	ll.Info("Registering Webhook server handlers")
-	if err = server.StartStandalone(ctrl.SetupSignalHandler(), scheme); err != nil {
-		return err
+	clusterID := os.Getenv(doClusterIDEnv)
```
timoreimann, line 117: It seems like we're using this for the tag of the LB's target droplets. At some point, I suggest to not conflate the cluster ID parameter with the droplet tag: this happens to be same for DOKS clusters but certainly does not have to be for DIY clusters. So my suggestion would be to have a new, dedicated environment flag for the webhook (like `DROPLET_TAG`) and use that.

I'm still not overly fond of the idea to use a tag and would much rather get rid of the need to having to specify a droplet at all; but let's defer that discussion to later. My prime interest here is that we do not couple existing parameters we later need to uncouple again.

cloud-controller-manager/cmd/digitalocean-webhook-server/main.go
```diff
@@ -40,25 +102,40 @@ func main() {
...
+	if clusterID == "" {
+		fmt.Println("missing cluster id")
```
timoreimann, line 119: This should yield an error.

cloud-controller-manager/cmd/digitalocean-webhook-server/main.go
```diff
@@ -40,25 +102,40 @@ func main() {
...
+	}
+
+	region := os.Getenv(regionEnv)
+	if region == "" {
+		fmt.Println("missing region id")
+	}
+
+	server.Register("/validate-doks-lb-service", &webhook.Admission{Handler: &do.KubernetesLBServiceValidator{
+		Log:       ll,
+		GClient:   doClient,
```
timoreimann, line 129: Can we call the field `DOClient` as well to be consistent?

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,152 @@
...
+		return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to decode request: %v", err))
+	}
+
+	v.Log.V(6).Info("checking received request")
+	if svc.Spec.Type != v1.ServiceTypeLoadBalancer || svc.DeletionTimestamp != nil {
+		return admission.Allowed("ignoring the service which is either not a load balancer or is being deleted")
+	}
+
+	// TODO: document the requirement for a droplet-tag annotation for DIY users that need to pass in an identifier to droplets they wish to have behind their lb
+	tag := svc.Annotations["kubernetes.digitalocean.com/droplet-tag"]
```
timoreimann, line 73: Why do we have this alternative way? Could we remove it and require a tag to be specified until we are clear on if there's a way to validate requests without having to pass in a droplet?

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,152 @@
...
+			TargetProtocol: "http",
+			TargetPort:     80,
+			CertificateID:  "",
+			TlsPassthrough: false,
+		}}
+
+	lbRequest := v.buildRequest(svc.Name, v.Region, tag, forwardingRules)
+
+	if req.OldObject.Object == nil {
+		if len(svc.Status.LoadBalancer.Ingress) == 0 || svc.Status.LoadBalancer.Ingress[0].IP == "" {
```
timoreimann, line 94: Did you verify at which point the `.ingress` field is added to the Service's status as I [suggested in an earlier comment](https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/586#discussion_r1152104694)?

If it gets added right after an LB create request has been made (even when the IP address isn't populated yet), then we could use that as our sole trigger without looking at the old object or the LB annotation. That should simplify the logic significantly.

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,152 @@
...
+			v.Log.Error(err, "failed to validate lb creation")
+			return admission.Denied(fmt.Sprintf("failed to validate lb creation: %v", err))
+		}
+		return admission.Allowed("can not validate lb without annotation: kubernetes.digitalocean.com/load-balancer-id annotation")
+	}
+
+	err = v.validateUpdate(ctx, currentLBID, lbRequest)
+	if err != nil {
+		v.Log.Error(err, "failed to update load balancer")
+		return admission.Denied(fmt.Sprintf("failed to validate load balancer update: %v", err))
```
timoreimann, line 124: Curious: does the validation API / response semantics allow distinguishing between validation rejections and other errors (e.g., network connectivity problems) that we could/should use to more clearly indicate to the caller why the webhook request failed?

cloud-controller-manager/cmd/digitalocean-webhook-server/Dockerfile
```diff
@@ -22,22 +22,22 @@ COPY go.mod go.mod
 COPY go.sum go.sum
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
-COPY vendor/ vendor/
 
+COPY vendor/ vendor/
 
 # Copy the go source
 COPY  cloud-controller-manager/cmd/digitalocean-webhook-server/main.go main.go
```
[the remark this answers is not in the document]
elijahomolo: This requires running the go mod vendor command prior to running the build

cloud-controller-manager/do/kubernetes_lb_validator.go
elijahomolo: I've added a method to check the error status code then use the return to determine whether the admission response should be an error or denial

cloud-controller-manager/do/kubernetes_lb_validator.go
timoreimann: As discussed out-of-band, my suggestion doesn't work because the `ingress` status field is always there (and only populated when the IP address has been issued).

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,165 @@
...
+			TlsPassthrough: false,
+		}}
+
+	lbRequest := v.buildRequest(svc.Name, v.Region, v.DropletTag, forwardingRules)
+
+	var resp *godo.Response
+
+	if req.OldObject.Object == nil {
+		if len(svc.Status.LoadBalancer.Ingress) == 0 || svc.Status.LoadBalancer.Ingress[0].IP == "" {
+			resp, err = v.validateCreate(ctx, lbRequest)
```
timoreimann, line 89: We cannot assume a create workflow based on the ingress IP address being absent since IP address assignment takes a while: between a managed LB create request having been submitted and CCM polling for an IP address, we should do update requests. (Strictly speaking, those will fail as well since the LB API does not support updates while LBs are being created currently, though that could theoretically change in the future.)

I think the logic to determine if we do a create or update validation request can be as following:

1. If we have an old service object AND identify it as being associated to a created LB (see below) => do an update request
1. otherwise (which means the service is brand new or it wasn't a service of type `LoadBalancer` before) => do a create request

The criteria for identifying a created LB from a service is that any of the following conditions must be true:

- the ingress slice in the status field is non-empty
- The LB ID annotation is found

I think this should cover it, but please sanity-check my logic.

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,165 @@
...
+
+	err = v.decoder.DecodeRaw(req.OldObject, svc)
+	if err != nil {
+		v.Log.Error(err, "failed to decode existing object")
+		return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to decode request: %v", err))
+	}
+
+	currentLBID := svc.Annotations["kubernetes.digitalocean.com/load-balancer-id"]
+	if currentLBID == "" {
+		v.Log.Error(fmt.Errorf("missing loadbalancer id in annotation"), "missing kubernetes.digitalocean.com/load-balancer-id annotation")
```
timoreimann, line 110: This should not be considered an error: the old service object might not be of type `LoadBalancer`; or it might be but the configuration is invalid, or something prevents the annotation from being set (e.g., a conflicting operator/controller in the cluster).

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (v *KubernetesLBServiceValidator) validateUpdate(ctx context.Context, currentLBID string, lbRequest *godo.LoadBalancerRequest) (*godo.Response, error) {
+	_, resp, err := v.GClient.LoadBalancers.Update(ctx, currentLBID, lbRequest)
+	return resp, err
+}
+
+func (v *KubernetesLBServiceValidator) buildRequest(name string, region string, tag string, forwardingRules []godo.ForwardingRule) *godo.LoadBalancerRequest {
+	return &godo.LoadBalancerRequest{
+		Name:            name,
+		Tag:             tag,
```
timoreimann, line 148: As clarified out of band, an LB validation request does not require passing a list of droplet IDs or a tag. This means we can rip out any logic related to that and focus on validating the configuration only.

cloud-controller-manager/cmd/digitalocean-webhook-server/Dockerfile
```diff
@@ -22,22 +22,23 @@ COPY go.mod go.mod
...
+COPY cloud-controller-manager/cmd/digitalocean-webhook-server/ cloud-controller-manager/cmd/digitalocean-webhook-server/
 
 # Copy the go source
 COPY  cloud-controller-manager/cmd/digitalocean-webhook-server/main.go main.go
 COPY  cloud-controller-manager/do cloud-controller-manager/do
 
-
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -v -a -o doks_webhooks main.go
+RUN go mod vendor
```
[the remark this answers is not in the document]
elijahomolo: The container build failed referencing that running go mod vendor was required, I'll replicate again and grab the exact error.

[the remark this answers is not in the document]
elijahomolo: I still get the error after executing go mod tidy then vendor  prior to building the container. I think this is because I set -mod=vendor

```
#14 0.226       To ignore the vendor directory, use -mod=readonly or -mod=mod.
#14 0.226       To sync the vendor directory, run:
#14 0.226               go mod vendor
```

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,181 @@
...
+		// decode raw object
+		err = v.decoder.DecodeRaw(req.OldObject, svc)
+		if err != nil {
+			v.Log.Error(err, "failed to decode existing object")
+			return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to decode request: %v", err))
+		}
+
+		currentLBID := svc.Annotations["kubernetes.digitalocean.com/load-balancer-id"]
+		// perform create if load-balancer-id annotation is missing
+		if currentLBID == "" {
```
timoreimann, line 105: Please invert the logic: check if the conditions apply to make an update validation request (annotation found and ingress set), and otherwise do a create validation request. This way, we don't need multiple `validateCreate` requests, which is currently the case.

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,181 @@
...
+			resp, err = v.validateCreate(ctx, lbRequest)
+			if err != nil {
+				if v.checkError(resp) {
+					v.Log.Error(err, "failed to validate lb creation")
+					return admission.Denied(fmt.Sprintf("failed to validate lb creation: %v", err))
+				}
+				v.Log.Error(err, "failed to validate lb creation")
+				return admission.Errored(int32(resp.StatusCode), err)
+			}
+			return admission.Allowed("can not validate lb without annotation: kubernetes.digitalocean.com/load-balancer-id annotation")
```
timoreimann, line 116: Does this make sense? We just executed a validation successfully, haven't we?

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+func (v *KubernetesLBServiceValidator) InjectDecoder(d *admission.Decoder) error {
+	v.decoder = d
+	return nil
+}
+
+func (v *KubernetesLBServiceValidator) checkError(response *godo.Response) bool {
+	if response.StatusCode == http.StatusUnprocessableEntity {
```
timoreimann, line 177: This will cause a nil pointer dereference if `response` is nil, which is the case whenever we hit a network error (i.e., we don't even get an HTTP response). We'll need to check for that.

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,181 @@
...
+		ValidateOnly:    true,
+	}
+}
+
+func (v *KubernetesLBServiceValidator) InjectDecoder(d *admission.Decoder) error {
+	v.decoder = d
+	return nil
+}
+
+func (v *KubernetesLBServiceValidator) checkError(response *godo.Response) bool {
```
timoreimann, line 176: Can we rename this to something like `isValidationError`? The current name isn't very informative.

cloud-controller-manager/cmd/digitalocean-webhook-server/main.go
```diff
@@ -40,25 +101,34 @@ func main() {
...
 	server := webhook.Server{}
-	c, err := client.New(config, client.Options{Scheme: scheme})
+
+	DOClient, err := initDOClient()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "failed to construct cr client %v\n", err)
+		return fmt.Errorf("failed to initialize DO client %v", err)
+	}
+
+	region := os.Getenv(regionEnv)
```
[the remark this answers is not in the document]
elijahomolo: I left this for DIY users that may choose to pass their own variable but I'll remove it

cloud-controller-manager/cmd/digitalocean-webhook-server/main.go
```diff
@@ -40,25 +100,29 @@ func main() {
...
+		GClient: DOClient,
+	}})
+
+	ll.Info("Registering loadbalancer validation webhook server handlers")
+	if err := server.StartStandalone(ctrl.SetupSignalHandler(), scheme); err != nil {
+		ll.Error(err, "failed to start loadbalancer validation webhook server")
+		os.Exit(1)
 	}
-	ll.Info("Webhooks server started")
+	ll.Info("loadbalancer validation webhook server started")
```
timoreimann, line 125: [AFAICS](https://github.com/kubernetes-sigs/controller-runtime/blob/cd65cb25d314f40a329a688f4714fe3282589e97/pkg/webhook/server.go#L291C16-L293), `StartStandalone` only returns when the server terminates. it also logs graceful shutdowns and errors, meaning we can and should drop this log line.

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,176 @@
...
+	"net/http"
+
+	"github.com/digitalocean/godo"
+	"github.com/go-logr/logr"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	lbIDAnnotation string = "kubernetes.digitalocean.com/load-balancer-id"
```
timoreimann, line 31: Please don't copy but export and reuse [this one](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/0fea136d928d94a435a3d89ecb03b212c9ee4e3b/cloud-controller-manager/do/lb_annotations.go#L22).

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,176 @@
...
+			v.Log.Error(err, "failed to decode existing object")
+			return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to decode request: %v", err))
+		}
+
+		currentLBID := svc.Annotations[lbIDAnnotation]
+		// perform update if associated with created lb
+		if (len(svc.Status.LoadBalancer.Ingress) > 0 && svc.Status.LoadBalancer.Ingress[0].IP != "") || currentLBID != "" {
+			v.Log.Info(fmt.Sprintf("updating lb id: %v", currentLBID))
+			resp, err = v.validateUpdate(ctx, currentLBID, lbRequest)
+			if resp == nil {
```
timoreimann, line 109: I don't think we need to check `resp`: godo always returns a non-nil error when the returned HTTP code is >= 400, meaning we should be able to work with `err` only.

Similar for the create path.

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,176 @@
...
+				v.Log.Error(err, "failed to validate lb update")
+				return admission.Denied(fmt.Sprintf("failed to validate lb update, could not return response body: %v", err))
+			}
+			if err != nil {
+				if v.checkError(resp) {
+					v.Log.Error(err, "failed to validate lb update")
```
timoreimann, line 115: Please use different error messages in this line and in line 118 to distinguish easily.

Similar for the create path.

cloud-controller-manager/do/kubernetes_lb_validator.go
timoreimann: Please address this comment.

cloud-controller-manager/do/kubernetes_lb_validator.go
timoreimann: Please address this comment.

cloud-controller-manager/do/kubernetes_lb_validator.go
elijahomolo: I thought this would resolve this [comment](https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/586#discussion_r1176092293):

timoreimann: Sorry, I meant to move that check into `isValidationError`: that way, the function is safe to be called no matter what `resp` looks like.

elijahomolo: Does [this](https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/586/files#diff-0b037574b49ca1618f3243d7b959eb12569ec91a928aaa9eab1d10859e31092dR158) refactor address the check?

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,401 @@
...
+		name            string
+		req             admission.Request
+		oldObject       runtime.RawExtension
+		gCLient         *godo.Client
+		lbRequest       *godo.LoadBalancerRequest
```
timoreimann, line 165: Is this used anywhere?

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,401 @@
...
+		expectedAllowed bool
+		expectedReason  string
```
timoreimann, line 167: If the reason is part of every denied and errored validation response, then please specify one everywhere where `expectedAllowed` is false: this ensures we are testing the right code path and not fall prey to a false positive.

It's fine to use an identifying reason substring (as opposed to exact string) for that to ease matching.

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,401 @@
...
+					Spec: corev1.ServiceSpec{
+						Type: corev1.ServiceTypeLoadBalancer,
+					},
+				},
+				nil,
+			)},
+			expectedAllowed: true,
+		},
+		{
+			name: "Deny CREATE happy path",
```
timoreimann, line 224: This sub-test looks 100% identical to the previous one except for `expectedAllowed` being different, and yet the test does not fail. If I am correct, then this indicates a bug in the test implementation (which I guess is around not comparing all combinations of the expected and actual allowed values as I indicated below).

Same thing for the `Allow Update happy path` and `Deny Update happy path` tests I believe.

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,401 @@
...
+			ctrlruntimelog.SetLogger(ll)
+
+			validator := &KubernetesLBServiceValidator{
+				Log:     ll,
+				decoder: decoder,
+				GClient: gClient,
+			}
+
+			res := validator.Handle(context.TODO(), test.req)
+			// if test.allowed & res.allowed are not equal
```
timoreimann, line 297: Can we drop this?

test/clients/controller_runtime.go
```diff
@@ -0,0 +1,304 @@
...
+// thereby allowing a (lightweight) mock-like usage.
+type ControllerRuntimeTestClient struct {
+	Fake client.WithWatch
+	ControllerRuntimeStubFuncs
+	ssaPatchedObjects []runtime.Object
+}
+
+// ControllerRuntimeStubFuncs are controller-runtime functions that can be
+// stubbed out.
+type ControllerRuntimeStubFuncs struct {
```
timoreimann, line 57: Do we need any of this at all?

timoreimann (commented):
I feel like we're still missing a few test cases around combinations of denied/errored responses in the create and update code paths. Could you please make sure we have these covered and validate all test cases are implemented correctly?

cloud-controller-manager/do/kubernetes_lb_validator_test.go
elijahomolo: Dropping this because it's handled by checking for error types as well as additional coverage will be added when individual cases for different annotations is added

timoreimann: Do we check the error types anywhere? I don't see it right now.

Also, even if we did I think we still want to test for exact or substring reason matching: otherwise, we cannot distinguish between different errors from the same type (e.g., create denied vs update denied) which is a common source for false positives in unit tests.

If we do this, then we can also drop `expectedAllow` in favor of `expectedError` (or `expectedErrorMessage`) where a nil/empty value implies that allowance should have been granted.

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,271 @@
...
+			resp:            newFakeUnprocessableResponse(),
+			err:             newFakeUnprocessableErrorResponse(),
+		},
+		{
+			name: "Deny Update validation error",
+			req: admission.Request{AdmissionRequest: fakeAdmissionRequest(
+				fakeService("test2"), fakeService("old-service"))},
+			expectedAllowed: false,
+			resp:            newFakeUnprocessableResponse(),
+			err:             newFakeUnprocessableErrorResponse(),
```
timoreimann, line 173: I don't think `resp` and `err` here can be same as in the `"Deny Update invalid configuration"` sub-test: AFAIU, they should yield different responses (errored vs denied), and so the faked response/error must be different too?

(This is probably also a good example why checking for exact/substring errors is important.)

elijahomolo: @timoreimann Errored response doesn't return a message like alllow and deny do. It does however return the passed in status code while the others return 200 or 403. I've added an assertion on the status code

timoreimann: Aren't we returning a `admission.Errored()` response when we hit any kind of (non-validation) error?

This returns a `Response` struct (just like `admission.Allowed()` and `admission.Denied()` do) which embeds the passed in error [here](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/a69e9c3b138c092f51787fcb90a796d33eccc1af/vendor/sigs.k8s.io/controller-runtime/pkg/webhook/admission/response.go#L56). This is what we should be able to assert on, thereby allowing us to distinguish between different error scenarios reliably (because the response code alone wouldn't be sufficient in all cases).

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,283 @@
...
+					return []godo.Region{{Name: "nyc3", Slug: "nyc3"}}, newFakeOKResponse(), nil
+				}}
+
+			decoder, err := admission.NewDecoder(scheme)
+			if err != nil {
+				t.Fatalf("failed to initialize decoder %s", err)
+			}
+
+			var logOpts []zap.Opts
+			ll := zap.New(logOpts...).WithName("webhook-validation-server")
```
timoreimann, line 212: It looks like `zap.New()` is variadic and thus would accept zero parameters as well, rendering `logOpts` obsolete.

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,283 @@
...
+				decoder: decoder,
+				GClient: gClient,
+			}
+
+			res := validator.Handle(context.TODO(), test.req)
+			if res.Allowed != test.expectedAllowed {
+				t.Fatalf("got allowed %v, want %v", res.Allowed, test.expectedAllowed)
+			}
+			if res.Result.Code != test.expectedStatusCode {
+				t.Fatalf("got allowed %v, want %v", res.Result.Code, test.expectedStatusCode)
```
timoreimann, line 226: s/allowed/status code/

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,283 @@
...
+			}
```
timoreimann, line 227: Here is where we should use `expectedMessage` as well based on the response message that I linked to in another comment.

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,295 @@
...
+				Log:     ll,
+				decoder: decoder,
+				GClient: gClient,
+			}
+
+			res := validator.Handle(context.TODO(), test.req)
+			if res.Allowed != test.expectedAllowed {
+				t.Fatalf("got %v, want %v", res.Allowed, test.expectedAllowed)
+			}
+			if res.Result.Reason != "" && string(res.Result.Reason) != test.expectedMessage {
```
timoreimann, line 233: I think the current logic misses a case: if neither the reason nor the message get populated (due to a bug) while there is an expected message, the test passes even though it should not.

With my suggestion elsewhere to assert on the error status code and use the `test.expectedErrorStatusCode` field to determine which field to compare against, the required adjustment should become more straight-forward. It could look something like this (untested, so please verify):

```golang
gotMessage := string(res.Result.Reason)
label := "reason"
if test.expectedErrorStatusCode != 0 {
  gotMessage = res.Result.Message
  label = "message"
}

if gotMessage != test.expectedMessage {
  t.Fatalf("got %s %q, want %q", label, gotMessage, test.expectedMessage)
}
```

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,295 @@
...
+				t.Fatalf("got %v, want %v", res.Result.Reason, test.expectedMessage)
```
timoreimann, line 234: If we use `%q` here and in line line 237, the output becomes slightly more readable since the message will be put in quotes.

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,295 @@
...
+	os.Setenv(regionEnv, "nyc3")
+
+	testcases := []struct {
+		name            string
+		req             admission.Request
+		gCLient         *godo.Client
+		expectedAllowed bool
+		resp            *godo.Response
+		err             error
+		expectedMessage string
```
timoreimann, line 84: I suggest we also add `expectedErrorStatusCode` that we populate for test cases where we expect `admission.Errored()` to be called. This allows us to assert that we pass down the upstream response code from the LB API response (that we fake out in our test) correctly to the caller.

As a bonus point, this also enables us to determine reliably if we should compare `expectedMessage` against the reason or the message field: it should be the former when `expectedErrorStatusCode` is zero and otherwise the latter.

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -0,0 +1,295 @@
...
+
+func newFakeUnprocessableErrorResponse() *godo.ErrorResponse {
+	return &godo.ErrorResponse{
+		Response: &http.Response{
+			Request: &http.Request{
+				Method: "FAKE",
+				URL:    &url.URL{},
+			},
+			StatusCode: http.StatusUnprocessableEntity,
+			Body:       ioutil.NopCloser(bytes.NewBufferString("test")),
```
timoreimann, line 69: nit: would use a string like "validation failure" so that it's clear what the fake is supposed to represent in test outputs.

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -121,15 +126,15 @@ func (v *KubernetesLBServiceValidator) Handle(ctx context.Context, req admission
...
+			v.Log.Error(err, "failed to validate lb creation")
+			return admission.Errored(int32(errorCode), errors.Wrap(err, "failed to validate lb create"))
 		}
-		v.Log.Error(err, "failed to validate lb creation, invalid configuration")
-		return admission.Denied(fmt.Sprintf("failed to validate lb creation: %v", err))
+		v.Log.Error(err, "invalid LB update configuration")
+		return admission.Denied(fmt.Sprintf("invalid LB update configuration: %v", err))
 	}
-	v.Log.Info(fmt.Sprintf("allowing create"))
+	v.Log.Info("allowing create")
```
timoreimann, line 137: nit: this says "allowing create" while in the upgrade path we say "lb update validated". A consistent wording would probably read slightly better.

cloud-controller-manager/do/kubernetes_lb_validator_test.go
```diff
@@ -154,41 +157,50 @@ func Test_Handle(t *testing.T) {
...
+			req: admission.Request{AdmissionRequest: fakeAdmissionRequest(
+				fakeService("new-test", annotations{}), fakeServiceWithStatus())},
 			expectedAllowed: true,
 			expectedMessage: "valid update request",
 		},
 		{
 			name: "Deny Update invalid configuration",
 			req: admission.Request{AdmissionRequest: fakeAdmissionRequest(
-				fakeService("test2", annotations{}), fakeService("old-service", annotations{annDOLoadBalancerID: "test"}))},
+				fakeService("new-test", annotations{}), fakeService("old-service", annotations{annDOLoadBalancerID: "test"}))},
```
timoreimann, line 189: Previous suggestion was to say `new-service` here.

cloud-controller-manager/do/kubernetes_lb_validator.go
```diff
@@ -0,0 +1,172 @@
...
+		{
+			EntryProtocol:  "http",
+			EntryPort:      80,
+			TargetProtocol: "http",
+			TargetPort:     80,
+			CertificateID:  "",
+			TlsPassthrough: false,
+		}}
+
+	v.Region, err = dropletRegion(v.GClient.Regions)
```
timoreimann, line 88: This should be part of the initialization: we don't want to hit the metadata service on each `Handle()` request.

timoreimann (commented):
I think one more round, then we should be good.

timoreimann:
@elijahomolo you seem to have merge commit markers in your `go.mod` file now. You need to remove them and run `go mod tidy && go mod vendor` afterwards to be safe.

timoreimann:
Closing in favor of #676 and several follow-up PRs that were submitted/merged by now to continue and complete the validation web hook work.

PR #631, opened 2023-05-12
Have dependabot ignore Kubernetes dependencies

timoreimann:
We always update all Kubernetes dependencies atomically and manually when we work on supporting the next minor version.


cpanato (approved):
Thanks!

PR #650, opened 2023-09-08
Add steps to update Go dependencies and version

llDrLove:


README.md
```diff
@@ -170,6 +170,15 @@ This will create a binary with version `dev` and docker image pushed to
...
 ## Release a new version
 
+### Update Go and dependencies 
+1. Update Go dependencies
+   ```shell
+   go get -u ./...
+   go mod tidy
+   go mod vendor
+   ```
+2. [Update Go version to latest GA version](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/go.mod#L3)
```
timoreimann, line 180: I think there's a way to reference a file without the full repo so that it stays stable even when, say, the repo is forked.

Additionally, I'd not reference a specific line because that's bound to become outdated if the target file is modified.

llDrLove: Sounds good. I refactored it to use a relative path from the README.md file hierarchy.

timoreimann (approved):

PR #676, opened 2024-01-16
feat(CON-7851): Register LB admission webhook

prichrd:
This change introduces an admission webhook validating load balancer services. The endpoint builds the same request that CCM would, and queries DO API with a validation flag to check for potential configuration issues.

Some changes to the Makefile also allow the new binary `digitalocean-cloud-controller-manager-admission-server` to be built and packaged in a docker image.


cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/main.go
prichrd, line 1: I decided to rename the binary to `digitalocean-cloud-controller-manager-admission-server`. 
It is a very long name, but it adds a bit more context than `digitalocean-webhook-server`.  
We need to publish a docker image for it, so it needs to have a distinctive name.

I'm open to suggestions if someone has a better name idea!

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/main.go
```diff
@@ -0,0 +1,122 @@
...
+func (t *tokenSource) Token() (*oauth2.Token, error) {
+	token := &oauth2.Token{
+		AccessToken: t.AccessToken,
+	}
+	return token, nil
+}
+
+func main() {
+	if err := startAdmissionServer(); err != nil {
+		fmt.Fprintf(os.Stderr, "failed to start admission server: %v\n", err)
```
timoreimann, line 59: I'd use `%s` to make sure we're always getting a string representation.

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/main.go
```diff
@@ -0,0 +1,122 @@
...
+	server := webhook.NewServer(webhook.Options{})
+
+	godoClient, err := getGodoClient()
+	if err != nil {
+		return fmt.Errorf("getting godo client: %w", err)
+	}
+
+	lbAdmissionHandler := do.NewLBServiceAdmissionHandler(&ll, godoClient)
+	if err := lbAdmissionHandler.WithRegion(); err != nil {
+		return fmt.Errorf("injecting region into LB validator: %w", err)
```
timoreimann, line 76: Appreciate if we could apply two small adjustments to the error we produce here and elsewhere in this PR:

- Start off the message with `failed to [...]` for consistency and readability reasons
- Change `%w` to `%s` to avoid extending the API surface error -- even though it wouldn't make a difference now, it could do so through a future refactoring (for a rationale, see [this section](https://go.dev/blog/go1.13-errors#whether-to-wrap) from an official Go blog post)

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/main.go
```diff
@@ -0,0 +1,122 @@
...
+	if vpcID == "" {
+		return fmt.Errorf("environment variable %q is required", doClusterVPCIDEnv)
+	}
+	lbAdmissionHandler.WithVPCID(vpcID)
+
+	ll.Info("registering admission handlers")
+	server.Register("/lb-service", &webhook.Admission{Handler: lbAdmissionHandler})
+
+	if err = server.Start(signals.SetupSignalHandler()); err != nil {
+		return err
```
timoreimann, line 95: Let's also annotate/wrap the error with a descriptive message via `fmt.Errorf`.

cloud-controller-manager/do/lb_service_admission_handler.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+}
+
+// Handle handles admissions requests for load balancer services.
+func (h *LBServiceAdmissionHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	resp := h.handle(ctx, req)
+
+	logFields := []any{"object_name", req.Name, "object_namespace", req.Namespace, "object_kind", req.Kind.String()}
+	if resp.Allowed {
+		h.log.Info("allowing admission request", logFields...)
```
timoreimann, line 58: I wonder if we should use `Debug` here given busy clusters could be receiving a large number of requests should nodes come and up frequently (e.g., with autoscaling clusters), which in turn could "spam" the logs. 🤔 No super strong opinions though. Any thoughts?

cloud-controller-manager/do/lb_service_admission_handler.go
```diff
@@ -0,0 +1,166 @@
...
+
+func (h *LBServiceAdmissionHandler) handle(ctx context.Context, req admission.Request) admission.Response {
+	var svc corev1.Service
+	err := h.decoder.Decode(req, &svc)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("decoding admission request: %w", err))
+	}
+
+	if svc.Spec.Type != corev1.ServiceTypeLoadBalancer {
+		return admission.Allowed("allowing service as it is not a load balancer")
```
timoreimann, line 74: nit: maybe go with slightly less fancy English here (which I might have suggested in the first place during a Shakespeare-y moment of mine):
```suggestion
		return admission.Allowed("allowing service that is not of type load balancer")
```

cloud-controller-manager/do/lb_service_admission_handler.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	if svc.DeletionTimestamp != nil {
+		return admission.Allowed("allowing service as it is being deleted")
```
timoreimann, line 78: Similar:

```suggestion
		return admission.Allowed("allowing service that is being deleted")
```

cloud-controller-manager/do/lb_service_admission_handler.go
```diff
@@ -0,0 +1,166 @@
...
+	case err == nil:
+		return admission.Allowed("valid load balancer definition")
+	case resp == nil: // This case should never happen, asserting to prevent nil pointer exceptions.
+		return admission.Errored(http.StatusInternalServerError, errors.New("expected a DO API response"))
+	case resp.StatusCode < 400:
+		return admission.Allowed("valid load balancer definition")
+	case resp.StatusCode < 500:
+		return admission.Denied(fmt.Sprintf("invalid load balancer definition: %s", err))
+	default:
+		return admission.Allowed(fmt.Sprintf("received unexpected status code (%d) from DO API, allowing to prevent blocking: %s", resp.StatusCode, err))
```
timoreimann, line 144: As discussed via DM, shouldn't it be okay here to return `admission.Errored` or `admission.Denied` and still get the request allowed if the failure policy is set to `Ignore` (and likewise get it rejected should users choose to go with `Fail`)?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -591,32 +585,48 @@ func (l *loadBalancers) buildLoadBalancerRequest(ctx context.Context, service *v
...
 		Firewall:                     buildFirewall(service),
 	}, nil
 }
 
+// buildLoadBalancerRequest returns a *godo.LoadBalancerRequest to balance
+// requests for service across nodes.
+func (l *loadBalancers) buildLoadBalancerRequest(ctx context.Context, service *v1.Service, nodes []*v1.Node) (*godo.LoadBalancerRequest, error) {
+	req, err := buildLoadBalancerRequest(service)
+	if err != nil {
+		return nil, err
```
timoreimann, line 610: Let's wrap/annotate the errors from this method via `fmt.Errorf()` to ease distinguishing the possible error paths.

timoreimann (commented):
This is looking pretty good. 👍 Left a few small-ish suggestions.

cloud-controller-manager/do/lb_service_admission_handler.go
prichrd: Unfortunately, returning `admission.Errored` or `admission.Denied` isn't ignored when we set the failure policy to `Ignore`. I dug through the [webhook admission code](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/validating/dispatcher.go#L170) to find the true meaning of the failure policy. It will ignore client errors such as: the webhook isn't responsive, etc.

For those reasons, I think the best solution is to:
- Configure the failure policy to `Ignore` to catch any potential failures when getting an answer from the webhook
- Have the webhook `Allow` the requests when faced with a 5xx (transient failure)
- Have the webhook `Deny` the requests when faced with a 4xx (non-transient failures)
- Have the webhook `Error` for unexpected errors within the whole logic.

cloud-controller-manager/do/lb_service_admission_handler.go
prichrd: I initially added that log message because I found it interesting to see the passed vs failed requests, but I agree it can become very noisy. I'm going to set it to `Debug` and work on a subsequent PR for metrics reporting. Reporting `Allowed`/`Denied`/`Errored` admissions per endpoint should do the trick.

timoreimann: It's worth double-checking what kind of metrics may already be produced by controller-runtime. Best case, we don't have to do anything in this regard, but I'm not certain. We'll have to check.

prichrd: I looked at what metrics were collected yesterday, and I was a bit surprised that webhook-specific metrics didn't seem to report status: [pkg/webhook/internal/metrics/metrics.go](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/internal/metrics/metrics.go). 

It will require a bit of work, considering that metrics aren't served by default. I have a beginning of something stashed.

cloud-controller-manager/do/loadbalancers.go
prichrd: I might just ignore this wrapping as this is in the CCM LoadBalancer interface implementation. I don't want to add a layer of verbosity in errors there if it's not required.

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Copyright 2023 DigitalOcean
```
timoreimann, line 1: We need to update the year at this point (other files as well).

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/main.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	lbAdmissionHandler := do.NewLBServiceAdmissionHandler(&ll, godoClient)
+	if err := lbAdmissionHandler.WithRegion(); err != nil {
+		return fmt.Errorf("failed to inject region into lb service admission handler: %w", err)
+	}
+
+	clusterID := os.Getenv(doClusterIDEnv)
+	if clusterID == "" {
+		return fmt.Errorf("environment variable %q is required", doClusterIDEnv)
```
timoreimann, line 75: I don't think this is a required parameter, so we shouldn't error out if it's not set.

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/main.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	lbAdmissionHandler.WithClusterID(clusterID)
+
+	vpcID := os.Getenv(doClusterVPCIDEnv)
+	if vpcID == "" {
+		return fmt.Errorf("environment variable %q is required", doClusterVPCIDEnv)
```
timoreimann, line 81: The VPC ID is for sure needed when LBs are involved, but otherwise may not be required. My quick check tells me we don't error hard on the omission in core CCM, but we should double-check.

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/main.go
```diff
@@ -0,0 +1,127 @@
...
+	if err = server.Start(signals.SetupSignalHandler()); err != nil {
+		return err
+	}
+	ll.Info("Webhooks server started")
+
+	return nil
+}
+
+func getGodoClient() (*godo.Client, error) {
+	token := os.Getenv(doAccessTokenEnv)
```
timoreimann, line 97: Env vars should ideally only be read/parsed at the top-most (main) function, and the values should be passed into functions.

We might have just refactored this from elsewhere though, in which case I guess we could keep it the way it is for now.

cloud-controller-manager/do/lb_service_admission_handler.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	return resp
+}
+
+func (h *LBServiceAdmissionHandler) handle(ctx context.Context, req admission.Request) admission.Response {
+	var svc corev1.Service
+	err := h.decoder.Decode(req, &svc)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed to decode admission request: %w", err))
```
timoreimann, line 71: Please use `%s` instead of `%w` here (and wherever else it's used) for reasons mentioned before (tl;dr: no unnecessary API surface expansion).

cloud-controller-manager/do/lb_service_admission_handler.go
```diff
@@ -0,0 +1,171 @@
...
+	lbReq.VPCUUID = h.vpcID
+	lbReq.Tags = []string{buildK8sTag(h.clusterID)}
+	return lbReq, nil
+}
+
+func (h *LBServiceAdmissionHandler) mapGodoRespToAdmissionResp(resp *godo.Response, err error) admission.Response {
+	switch {
+	case err == nil:
+		return admission.Allowed("valid load balancer definition")
+	case resp == nil: // This case should never happen, asserting to prevent nil pointer exceptions.
```
timoreimann, line 142: This can legitimately happen when a networking-related error occurred (e.g., I/O problem). I think we want to include `err` in that case.

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/Dockerfile
prichrd: Will be addressed in a separaete PR.

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/main.go
prichrd: I'll address when I'll work on metrics serving.

PR #683, opened 2024-01-30
feat(CON-8768): add LB admission webhook manifests

prichrd:
This change introduces the k8s manifest changes required to deploy the LB admission webhook alongside the existing CCM deployment. Adds:

- Validation webhook definition
- Self signed issuer for signing the admission server certs
- Serving certs for the admission server
- Service fronting the CCM pod
- Additional container running the admission server within the CCM pod.

### DIFF
--- a/releases/dev.yml
+++ b/releases/dev.yml
@@ -50,7 +50,33 @@ spec:
               secretKeyRef:
                 name: digitalocean
                 key: access-token
-
+      - image: digitalocean/digitalocean-cloud-controller-manager-admission-server:dev
+        name: digitalocean-cloud-controller-manager-admission-server
+        command:
+          - "/bin/digitalocean-cloud-controller-manager-admission-server"
+        resources:
+          requests:
+            cpu: 100m
+            memory: 50Mi
+        env:
+          - name: DO_ACCESS_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: digitalocean
+                key: access-token
+        ports:
+        - containerPort: 9443
+          name: admission
+          protocol: TCP
+        volumeMounts:
+        - mountPath: /tmp/k8s-webhook-server/serving-certs
+          name: digitalocean-cloud-controller-manager-serving-certs
+          readOnly: true
+      volumes:
+      - name: digitalocean-cloud-controller-manager-serving-certs
+        secret:
+          defaultMode: 420
+          secretName: digitalocean-cloud-controller-manager-serving-certs
 ---
 apiVersion: v1
 kind: ServiceAccount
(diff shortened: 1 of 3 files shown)


releases/dev.yml
```diff
@@ -50,7 +50,33 @@ spec:
               secretKeyRef:
                 name: digitalocean
                 key: access-token
-
+      - image: digitalocean/digitalocean-cloud-controller-manager-admission-server:dev
```
varshavaradarajan, line 53: Please update the bump-version make target to update the tag for release.

releases/dev.yml
```diff
@@ -152,3 +178,68 @@ subjects:
...
+spec:
+  selector:
+    app: digitalocean-cloud-controller-manager
+  ports:
+    - protocol: TCP
+      port: 443
+      targetPort: 9443
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
```
varshavaradarajan, line 196: please update the README to have a prereq to install cert-manager

releases/dev.yml
```diff
@@ -152,3 +178,68 @@ subjects:
...
+webhooks:
+- name: validation-webhook.loadbalancer.doks.io
+  admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      namespace: "kube-system"
+      name: "digitalocean-cloud-controller-manager"
+      path: "/lb-service"
+  failurePolicy: Fail
```
varshavaradarajan, line 233: I think this should be `Ignore` 

k8s docs: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy

If there's a problem calling the webhook server, we should allow the user to proceed to create the service, right? 
If there's no error calling the webhook server, but the webhook server returns Denied, not really sure what happens, can you please check?

releases/dev.yml
```diff
@@ -152,3 +178,68 @@ subjects:
...
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - services
+    scope: '*'
```
varshavaradarajan, line 244: scope should be Namespaced.

varshavaradarajan (changes requested):

releases/dev.yml
prichrd: This is an artifact of testing; I forgot to set it back to `Ignore`. 

@timoreimann and I have dug into the failure policies, and we definitely want to set it to `Ignore`. The `Ignore` failure policy will only ensure that transient failures and poor webhook configurations will not result in a denial of the request. We have also discovered that the `Ignore` policy treats an `admission.Errored` the same way it treats an `admission.Denied`: it denies it.

varshavaradarajan (approved):

PR #689, opened 2024-02-06
Document admission server and make it optional

prichrd:
The admission webhook is missing a lot of docs and context. Aside from the code artifacts, it isn't clear what it does and why. It was also concluded with @timoreimann that the admission server should be optional.

This change:
* Adds documentation in the `README.md`, and in the `docs/admission-server.md`. These documents should address the *what*, the *why* and the *how*.
* Splits the release manifests of the `digitalocean-cloud-controller-manager` and `digitalocean-cloud-controller-manager-admission-server` in two separate directories.

This PR should also help address some of the concerned expressed in this discussion: https://github.com/digitalocean/digitalocean-cloud-controller-manager/discussions/688


releases/digitalocean-cloud-controller-manager-admission-server/dev.yml
```diff
@@ -0,0 +1,109 @@
...
+  selfSigned: {}
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: kube-system/digitalocean-cloud-controller-manager-admission-server-serving-certs
+  name: digitalocean-cloud-controller-manager-admission-webhook
+webhooks:
+- name: validation-webhook.loadbalancer.do.io
```
timoreimann, line 89: Let's go with `validation-webhook.cloud-controller-manager.digitalocean.com` here.

PR #712, opened 2024-04-24
go.mod: update go to 1.22

dhij:


dhij:
@timoreimann @llDrLove -

Please note I added a bunch of replace directives in `go.mod` because of interdependencies between different packages. 

For example, when `controller-runtime` was updated to `0.30.0` along with `go mod tidy`, it gave me the [undefined: leaderelection.SwitchMetric error in the CI](https://github.com/digitalocean/digitalocean-cloud-controller-manager/actions/runs/8819117066/job/24209549064#step:4:155). The `leaderelection.SwitchMetric` is from the `k8s.io/client-go/tools/leaderelection` package. So I added a replace directive `k8s.io/client-go v0.30.0 => k8s.io/client-go v0.29.0`. But then that led to another error from the `k8s.io/apiserver` complaining that now it had another undefined method, and so on. That led to downgrading multiple packages and dependencies.

That might not be the best approach but please let me know if you have any other suggestions 👍

Makefile
```diff
@@ -35,7 +35,7 @@ all: test
 
 publish: clean build push
 
-ci: check-headers check-unused gofmt govet test
+ci: check-headers gofmt govet test
```
timoreimann, line 38: We may want to keep this to protect against missed vendoring.

.github/workflows/test.yml
```diff
@@ -19,9 +19,12 @@ jobs:
         with:
           go-version-file: go.mod
 
-      - name: Run tests
+      - name: Check headers and go mod
```
timoreimann, line 22: The `ci` target includes `test` again. Should we move all dependent targets except for `test` into a new target that we call from here?

Alternatively, we could retain just the old structure. I don't have strong opinions on either.

dhij: I removed the `test` from the `ci` target 🙂 https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/712/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R38

but yes i agree we could just retain the old structure

dhij:
@timoreimann @llDrLove Oh I just realized maybe this is the part where I might have to hold off and wait for our dependencies to update their dependencies instead of using the replace directives as I describe above 🙂

.github/workflows/test.yml
timoreimann: Sorry missed that, thanks!

dhij:
@timoreimann @llDrLove we still need 3 replace directives but now thats much less than what we had before and all the `k8s.io/*` packages are up-to-date at `v0.30.0` so i think this should be good 👍

timoreimann (approved):

Issue #573, opened 2023-01-23
do-loadbalancer should accept a certificate name as an alternative to the certificate ID


dverbeir:
When using a CD system such as Flux to deploy an application exposed through a LoadBalancer service with a pre-provisioned certificate, if the certificate was renewed and its certificate ID consequently changed, a later re-deployment of the service by the CD system will be stuck with an external IP in pending state because the `service.beta.kubernetes.io/do-loadbalancer-certificate-id` annotation will be set to an earlier ID of the otherwise unmodified certificate.

To fix the deployment (after figuring out what is wrong...), the source repository used to drive the CD system must be updated. This is very undesirable and inconvenient as nothing really changed in the (intent of the) deployment specification.

I think that one possible and fairly simple solution would be to allow specifying the certificate name at time of deployment of the service. The name could then be converted to an ID by the CCM (using existing DO certificates API). The remainder and process could remain unchanged, with automatic updates of the certificate reflected in its ID.

I created PR https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/572 proposing this approach.
It uses the 'service.beta.kubernetes.io/do-loadbalancer-certificate-name' service annotation to specify the certificate by name.

timoreimann:
This was addressed by the linked PR.

PR #722, opened 2024-05-15
Add new annotation for using certificate name when configuring load balancer TLS

bbassingthwaite:
This will improve the users experience when using a certificate. A name is easier to identify then a UUID. If they are using lets encrypt, the underlying UUID will change on each renewal which causes headaches for users. Certificate names are also now unique so we can identify a certificate by its name even if the underlying UUID has changed.

The existing certificate ID annotation will take precedence. If they specify a name, we will use the certificate API to look it up and get the certificate ID.

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -547,7 +547,7 @@ func buildLoadBalancerRequest(ctx context.Context, service *v1.Service, godoClie
 			return nil, err
 		}
 	} else {
-		forwardingRules, err = buildForwardingRules(service, godoClient)
+		forwardingRules, err = buildForwardingRules(ctx, service, godoClient)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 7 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -1084,12 +1094,37 @@ func getStrings(service *v1.Service, anno string) []string {
...
+	if certificateID != "" {
+		return certificateID, nil
+	}
+	certificateName := getCertificateName(service)
+	if certificateName == "" {
+		return "", nil
+	}
+	lbCert, _, err := godoClient.Certificates.ListByName(ctx, certificateName, &godo.ListOptions{Page: 1, PerPage: 1})
+	if err != nil {
+		return "", fmt.Errorf("failed to get certificate by name: %s error: %w", certificateName, err)
```
timoreimann, line 1120:
```suggestion
		return "", fmt.Errorf("failed to get certificate by name %q: %s", certificateName, err)
```

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -1084,12 +1094,37 @@ func getStrings(service *v1.Service, anno string) []string {
...
+	}
+	if len(lbCert) == 0 {
+		return "", fmt.Errorf("certificate with name: %s not found", certificateName)
```
timoreimann, line 1123:
```suggestion
		return "", fmt.Errorf("certificate with name %q not found", certificateName)
```

timoreimann (commented):
Overall, this looks like a good addition to me.

Could you please also update the annotation documentation and add a change log entry?

timoreimann (approved):

timoreimann:
Fixes #573.

PR #731, opened 2024-05-29
Add support for loadBalancerSourceRanges

gottwald:
This change introduces support for loadBalancerSourceRanges which is the preferred and recommended way to add firewall allow rules and therefore takes presedence over annotation based allow rules.

Thanks to @J0sh0nat0r for submitting the original pull request #584.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## unreleased
 
+* Adding support for specifiying `loadBalancerSourceRanges` in the service spec. Source ranges take precedence over annotation based allow rules (`service.beta.kubernetes.io/do-loadbalancer-allow-rules`).
+
 ## v0.1.51 (beta) - May 28, 2024
 
 * Adjusts load balancer health check behaviour to probe Kubernetes components correctly, ensuring that LB traffic stops
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/cloud-provider/api"
 	servicehelpers "k8s.io/cloud-provider/service/helpers"
 	"k8s.io/klog/v2"
+	utilnet "k8s.io/utils/net"
 
 	"github.com/digitalocean/godo"
 )
@@ -604,6 +605,11 @@ func buildLoadBalancerRequest(ctx context.Context, service *v1.Service, godoClie
 		return nil, err
 	}
 
+	fw, err := buildFirewall(service)
+	if err != nil {
+		return nil, err
+	}
+
 	return &godo.LoadBalancerRequest{
 		Name:                         lbName,
 		SizeSlug:                     sizeSlug,
@@ -617,7 +623,7 @@ func buildLoadBalancerRequest(ctx context.Context, service *v1.Service, godoClie
 		EnableBackendKeepalive:       enableBackendKeepalive,
 		DisableLetsEncryptDNSRecords: &disableLetsEncryptDNSRecords,
 		HTTPIdleTimeoutSeconds:       httpIdleTimeoutSeconds,
-		Firewall:                     buildFirewall(service),
+		Firewall:                     fw,
 		Type:                         lbType,
 	}, nil
 }
@@ -866,9 +872,12 @@ func buildForwardingRule(service *v1.Service, port *v1.ServicePort, protocol, ce
 	return &forwardingRule, nil
 }
 
-func buildFirewall(service *v1.Service) *godo.LBFirewall {
+func buildFirewall(service *v1.Service) (*godo.LBFirewall, error) {
 	var allow []string
-	sourceRanges := getSourceRangeRules(service)
+	sourceRanges, err := getSourceRangeRules(service)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get source range rules: %s", err)
+	}
 	if len(sourceRanges) > 0 {
 		// source ranges should take precedence over annotation based allow rules
 		allow = sourceRanges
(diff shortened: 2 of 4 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -866,12 +866,32 @@ func buildForwardingRule(service *v1.Service, port *v1.ServicePort, protocol, ce
...
 	}
 }
 
+// getSourceRangeRules returns loadBalancerSourceRanges in the format
+// `cidr:{source}`
+func getSourceRangeRules(service *v1.Service) []string {
+	sourceRanges := service.Spec.LoadBalancerSourceRanges
+
+	for i, sourceRange := range sourceRanges {
+		sourceRanges[i] = "cidr:" + sourceRange
```
timoreimann, line 889: Curious, is the source range format standardized / specified anyhow? The [API documentation](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec) seems scarce on this point. IIUC, we're assuming `<v.w.x.y/z>`?

Should we validate the entries here so that we don't need an extra roundtrip to the LB API to validate?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ allow LBs to revert to the old health check behaviour. The annotation and old he
   needs to be manually updated when using Let's Encrypt certificates. This is due to the certificate ID updating each time the
   certificate is rotated. If both annotations are provided, then `service.beta.kubernetes.io/do-loadbalancer-certificate-id` takes
   precedence.
+* Adding support for specifiying `loadBalancerSourceRanges` in the service spec. Source ranges take precedence over annotation based allow rules (`service.beta.kubernetes.io/do-loadbalancer-allow-rules`).
```
timoreimann, line 15: The change log items right above have been released already as v0.1.51. I'm rather confused how the new entry added here does not lead to a conflict from [the current master change log](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/0d3127f065dc0d25b88f3836c5029c2dfbc03507/CHANGELOG.md). 🤔 

The source range announcement should be the only unreleased change we have by now.

timoreimann (commented):
Please also update https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/0d3127f065dc0d25b88f3836c5029c2dfbc03507/docs/controllers/services/annotations.md#servicebetakubernetesiodo-loadbalancer-allow-rules about the new precedence ordering.

cloud-controller-manager/do/loadbalancers.go
gottwald: I haven't found anything that would document a certain format. However if you follow the code from [this](https://github.com/kubernetes/kubernetes/blob/8565e375251450a291f0af3c6195c7a5bf890292/pkg/api/service/util.go#L44) then you'll see that the parsing function only allows the '/' CIDR notation, which is what I would expect give the field name.
We could maybe think about re-using the function which would make sure it parses like upstream wants it to?
There's the same function in the [cloud-provider package](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/4e844752045f0fbe4925723142c8ae6bfa949d92/vendor/k8s.io/cloud-provider/service/helpers/helper.go#L60-L60) which is already imported.

CHANGELOG.md
gottwald: I had an older state of master that I branched, that's why in mine this was still unreleased.
There's no conflict because git doesn't care about the headline movement and the line itself doesn't conflict.
I'll rebase on the current master and will move this one around.

cloud-controller-manager/do/loadbalancers.go
gottwald: Yes, lets do that. I also don't want to adopt yet another annotation.

docs/controllers/services/annotations.md
```diff
@@ -223,4 +223,6 @@ Specifies the comma seperated ALLOW firewall rules for the load-balancer
 
 **Note**
 
-Rules must be in the format `{type}:{source}` (ex. `ip:1.2.3.4,cidr:2.3.0.0/16`)
+Rules must be in the format `{type}:{source}` (ex. `ip:1.2.3.4,cidr:2.3.0.0/16`).
+
+These rules will be ignored if `LoadBalancerSourceRanges` is set, which is the prefered way to enter allow rules.
```
timoreimann, line 228: grammar: s/prefered/preferred/

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -866,11 +872,41 @@ func buildForwardingRule(service *v1.Service, port *v1.ServicePort, protocol, ce
...
+// getSourceRangeRules returns loadBalancerSourceRanges in the format
+// `cidr:{source}`
+func getSourceRangeRules(service *v1.Service) ([]string, error) {
+	if len(service.Spec.LoadBalancerSourceRanges) == 0 {
+		return nil, nil
+	}
+	specs := service.Spec.LoadBalancerSourceRanges
+	ipnets, err := utilnet.ParseIPNets(specs...)
+	if err != nil {
+		return nil, fmt.Errorf("service.Spec.LoadBalancerSourceRanges: %v is not valid. Expecting a list of IP ranges. For example, 10.0.0.0/24. Error msg: %v", specs, err)
```
timoreimann, line 902: This [can be stuttering](https://go.dev/play/p/p8c7lDjyRT9) since `net.ParseCIDRs` (which `ParseIPNets` eventually calls) may already be talking about addresses being invalid.

Can we address that and also be more concise, e.g.:

```golang
fmt.Errorf("failed to parse service.Spec.LoadBalancerSourceRanges %q: %s (expected format for an IP address is 10.0.0.0/24)", specs, err)
```

docs/controllers/services/annotations.md
gottwald: done and some more..

timoreimann (approved):

PR #734, opened 2024-06-04
Add support for internal load balancers

bbassingthwaite:
- Currently closed alpha and will document annotations when we open it up to beta
- When network=INTERNAL, the IP of the load balancer will be from the underlying VPC. This IP will only be accessible from within the VPC.

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -541,6 +541,10 @@ func buildLoadBalancerRequest(ctx context.Context, service *v1.Service, godoClie
 	if err != nil {
 		return nil, err
 	}
+	lbNetwork, err := getNetwork(service)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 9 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -1388,6 +1393,17 @@ func getType(service *v1.Service) (string, error) {
 	return name, nil
 }
 
+func getNetwork(service *v1.Service) (string, error) {
+	network, ok := service.Annotations[annDONetwork]
+	if !ok || network == "" {
```
timoreimann, line 1398: nit: could be simplified to

```golang
if network == "" {
```

since the default value matches the absent one.

(I realize we have this redundant in a few spots in CCM.)

timoreimann (commented):
LGTM, subject to us possibly wanting to QA it pre-merge.

timoreimann (approved):

gottwald (approved):

PR #738, opened 2024-06-11
Disable automatic health check configuration on Proxy Protocol

timoreimann:
The default health check endpoints do not expect a Proxy Protocol header.


timoreimann:
Closing in favor of #739.

PR #748, opened 2024-07-26
Add a firewall rule to open the LB health check and service port(s) when using `service.beta.kubernetes.io/do-loadbalancer-type=REGIONAL_NETWORK`

bbassingthwaite:
Since NLB will route traffic over the public interface for both the LB data plane and health checks, we will need to open the appropriate health check port. This change only affects LBs who set the annotation `service.beta.kubernetes.io/do-loadbalancer-type=REGIONAL_NETWORK` and the network is `EXTERNAL` (default).

### DIFF
--- a/cloud-controller-manager/do/firewall_controller.go
+++ b/cloud-controller-manager/do/firewall_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"sort"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cloud-controller-manager/do/firewall_controller.go
```diff
@@ -312,14 +313,39 @@ func (fm *firewallManager) createReconciledFirewallRequest(serviceList []*v1.Ser
 					},
 				)
 			}
+		} else if svc.Spec.Type == v1.ServiceTypeLoadBalancer {
+			lbType, err := getType(svc)
+			if err != nil {
+				return nil, err
```
timoreimann, line 319: Can we add `fmt.Errorf()` around `err` to provide context, specifically the name of the Service affected?

Similar for the error returned in line 323.

cloud-controller-manager/do/firewall_controller_test.go
```diff
@@ -638,7 +705,11 @@ func TestFirewallController_createReconciledFirewallRequest(t *testing.T) {
 				workerFirewallTags: testWorkerFWTags,
 				workerFirewallName: testWorkerFWName,
 			}
-			fwReq := fm.createReconciledFirewallRequest(test.serviceList)
+			fwReq, err := fm.createReconciledFirewallRequest(test.serviceList)
+
+			if (err != nil && test.expectedError == nil) || (err == nil && test.expectedError != nil) {
+				t.Errorf("incorrect firewall config\nwant: %#v\n got: %#v", test.expectedError, err)
```
timoreimann, line 711: This should probably say

```golang
t.Fatalf("expected error %q, got %q", test.expectedError, err)
```

timoreimann (dismissed):

cloud-controller-manager/do/firewall_controller.go
```diff
@@ -271,8 +271,9 @@ func (fm *firewallManager) updateFirewall(ctx context.Context, fwID string, fr *
 }
 
 // createReconciledFirewallRequest creates a firewall request that has the correct rules, name and tag
-func (fm *firewallManager) createReconciledFirewallRequest(serviceList []*v1.Service) *godo.FirewallRequest {
+func (fm *firewallManager) createReconciledFirewallRequest(serviceList []*v1.Service) (*godo.FirewallRequest, error) {
 	var nodePortInboundRules []godo.InboundRule
+	healthCheckPorts := make(map[int]struct{})
 	for _, svc := range serviceList {
 		managed, err := isManaged(svc)
```
timoreimann, line 278: We may need to scope the non-management restriction down to the NodePort logic while continuing to reconcile ports for NLBs.

Withdrawing approval for now so that we don't forget.

timoreimann (approved):

Issue #378, opened 2021-06-04
do-loadbalancer-protocol: 'http2' results in 'http2' --> 'http'


evilaugust:
When setting the load balancer protocol, the droplet protocol defaults to HTTP instead of HTTP2. Is there an annotation to set the droplet protocol?

`apiVersion: v1
kind: Service
metadata:
  annotations:
    service.beta.kubernetes.io/do-loadbalancer-enable-proxy-protocol: 'false'
    service.beta.kubernetes.io/do-loadbalancer-name: 'ingress-lb-nyc1-01'
    service.beta.kubernetes.io/do-loadbalancer-protocol: 'http2'
    service.beta.kubernetes.io/do-loadbalancer-certificate-id: 'e22edb46-3d40-4b4b-a608-bd741da45562'
    service.beta.kubernetes.io/do-loadbalancer-algorithm: 'least_connections'
    service.beta.kubernetes.io/do-loadbalancer-size-slug: 'lb-small'
    service.beta.kubernetes.io/do-loadbalancer-http2-ports: '5000'
  labels:
    app.kubernetes.io/name: ingress-nginx
    app.kubernetes.io/instance: ingress-nginx
    app.kubernetes.io/component: controller
  name: ingress-nginx-controller
  namespace: ingress-nginx
spec:
  type: LoadBalancer
  externalTrafficPolicy: Cluster
  ports:
    - name: grpc
      port: 5000
      protocol: TCP
      targetPort: 5000
  selector:
    app.kubernetes.io/name: ingress-nginx
    app.kubernetes.io/instance: ingress-nginx
    app.kubernetes.io/component: controller`

timoreimann:
Hi @evilaugust 👋 

IIUC, you'd like to set different target and entry protocols. This is similar to #367 but unfortunately, right now it is not supported. I'm happy to keep the ticket open to track support.

evilaugust:
@timoreimann Thank you for your response. Actually, I want the target and entry to be the same. They should both be HTTP2, but when I set service.beta.kubernetes.io/do-loadbalancer-protocol: 'http2', the target defaults to HTTP while the entry protocol gets set to HTTP2, which is essentially pointless. I just need them both to be HTTP2. Is that possible?

evilaugust:
[...]

PR #751, opened 2024-08-06
Support explicit target protocol annotation for load balancers

### LINKED ISSUE #367: Cannot get an HTTPS -> HTTPS LB setup
I am trying to setup a basic Load-Balancer using kubernetes. I tried everything and looked at every configuration and even in the source code, but I cannot figure out how to achieve the configuration I want.

What I cant to achieve is this:

```
HTTP on port 80 -> HTTP on port 80
HTTPS on port 443 -> HTTPS on port 443
```
Note that this is not a passthrough. The load-Balancer unencrypts the request using the certificate in my account, then re-sends it via HTTPS to my pod using a self-signed certificate.

When I set this up manually via the interface, it works well. However it does not seem to be possible using Kubernetes.  The kubernetes service always create this setup:

```
HTTP on port 80 -> HTTP on port 80
HTTPS on port 443 -> HTTP on port 443
```

So it tries to do HTTP on my port 443 and it does not work (note: In this example I redacted the real target port so it’s easier to understand).

This seems to be because of this part of the code: https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/e4fb291b26487b16367af97b1e625ddc982caae7/cloud-controller-manager/do/loadbalancers.go#L862

Here is my current Kubernetes service for reference:

```
apiVersion: v1
kind: Service
metadata:
  name: server
  annotations:
      service.beta.kubernetes.io/do-loadbalancer-name: "kanjimi"
      service.beta.kubernetes.io/do-loadbalancer-size-slug: "lb-small"
      service.beta.kubernetes.io/do-loadbalancer-protocol: "http"
      service.beta.kubernetes.io/do-loadbalancer-http-ports: "80"
      service.beta.kubernetes.io/do-loadbalancer-tls-ports: "443"
      service.beta.kubernetes.io/do-loadbalancer-tls-passthrough: "true"
      service.beta.kubernetes.io/do-loadbalancer-redirect-http-to-https: "true"
      service.beta.kubernetes.io/do-loadbalancer-enable-proxy-protocol: "true"
      service.beta.kubernetes.io/do-loadbalancer-certificate-id: "xxx"
      service.beta.kubernetes.io/do-loadbalancer-algorithm: "least_connections"
      service.beta.kubernetes.io/do-loadbalancer-healthcheck-port: "443"
      service.beta.kubernetes.io/do-loadbalancer-healthcheck-protocol: "https"
      service.beta.kubernetes.io/do-loadbalancer-healthcheck-path: "/api/health-check"
      service.beta.kubernetes.io/do-loadbalancer-healthcheck-check-interval-seconds: "5"
      service.beta.kubernetes.io/do-loadbalancer-healthcheck-response-timeout-seconds: "5"
      service.beta.kubernetes.io/do-loadbalancer-healthcheck-unhealthy-threshold: "3"
      service.beta.kubernetes.io/do-loadbalancer-healthcheck-healthy-threshold: "2"
spec:
  type: LoadBalancer
  selector:
    app: server
  ports:
    - name: http
      protocol: TCP
      port: 80
      targetPort: 80
    - name: https
      protocol: TCP
      port: 443
      targetPort: 443
```

asaha2:
- introduce `service.beta.kubernetes.io/do-loadbalancer-target-protocol` annotation that provides explicit specifier for forwarding rules target protocols, if set correctly overrides the implicit default of `http` target protocol
- default behavior in absence of this new annotation is unchanged

Fixes #367 #378

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,14 @@
 ## unreleased
 
+* When using the LoadBalancer `service.beta.kubernetes.io/do-loadbalancer-target-protocol` annotation with any of `http`, 
+  `https`, `http2` or `http3` specification, the forwarding target protocol is overridden from the implicit default `http`
+  protocol.
+
 ## v0.1.55 (beta) - July 29, 2024
 
 * When using the LoadBalancer `service.beta.kubernetes.io/do-loadbalancer-type=REGIONAL_NETWORK` (under closed beta), firewall rules
-are added to open up the underlying health check port and all the defined (port, protocols) defined on the service. This is to
-permit traffic to arrive directly on the underlying worker nodes.
+  are added to open up the underlying health check port and all the defined (port, protocols) defined on the service. This is to
+  permit traffic to arrive directly on the underlying worker nodes.
 
 ## v0.1.54 (beta) - June 12, 2024
 
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -733,13 +733,24 @@ func buildHTTP3ForwardingRule(ctx context.Context, service *v1.Service, godoClie
 		return nil, errors.New("certificate ID is required for HTTP3")
 	}
 
+	var targetProtocol string
+	if _, err := getProtocol(service, annDOTargetProtocol, func(protocol string) { targetProtocol = protocol }); err != nil {
+		return nil, err
+	}
+	if targetProtocol == protocolTCP || targetProtocol == protocolUDP {
+		return nil, fmt.Errorf("target protocol annotation is not supported for TCP or UDP")
+	}
+	if targetProtocol == "" {
+		targetProtocol = protocolHTTP
+	}
+
 	for _, port := range service.Spec.Ports {
 		if port.Port == int32(http3Port) {
 			return &godo.ForwardingRule{
 				EntryProtocol:  protocolHTTP3,
 				EntryPort:      http3Port,
 				CertificateID:  certificateID,
-				TargetProtocol: protocolHTTP,
+				TargetProtocol: targetProtocol,
 				TargetPort:     int(port.NodePort),
 			}, nil
 		}
@@ -753,7 +764,7 @@ func buildHTTP3ForwardingRule(ctx context.Context, service *v1.Service, godoClie
 func buildForwardingRules(ctx context.Context, service *v1.Service, godoClient *godo.Client) ([]godo.ForwardingRule, error) {
 	var forwardingRules []godo.ForwardingRule
 
-	defaultProtocol, err := getProtocol(service)
+	defaultProtocol, err := getProtocol(service, annDOProtocol)
 	if err != nil {
 		return nil, err
 	}
@@ -928,6 +939,14 @@ func buildTLSForwardingRule(forwardingRule *godo.ForwardingRule, service *v1.Ser
 		return errors.New("either certificate id should be set or tls pass through enabled, not both")
 	}
 
+	var targetProtocol string
+	if _, err := getProtocol(service, annDOTargetProtocol, func(protocol string) { targetProtocol = protocol }); err != nil {
+		return err
+	}
+	if targetProtocol == protocolTCP || targetProtocol == protocolUDP {
+		return fmt.Errorf("target protocol annotation is not supported for TCP or UDP")
+	}
+
 	if tlsPassThrough {
 		forwardingRule.TlsPassthrough = tlsPassThrough
 		// We don't explicitly set the TargetProtocol here since in buildForwardingRule
@@ -936,7 +955,11 @@ func buildTLSForwardingRule(forwardingRule *godo.ForwardingRule, service *v1.Ser
 		// to match the EntryProtocol.
 	} else {
 		forwardingRule.CertificateID = certificateID
-		forwardingRule.TargetProtocol = protocolHTTP
+		if targetProtocol != "" {
+			forwardingRule.TargetProtocol = targetProtocol
+		} else {
+			forwardingRule.TargetProtocol = protocolHTTP
+		}
 	}
 
 	return nil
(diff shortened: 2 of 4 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -928,6 +939,14 @@ func buildTLSForwardingRule(forwardingRule *godo.ForwardingRule, service *v1.Ser
 		return errors.New("either certificate id should be set or tls pass through enabled, not both")
 	}
 
+	var targetProtocol string
+	if _, err := getProtocol(service, annDOTargetProtocol, func(protocol string) { targetProtocol = protocol }); err != nil {
```
timoreimann, line 943: Passing in an anonymous function seems a bit too complicated to me; and I'm still thinking about whether the new logic would do the right thing under all circumstances.

How about we design it like this (untested):

```golang
function getEntryProtocol(service *v1.Service) (string, error) {
  protocol, err := getProtocol(service, annDOProtocol, supportedEntryProtocols)
  if err != nil {
    return "", err
  }

  if protocol == "" {
    protocol = protoTCP
  }
  
  return protocol, nil
}

function getTargetProtocol(service *v1.Service) (string, error) {
  protocol, err := getProtocol(service, annDOTargetProtocol, supportedTargetProtocols)
  if err != nil {
    return "", err
  }

  if protocol == "" {
    protocol = protoHTTP
  }

  return protocol, nil
}

func getProtocol(service *v1.Service, annotation string, supportedProtocols []string) (string, error) {
    protocol, ok := service.Annotations[annotation]
    if !ok {
      return "", nil
    }

    if !slices.Contains(supportedProtocols, protocol) {
      return "", fmt.Errorf("invalid protocol %q specified in annotation %q", protocol, annotation)
    }

    return protocol, nil
}
```

WDYT?

CHANGELOG.md
```diff
@@ -1,10 +1,14 @@
...
+* When using the LoadBalancer `service.beta.kubernetes.io/do-loadbalancer-target-protocol` annotation with any of `http`, 
+  `https`, `http2` or `http3` specification, the forwarding target protocol is overridden from the implicit default `http`
+  protocol.
+
 ## v0.1.55 (beta) - July 29, 2024
 
 * When using the LoadBalancer `service.beta.kubernetes.io/do-loadbalancer-type=REGIONAL_NETWORK` (under closed beta), firewall rules
-are added to open up the underlying health check port and all the defined (port, protocols) defined on the service. This is to
-permit traffic to arrive directly on the underlying worker nodes.
+  are added to open up the underlying health check port and all the defined (port, protocols) defined on the service. This is to
```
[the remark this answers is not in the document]
asaha2: Nit whitespace adjustment really, I just noticed that the last version update didn't have a tab prepended to new lines stanza as other updates. I can revert this.

cloud-controller-manager/do/loadbalancers.go
asaha2: Addressed in 79364406117831963aa9814fd3fcc47b212f5476.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -733,13 +737,24 @@ func buildHTTP3ForwardingRule(ctx context.Context, service *v1.Service, godoClie
 		return nil, errors.New("certificate ID is required for HTTP3")
 	}
 
+	var targetProtocol string
+	if targetProtocol, err = getTargetProtocol(service); err != nil {
+		return nil, err
+	}
+	if targetProtocol == protocolTCP || targetProtocol == protocolUDP {
```
timoreimann, line 744: Do we need this explicit check still given that `getTargetProtocol()` now returns an error if a protocol other than from the allowed list (which does not include TCP or UDP) is specified?

Similar question for other occurrences of this check.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -733,13 +737,24 @@ func buildHTTP3ForwardingRule(ctx context.Context, service *v1.Service, godoClie
...
+		return nil, errors.New("target protocol annotation is not supported for TCP or UDP")
+	}
+	if targetProtocol == "" {
+		targetProtocol = protocolHTTP
```
timoreimann, line 748: It looks like we don't need this anymore either given that `getTargetProtocol()` already returns `protocolHTTP` if the annotation is missing.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -936,7 +962,11 @@ func buildTLSForwardingRule(forwardingRule *godo.ForwardingRule, service *v1.Ser
 		// to match the EntryProtocol.
 	} else {
 		forwardingRule.CertificateID = certificateID
-		forwardingRule.TargetProtocol = protocolHTTP
+		if targetProtocol != "" {
```
timoreimann, line 965: Similarly, I don't think we need the _if_ statement since `targetProtocol` is guaranteed to be populated by `getTargetProtocol()`.

CHANGELOG.md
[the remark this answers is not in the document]
timoreimann: No this is great. Thanks for cleaning up.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -80,6 +81,9 @@ const (
 
 var (
 	errLBNotFound = errors.New("loadbalancer not found")
+
+	supportedEntryProtocols  = []string{protocolTCP, protocolHTTP, protocolHTTPS, protocolHTTP2, protocolHTTP3}
+	supportedTargetProtocols = []string{protocolHTTP, protocolHTTPS, protocolHTTP2, protocolHTTP3}
```
timoreimann, line 86: To confirm, are TCP->TCP and UDP->UDP not supported?

trojanowski:
I think the approach from this PR, where the target protocol is set for all ports, could be problematic. Here's why:
* The default http → https redirection behavior (configured via the `service.beta.kubernetes.io/do-loadbalancer-redirect-http-to-https` annotation) may not suit all needs. For example, you might want to configure different rules, such as [returning an error instead of redirecting for API requests](https://jviide.iki.fi/http-redirects). With the solution implemented in this PR, we would need to configure TLS for the HTTP target port, which seems counterintuitive.
* That annotation would make exposing additional ports for protocols other than HTTP (e.g., SMTP/MQTT/AMQP) impossible.

My recommendation would be to implement an annotation like `service.beta.kubernetes.io/do-loadbalancer-target-https-ports` instead. This would explicitly define which ports should use the https protocol (or http2 if used by the load balancer) as the target protocol by providing a comma-separated list as a value.

d-honeybadger:
Closing for now since the work is incomplete and enough time has passed since opening to lose context.
Can definitely be revived later.

PR #776, opened 2024-11-18
migrate to instances v2

varshavaradarajan:


cloud-controller-manager/do/droplets.go
```diff
@@ -26,123 +26,39 @@ import (
...
-}
-
-// InstanceTypeByProviderID returns the type of the droplet identified by providerID.
-func (i *instances) InstanceTypeByProviderID(ctx context.Context, providerID string) (string, error) {
-	id, err := dropletIDFromProviderID(providerID)
-	if err != nil {
-		return "", err
-	}
+// cloudprovider.InstancesV2 methods
+// InstancesV2 require ProviderID to be present, so the interface methods all use providerID to get droplet.
```
varshavaradarajan, line 50: https://github.com/kubernetes/kubernetes/blob/021dbe9d195165cfcbe756c0b6dd5bb943b64a7c/staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go#L580

cloud-controller-manager/do/droplets.go
```diff
@@ -159,51 +75,52 @@ func (i *instances) InstanceExistsByProviderID(ctx context.Context, providerID s
...
+	}
+	return &cloudprovider.InstanceMetadata{
+		ProviderID:    node.Spec.ProviderID, // the providerID may or may not be present according to the interface doc. However, we set this from kubelet.
+		InstanceType:  droplet.SizeSlug,
+		Region:        droplet.Region.Slug,
+		NodeAddresses: nodeAddrs,
+	}, nil
+}
 
-	return nil, cloudprovider.InstanceNotFound
```
varshavaradarajan, line 206: Ah, I need to handle this in InstancesV2

varshavaradarajan: Handled, with a note that returning false with no error from InstanceExists will remove the node from the cluster.  We return nil error if godo err response is http.StatusNotFound.

varshavaradarajan:
Will merge this, cause dev image to be released, will run the iteration e2e tests with the dev image and release CCM is all goes well.

PR #796, opened 2025-01-17
fix comment on api results per page

d-honeybadger:
The comment became incorrect after adjusting the value in https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/795/files


m3co-code (approved):
Thank you!!

PR #802, opened 2025-02-14
CON-11819 spread sync of load balancers

llDrLove:
Spread the sync period of the loadbalancer unevenly to prevent a spike of loadbalancer API calls at the same time. We saw an issue when multiple clusters see their maintenance window happening at the same time and the CCM component is restarted at the same time. The sync tag period will match closely for all those CCM pods and all of those clusters will make a LIST api calls relatively close to each other resulting in spikes each 15minutes. This PR attempts to mitigate this issue by adding some randomness to the sync period (5minutes) as an initial delay. Thus, the new sync period will remain 15 minutes but the subsequent sync tag will be off between 0 to 300 seconds.

### DIFF
--- a/cloud-controller-manager/do/resources.go
+++ b/cloud-controller-manager/do/resources.go
@@ -95,7 +95,6 @@ func (s *tickerSyncer) Sync(name string, period time.Duration, stopCh <-chan str
 		case <-stopCh:
 			return
 		}
-		ticker.Reset(period + (time.Second * time.Duration(rand.Int31n(300))))
 	}
 }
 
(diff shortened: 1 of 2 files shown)


d-honeybadger (approved):

cloud-controller-manager/do/resources.go
```diff
@@ -85,6 +86,14 @@ func (s *tickerSyncer) Sync(name string, period time.Duration, stopCh <-chan str
 		klog.Errorf("%s failed: %s", name, err)
 	}
 
+	initialDelayTicker := time.NewTicker(initialDelay)
+	defer initialDelayTicker.Stop()
```
d-honeybadger, line 90: (nit) could use `time.After()` https://pkg.go.dev/time#After instead of ticker

cloud-controller-manager/do/resources.go
```diff
@@ -126,7 +135,7 @@ func (r *ResourcesController) Run(stopCh <-chan struct{}) {
 		klog.Info("No cluster ID configured -- skipping cluster dependent syncers.")
 		return
 	}
-	go r.syncer.Sync("tags syncer", controllerSyncTagsPeriod, stopCh, r.syncTags)
+	go r.syncer.Sync("tags syncer", controllerSyncTagsPeriod, time.Second*time.Duration(rand.Int31n(300)), stopCh, r.syncTags)
```
d-honeybadger, line 138: Now that it's initial delay, maybe let's increase the spread to ~10 min?

llDrLove: Bumped to 600 (10minutes)

cloud-controller-manager/do/resources.go
llDrLove: Using time.After() instead. Thanks

PR #827, opened 2025-05-19
Prepare release v0.1.60

okamidash:


go.mod
```diff
@@ -15,19 +15,19 @@ require (
...
 	golang.org/x/lint v0.0.0-20210508222113-6edffad5e616
 	golang.org/x/oauth2 v0.25.0
-	golang.org/x/sync v0.13.0
-	k8s.io/api v0.32.0
-	k8s.io/apimachinery v0.32.0
-	k8s.io/client-go v0.32.0
-	k8s.io/cloud-provider v0.32.0
-	k8s.io/component-base v0.32.0
+	golang.org/x/sync v0.14.0
+	k8s.io/api v0.32.5
```
d-honeybadger, line 19: I see the release doc doesn't explicitly say it, but we're supposed to update k8s packages to the next minor for CCM (just like all other DO-maintained component during the release time).
Normally `go get -u ./...` would actually make these updates cause it bumps all minor versions, but maybe it's gotten smarter and now understands that, for a package with `v0` major version, a minor version bump is actually a big change, and only does patch version bumps in these cases?
Anyway, if `go get -u ./...` didn't do it automatically, then you'll need to update them by hand. I think 0.33.1 is the latest one now

PR #829, opened 2025-05-22
CON-12382 backfill lb type annotation in k8s LB service object

llDrLove:
Backfill the LB type annotation in k8s LB services object. 

Existing and new LBs will always have the two following annotations: 

```yaml
  metadata:
    annotations:
      kubernetes.digitalocean.com/load-balancer-id: <cluster-uuid>
      service.beta.kubernetes.io/do-loadbalancer-type: <type>
```

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -226,6 +226,7 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
 		logLBInfo("CREATE", lbRequest, 2)
 
 		updateServiceAnnotation(service, annDOLoadBalancerID, lb.ID)
+		updateServiceAnnotation(service, annDOType, lb.Type)
 
 	default:
 		// unrecoverable LB retrieval error
@@ -428,6 +429,7 @@ func (l *loadBalancers) retrieveAndAnnotateLoadBalancer(ctx context.Context, ser
 	}
 
 	updateServiceAnnotation(service, annDOLoadBalancerID, lb.ID)
+	updateServiceAnnotation(service, annDOType, lb.Type)
 
 	return lb, nil
 }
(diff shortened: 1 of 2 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -226,6 +226,7 @@ func (l *loadBalancers) EnsureLoadBalancer(ctx context.Context, clusterName stri
 		logLBInfo("CREATE", lbRequest, 2)
 
 		updateServiceAnnotation(service, annDOLoadBalancerID, lb.ID)
+		updateServiceAnnotation(service, annDOType, lb.Type)
```
gottwald, line 229: Why is this needed? Shouldn't retrieveAndAnnotate take care of this already?
Especially since it's after the create?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -428,6 +429,7 @@ func (l *loadBalancers) retrieveAndAnnotateLoadBalancer(ctx context.Context, ser
 	}
 
 	updateServiceAnnotation(service, annDOLoadBalancerID, lb.ID)
+	updateServiceAnnotation(service, annDOType, lb.Type)
```
gottwald, line 432: We retrieve and update the object here, but where do we actually update the manifest in k8s api?

cloud-controller-manager/do/loadbalancers.go
llDrLove: No `lb, err = l.retrieveAndAnnotateLoadBalancer(ctx, service)` on line 210 will annotate only if the LB already exists. In the case the `err` is `errLBNotFound` meaning that the LB does not exists yet, we issue a Create API call. At this point (line 221), the LB is not yet annotated with DO id or DO lb type. Only on line 228 and 229 will it modify and add the two annotations.

cloud-controller-manager/do/loadbalancers.go
llDrLove: I wasn't seeing where it was called but it is because the loadbalancer struct we have here is implementing the `EnsureLoadBalancer` method which is defined upstream. We implement the `LoadBalancer` upstream interface (See [here](https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L138-L173)) which I guess does the annotation update. I didn't tested on a real cluster yet because of the inci. The PR is still draft...

gottwald (approved):

PR #830, opened 2025-05-23
CON-12387 Ensure NLB is the default loadbalancer in DO CCM

llDrLove:
Change the default LB type to be NLB instead of regional LB.


varshavaradarajan (approved):
A minor version release on CCM too for this, right?

PR #839, opened 2025-06-12
Restrict the kube-proxy health port to be accessible by the load balancer only instead of allowing all IPv4 and IPv6 addresses

llDrLove:
* Add optional `DEFAULT_LB_TYPE` environment variable to configure the default load balancer type between `REGIONAL` and `REGIONAL_NETWORK` 
* Restrict the kube-proxy health port to be accessible by the load balancer only instead of allowing all IPv4 and IPv6 addresses

### DIFF
--- a/cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/main.go
+++ b/cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/main.go
@@ -86,9 +86,6 @@ func startAdmissionServer() error {
 	vpcID := os.Getenv(doClusterVPCIDEnv)
 	lbAdmissionHandler.WithVPCID(vpcID)
 
-	// Default LB Type
-	lbAdmissionHandler.WithDefaultLBType(defaultLBType)
-
 	ll.Info("registering admission handlers")
 	server.Register("/lb-service", &webhook.Admission{Handler: lbAdmissionHandler})
 
--- a/cloud-controller-manager/do/lb_service_admission_handler.go
+++ b/cloud-controller-manager/do/lb_service_admission_handler.go
@@ -183,8 +183,3 @@ func (a *LBServiceAdmissionHandler) WithVPCID(vpcID string) {
 func (a *LBServiceAdmissionHandler) WithClusterID(clusterID string) {
 	a.clusterID = clusterID
 }
-
-// WithDefaultLBType sets the defaultLBType field of the handler.
-func (a *LBServiceAdmissionHandler) WithDefaultLBType(defaultLBType string) {
-	a.defaultLBType = defaultLBType
-}


cloud-controller-manager/do/firewall_controller.go
```diff
@@ -329,12 +330,22 @@ func (fm *firewallManager) createReconciledFirewallRequest(serviceList []*v1.Ser
...
+				if hcp == 0 {
 					continue
 				}
-				loadBalancerPorts[portProtocol{protocol: "tcp", port: healthCheckPort}] = struct{}{}
-
+				pp := portProtocol{
+					protocol: "tcp",
+					port:     hcp,
+				}
+				if hcp == kubeProxyHealthPort {
```
varshavaradarajan, line 341: I'm sorry for missing some context, if the health check port is something other than kube-proxy, why are we okay allowing all ipv4 and ipv6 for other health check ports defined?

cloud-controller-manager/do/firewall_controller_diff.go
```diff
@@ -115,8 +115,11 @@ func compFirewallsEqual(cf1, cf2 *comparableFirewall) (bool, string) {
 	// Ignore all fields on {In,Out}boundRules.{Sources,Destinations} other than
 	// "Addresses".
```
d-honeybadger, line 116: (nit) the comment needs to change to 
```
// Ignore all fields on {In,Out}boundRules.{Sources,Destinations} other than
// "Addresses" and "LoadBalancerUIDs"
```
right?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -80,8 +81,15 @@ const (
 
 var (
 	errLBNotFound = errors.New("loadbalancer not found")
+	defaultLBType = godo.LoadBalancerTypeRegionalNetwork
 )
 
+func init() {
```
d-honeybadger, line 87: Could this initialization be done inside the `newCloud` here https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/9d3e276756b74642ae201abf693384cf90907329/cloud-controller-manager/do/cloud.go#L86 and then the default value passed to the `newLoadBalancers` struct? To avoid globals and limit the number of `init` functions and have all startup configuration in one place

llDrLove: Sure, I moved it to the `newCloud()` function.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -80,6 +80,7 @@ const (
 
 var (
 	errLBNotFound = errors.New("loadbalancer not found")
+	defaultLBType = godo.LoadBalancerTypeRegionalNetwork
```
d-honeybadger, line 83: Can it be a part of the loadBalancers struct https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/2f22f6d4bfc2fb01ed8d8d26f1c2575fd1a88612/cloud-controller-manager/do/loadbalancers.go#L90 rather than a global?

llDrLove: Yeah, so I thought of that initially but it is not only used within the `loadbalancers` struct scope. Thus, not easily convertible to a method. We use the `getType()` in other places that do not have the `loadbalancer` struct at hand. I'll need to add it as a property on each of those structs too. Also, since they are already declared as env vars, they kind of are already global values. That variable is just there to prevent other parts of the code to always fetch from env values (which they could do) which should be slower.

Issue #844, opened 2025-06-30
Removing the certificate ID annotation causes a segfault


okamidash:
Tested on my DO cluster.
Steps to reproduce:

1. Create a cluster
2. Create a service and expose it via type `LoadBalancer`
3. Add a certificate ID to the LoadBalancer
4. Observe that the certificate ID is added to the LoadBalancer
5. Remove it
6. ????
7. Profit

```
E0630 13:12:53.541237       1 panic.go:262] "Observed a panic" panic="runtime error: invalid memory address or nil pointer dereference" panicGoValue="\"invalid memory address or nil pointer dereference\"" stacktrace=<
	goroutine 285 [running]:
	k8s.io/apimachinery/pkg/util/runtime.logPanic({0x2d13d50, 0xc00069adb0}, {0x2486bc0, 0x4465e90})
		/go/src/github.com/digitalocean/digitalocean-cloud-controller-manager/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:132 +0xbc
	k8s.io/apimachinery/pkg/util/runtime.handleCrash({0x2d13d88, 0xc0003ea820}, {0x2486bc0, 0x4465e90}, {0x0, 0x0, 0xc0006f96a8?})
		/go/src/github.com/digitalocean/digitalocean-cloud-controller-manager/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:107 +0x116
	k8s.io/apimachinery/pkg/util/runtime.HandleCrashWithContext({0x2d13d88, 0xc0003ea820}, {0x0, 0x0, 0x0})
		/go/src/github.com/digitalocean/digitalocean-cloud-controller-manager/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:78 +0x5a
	panic({0x2486bc0?, 0x4465e90?})
		/usr/local/go/src/runtime/panic.go:792 +0x132
	github.com/digitalocean/digitalocean-cloud-controller-manager/cloud-controller-manager/do.(*loadBalancers).recordUpdatedLetsEncryptCert(0xc000420280, {0x2d13d88, 0xc0003ea820}, 0xc000ceea08, {0xc000cb4d50, 0x24}, {0x0, 0x0})
		/go/src/github.com/digitalocean/digitalocean-cloud-controller-manager/cloud-controller-manager/do/loadbalancers.go:317 +0x2d4
[...]

PR #845, opened 2025-06-30
Handle removing Certificate Annotation

okamidash:
Fixes https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/844

When a customer removes a certificate ID from their annotations, it crashes. This adds logic in to check if the serviceCertID (derived from the annotation) even has anything populated before running further checks.

Our API erroneously returns a 200 OK when you pass `""` as an argument to `Certificates.Get` which is why this didn't get caught.

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -284,7 +284,7 @@ func getCertificateIDFromLB(lb *godo.LoadBalancer) string {
 // annotation on the Service gets newly-updated certificate ID from the
 // Load Balancer.
 func (l *loadBalancers) recordUpdatedLetsEncryptCert(ctx context.Context, service *v1.Service, lbCertID, serviceCertID string) error {
-	if lbCertID != "" && lbCertID != serviceCertID {
+	if lbCertID != "" && serviceCertID != "" && lbCertID != serviceCertID {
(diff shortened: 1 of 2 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -284,7 +284,7 @@ func getCertificateIDFromLB(lb *godo.LoadBalancer) string {
 // annotation on the Service gets newly-updated certificate ID from the
 // Load Balancer.
 func (l *loadBalancers) recordUpdatedLetsEncryptCert(ctx context.Context, service *v1.Service, lbCertID, serviceCertID string) error {
-	if lbCertID != "" && lbCertID != serviceCertID {
+	if lbCertID != "" && serviceCertID != "" && lbCertID != serviceCertID {
```
d-honeybadger, line 287: If both cert name and cert ID annotation need to be missing for the cert to get removed, I think the PR needs a tiny tweak, so that
- if both cert ID annotation and cert name annotation are empty, we leave those annotations empty and bail out from  recordUpdatedLetsEncryptCert
- if only cert ID annotation is empty, but cert name is set, then we update cert ID annotation to match the actual LB cert ID

Also, a code comment about 200 responses with empty cert ID would be helpful

more context in https://digitalocean.slack.com/archives/C4HREK3T6/p1756914553227899?thread_ts=1751279243.181349&cid=C4HREK3T6

PR #847, opened 2025-06-30
LBAAS-3627: add load balancer network-stack annotation to CCM

jvasilevsky:
add network-stack annotation to allow configuration of IPV4 or DUALSTACK networking

At this time REGIONAL_NETWORK LB only supports IPV4 network stack


StephenVarela (approved):

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -1453,6 +1458,26 @@ func getNetwork(service *v1.Service) (string, error) {
 	return network, nil
 }
 
+func getNetworkStack(service *v1.Service, lbType string, network string) (string, error) {
+	networkStack := service.Annotations[annDONetworkStack]
+	if networkStack == "" {
+		if network == godo.LoadBalancerNetworkTypeInternal || lbType == godo.LoadBalancerTypeRegionalNetwork {
```
m3co-code, line 1464: nit: I wonder if we should turn the default around, so that everything that isn't caught by a check, falls back to the original value. To avoid accidents.

m3co-code (changes requested):
left minor comments on docs, otherwise lgtm. please ping once fixed then I approve

m3co-code (approved):
lgtm!

cloud-controller-manager/do/lb_annotations.go
```diff
@@ -178,11 +178,16 @@ const (
...
 
-	// annDONetwork is the annotation used to specify the network type of the load balancer. Either EXTERNAL or INTERNAL (currently in closed alpha)
+	// annDONetwork is the annotation used to specify the network type of the load balancer. Either EXTERNAL or INTERNAL
 	// are permitted. If no network is provided, then it will default EXTERNAL.
 	annDONetwork = annDOLoadBalancerBase + "network"
+
+	// annDONetworkStack is the annotation used to specify the network stack of the load balancer. Either IPV4 or DUALSTACK
+	// are permitted. If load balancer type is REGIONAL then it will default to DUALSTACK. If load balancer type
+	// is REGIONAL_NETWORK then it will default to IPV4.
+	annDONetworkStack = annDOLoadBalancerBase + "network-stack"
```
gottwald, line 192: Do we have to make clear here that `REGIONAL_NETWORK` ipv6 is not supported, yet?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -1453,6 +1458,26 @@ func getNetwork(service *v1.Service) (string, error) {
...
+	if !(networkStack == godo.LoadBalancerNetworkStackIPv4 || networkStack == godo.LoadBalancerNetworkStackDualstack) {
+		return "", fmt.Errorf("only LB network stacks supported are (%s, %s)", godo.LoadBalancerNetworkStackIPv4, godo.LoadBalancerNetworkStackDualstack)
+	}
+	if lbType == godo.LoadBalancerTypeRegionalNetwork && networkStack == godo.LoadBalancerNetworkStackDualstack {
+		return "", fmt.Errorf("dual stack networking is not supported for regional network LB with kubernetes at this time")
+	}
+	if network == godo.LoadBalancerNetworkTypeInternal && networkStack == godo.LoadBalancerNetworkStackDualstack {
+		return "", fmt.Errorf("dual stack networking is not supported for internal load balancer")
+	}
+	return networkStack, nil
```
gottwald, line 1478: I feel like the possible combinations are getting quite tricky and easy to mess up for the user.
Would these errors get bubbled up as events to the user on the svc object?

cloud-controller-manager/do/lb_annotations.go
gottwald: Nevermind, the annotations readme should be enough.

PR #898, opened 2026-02-03
Add public ipv6 node address support

gottwald:
When used with droplets that have public ipv6 addresses, these will be added to the node metadata.

### DIFF
--- a/cloud-controller-manager/do/droplets_test.go
+++ b/cloud-controller-manager/do/droplets_test.go
@@ -177,6 +177,65 @@ func newFakeShutdownDroplet() *godo.Droplet {
 	}
 }
 
+func newFakeDropletWithIPv6() *godo.Droplet {
+	droplet := newFakeDroplet()
+	droplet.Networks.V6 = []godo.NetworkV6{
+		{
+			IPAddress: "2604:a880:800:10::1",
+			Type:      "public",
+			Gateway:   "2604:a880:800:10::1",
+			Netmask:   64,
+		},
+	}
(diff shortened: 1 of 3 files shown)


m3co-code (approved):
LGTM

cloud-controller-manager/do/droplets_test.go
```diff
@@ -291,106 +349,291 @@ func TestInstanceShutdown(t *testing.T) {
...
+		{
+			name:       "without provider ID - IPv4 and IPv6",
+			providerID: "",
+			fakeDroplet: &fakeDropletService{
+				listByNameFunc: func(ctx context.Context, name string, opt *godo.ListOptions) ([]godo.Droplet, *godo.Response, error) {
+					return []godo.Droplet{*newFakeDropletWithIPv6()}, newFakeOKResponse(), nil
+				},
+			},
+			expectedMeta: &cloudprovider.InstanceMetadata{
+				ProviderID:   "digitalocean://123",
```
varshavaradarajan, line 494: Seems like an incorrect test case without provider id?

varshavaradarajan (changes requested):

cloud-controller-manager/do/droplets_test.go
gottwald: I think this is correct. We don't pass a providerID (a few lines up), which causes a name-based lookup. But we still expect to get an ID from the looked up droplet.

varshavaradarajan (approved):

PR #902, opened 2026-02-12
Add IPv6 regional network LB support

gottwald:
This adds support for IPV6 network LBs.
These can only be used if all cluster nodes are fully dual stack enabled.

### DIFF
--- a/cloud-controller-manager/do/loadbalancers.go
+++ b/cloud-controller-manager/do/loadbalancers.go
@@ -531,8 +531,8 @@ const (
 
 // nodeState contains the classification and filtering results for nodes
 type nodeState struct {
-	// readyNodes are nodes that are ready to serve as LB backends
-	readyNodes         []*v1.Node
+	// lbReadyNodes are nodes that are ready to serve as LB backends
+	lbReadyNodes       []*v1.Node
 	filteredCount      int
 	singleStackV4Nodes []*v1.Node
 	singleStackV6Nodes []*v1.Node
@@ -578,7 +578,7 @@ func (ns *nodeState) isAllDualStack() bool {
 // and singleStackV6 nodes are filtered out (not supported for external connectivity).
 func filterAndClassifyNodes(nodes []*v1.Node, isInternalLB bool) *nodeState {
 	state := &nodeState{
-		readyNodes:         make([]*v1.Node, 0, len(nodes)),
+		lbReadyNodes:       make([]*v1.Node, 0, len(nodes)),
 		singleStackV4Nodes: make([]*v1.Node, 0),
 		singleStackV6Nodes: make([]*v1.Node, 0),
 		dualStackNodes:     make([]*v1.Node, 0),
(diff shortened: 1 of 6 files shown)


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -511,6 +519,184 @@ func updateServiceAnnotation(service *v1.Service, annotName, annotValue string)
...
+	nodeClassSingleStackV4    nodeClassification = "singleStackV4"    // Has IPv4, no IPv6
+	nodeClassSingleStackV6    nodeClassification = "singleStackV6"    // Has IPv6, no IPv4
+	nodeClassDualStack        nodeClassification = "dualStack"        // Has both IPv4 and IPv6
+	nodeClassPublicNetUnready nodeClassification = "publicNetUnready" // No external IPs, (yet)
+)
+
+// nodeState contains the classification and filtering results for nodes
+type nodeState struct {
+	// readyNodes are nodes that are ready to serve as LB backends
+	readyNodes         []*v1.Node
```
varshavaradarajan: nit: rename since this could be assumed on node readiness.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -1458,23 +1715,92 @@ func getNetwork(service *v1.Service) (string, error) {
...
+		// REGIONAL_NETWORK with explicit dual-stack annotation requires validation
+		// (only when we have node state - admission path skips this)
+		if hasNodeState && lbType == godo.LoadBalancerTypeRegionalNetwork {
+			if nodeState.hasSingleStackV4 {
+				if explicitAnnotation {
+					// User explicitly requested dual-stack but has singleStackV4 nodes - PERMANENT ERROR
+					return "", fmt.Errorf("%w: dual-stack networking requested but %d nodes lack IPv6 addresses: %s. REGIONAL_NETWORK load balancers require all nodes to have IPv6 for dual-stack mode",
+						ErrNetworkStackConfig, len(nodeState.singleStackV4Nodes), formatNodeNames(nodeState.singleStackV4Nodes, 3))
+				}
+				// Should not reach here due to default logic above, but handle defensively
```
varshavaradarajan, line 1800: Why not return an error? 

If not, we have to patch the explicit annotation to be ipv4, do we do that? There's a `retrieveAndAnnotateLoadBalancer` that's used, perhaps add for the network stack as well?

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -511,6 +520,176 @@ func updateServiceAnnotation(service *v1.Service, annotName, annotValue string)
...
+		if isIPv6(addr.Address) {
+			hasIPv6 = true
+		} else {
+			hasIPv4 = true
+		}
+	}
+
+	// Determine classification based on IP presence
+	switch {
+	case hasIPv4 && hasIPv6:
```
varshavaradarajan, line 645: The if-else part above means this condition won't pass?

varshavaradarajan (changes requested):

cloud-controller-manager/do/loadbalancers.go
gottwald: Why? I don't see it.
This condition is, that we have two external addresses, one v4 and one v6. Means the loop above had at least two iterations as there's no early break in it.

cloud-controller-manager/do/loadbalancers.go
gottwald: Yeah, I think erroring out is fine here. Will change it.

cloud-controller-manager/do/loadbalancers.go
varshavaradarajan: Ah sorry.

varshavaradarajan (approved):

PR #912, opened 2026-04-20
Replace alpine:3.15 with distroless static

thearyanahmed:
This PR Replaces `alpine:3.15` with `gcr.io/distroless/static-debian13:nonroot` for both images.

Also removes package installation (`apk add ca-certificates`) and simplify the image contents to just the compiled binaries.

### DIFF
--- a/cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/Dockerfile
+++ b/cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/Dockerfile
@@ -12,10 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.15
+FROM gcr.io/distroless/static-debian12:nonroot
 
-RUN apk add --no-cache ca-certificates
+COPY digitalocean-cloud-controller-manager-admission-server /bin/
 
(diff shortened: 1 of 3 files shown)


m3co-code (approved):
LGTM but I think we should verify this works as expected before merging.

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager-admission-server/Dockerfile
```diff
@@ -12,10 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.15
+FROM gcr.io/distroless/static-debian12:nonroot
```
gottwald: Can we make this debian 13?

cloud-controller-manager/cmd/digitalocean-cloud-controller-manager/Dockerfile
```diff
@@ -12,10 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.15
+FROM gcr.io/distroless/static-debian12:nonroot
```
gottwald: debian 13 here as well

thearyanahmed:
@gottwald updated to debian 13.

gottwald (approved):

PR #915, opened 2026-06-01
firewall_controller: honor firewall-managed annotation and loadBalancerSourceRanges for LoadBalancer services

### LINKED ISSUE #911: firewall_controller: no way to restrict inbound source CIDRs for LoadBalancer services on DOKS (hardcoded to 0.0.0.0/0)
## Problem

When using a DOKS cluster with a `LoadBalancer` type Service, the CCM automatically
manages a `k8s-public-access-*` firewall and adds inbound rules for the service ports
(e.g. 80, 443) with source hardcoded to `0.0.0.0/0` / `::/0`.

There is no way to restrict these rules to private or specific CIDRs only.
Any manual changes to the firewall are reverted by the controller on the next
reconcile.

## Expected behavior

Allow specifying allowed source CIDRs for the auto-managed firewall inbound rules,
for example via a Service annotation or a CCM environment variable — similar to how
`spec.loadBalancerSourceRanges` works for the Load Balancer itself.

## Use case

Clusters behind Cloudflare (or any CDN/proxy) should only accept traffic from
the CDN IP ranges — not from the entire internet. Right now this is only possible
at the Load Balancer level (`spec.loadBalancerSourceRanges`), but the node-level
firewall stays wide open regardless.

While digging into the source, found a related inconsistency worth noting:

The annotation `kubernetes.digitalocean.com/firewall-managed: "false"` already exists and **works for `NodePort` services** — when set, the controller skips adding that service's ports to the firewall ([firewall_controller.go ~L286](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/cloud-controller-manager/do/firewall_controller.go#L286)):

```go
// NodePort branch — annotation IS checked
if !managed {
    continue // port not added to k8s-public-access
}
```

But for `LoadBalancer` services, `isManaged()` is **never called** — the ports are always added with `0.0.0.0/0` regardless of the annotation ([same file, ~L323](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/cloud-controller-manager/do/firewall_controller.go#L323)):

```go
} else if svc.Spec.Type == v1.ServiceTypeLoadBalancer {
    // isManaged() is not called here — annotation is ignored
    loadBalancerPorts[portProtocol{...}] = struct{}{}
}
```

A minimal fix would be to extend the existing `isManaged()` check to the `LoadBalancer` branch as well. Alternatively, for source CIDR restriction, the controller could read `spec.loadBalancerSourceRanges` (already parsed in `loadbalancers.go` via `getSourceRangeRules()`) and use those as inbound rule sources instead of hardcoded `0.0.0.0/0`.

elkh510:
Fixes #911

### 1. Extend `isManaged()` check to LoadBalancer services

The `kubernetes.digitalocean.com/firewall-managed: "false"` annotation was silently ignored for `LoadBalancer` services - `isManaged()` was only called in the `NodePort` branch. This change adds the same check to the `LoadBalancer` branch so the annotation works consistently across both service types.

### 2. Use `spec.loadBalancerSourceRanges` as inbound rule sources

For `REGIONAL_NETWORK` + external `LoadBalancer` services, service port inbound rules were always created with hardcoded `0.0.0.0/0` / `::/0` sources. This change reads `spec.loadBalancerSourceRanges` (already parsed elsewhere for LB-level firewall rules) and uses those CIDRs as the source addresses in the worker firewall inbound rules when set.

Health check port rules are unaffected - they continue to use `LoadBalancerUIDs` as source.

Internal (`network: internal`) LBs are unaffected - they are excluded by the existing `lbNetwork == External` guard and no worker firewall rules are added for them.

## Tests

Two new test cases added to `TestFirewallController_createReconciledFirewallRequest`:
- `skip REGIONAL_NETWORK LB service when firewall-managed annotation is false`
- `use loadBalancerSourceRanges as inbound rule sources for REGIONAL_NETWORK LB`

### DIFF
--- a/cloud-controller-manager/do/firewall_controller.go
+++ b/cloud-controller-manager/do/firewall_controller.go
@@ -321,6 +321,14 @@ func (fm *firewallManager) createReconciledFirewallRequest(serviceList []*v1.Ser
 				)
 			}
 		} else if svc.Spec.Type == v1.ServiceTypeLoadBalancer {
+			managed, err := isManaged(svc)
+			if err != nil {
+				klog.Warningf("managing service %s/%s for which no correct management flag setting could be detected: %s", svc.Namespace, svc.Name, err)
+				managed = true
+			}
+			if !managed {
+				continue
+			}
 			lbType, err := getType(svc, fm.defaultLBType)
 			if err != nil {
 				return nil, fmt.Errorf("failed to get load balancer type for service %s/%s: %v", svc.Namespace, svc.Name, err)
@@ -350,6 +358,12 @@ func (fm *firewallManager) createReconciledFirewallRequest(serviceList []*v1.Ser
 
 				loadBalancerPorts[pp] = struct{}{}
 				// Add the services (port, protocol)
+				var svcPortSources *godo.Sources
+				if len(svc.Spec.LoadBalancerSourceRanges) > 0 {
+					svcPortSources = &godo.Sources{
+						Addresses: svc.Spec.LoadBalancerSourceRanges,
+					}
+				}
 				var protocol string
 				for _, servicePort := range svc.Spec.Ports {
 					switch servicePort.Protocol {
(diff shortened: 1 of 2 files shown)


elkh510:
hi @gottwald @sreeram-venkitesh @sriritwikdo — could you please take a look when you have a moment? This fixes the two issues described in #911.

PR #918, opened 2026-06-24
Add handling of private IP only droplets

gottwald:
This adds capability of handling private-IP-only droplets.

### DIFF
--- a/cloud-controller-manager/do/droplets.go
+++ b/cloud-controller-manager/do/droplets.go
@@ -110,11 +110,14 @@ func dropletByName(ctx context.Context, client *godo.Client, nodeName types.Node
 		if droplet.Name == string(nodeName) {
 			return &droplet, nil
 		}
-		addresses, _ := nodeAddresses(&droplet)
-		for _, address := range addresses {
-			if address.Address == string(nodeName) {
-				return &droplet, nil
-			}
+		if privateIP, err := droplet.PrivateIPv4(); err == nil && privateIP == string(nodeName) {
+			return &droplet, nil
+		}
+		if publicIPv4, err := droplet.PublicIPv4(); err == nil && publicIPv4 == string(nodeName) {
+			return &droplet, nil
+		}
+		if publicIPv6, err := droplet.PublicIPv6(); err == nil && publicIPv6 == string(nodeName) {
+			return &droplet, nil
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


cloud-controller-manager/do/droplets.go
```diff
@@ -110,11 +110,14 @@ func dropletByName(ctx context.Context, client *godo.Client, nodeName types.Node
 		if droplet.Name == string(nodeName) {
 			return &droplet, nil
 		}
-		addresses, _ := nodeAddresses(&droplet)
-		for _, address := range addresses {
-			if address.Address == string(nodeName) {
-				return &droplet, nil
-			}
+		if privateIP, err := droplet.PrivateIPv4(); err == nil && privateIP == string(nodeName) {
```
varshavaradarajan, line 113: privateIP won't be equal to nodeName?

varshavaradarajan (changes requested):

cloud-controller-manager/do/droplets.go
gottwald: Apparently it can. Or I'm not too sure tbh.
The old code that is removed above assumed that would be the case. 
I've never seen it tbh.
But there's also a good chance this code was never reached and the name match takes care of everything.

gottwald: It's also documented: https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/fdca6fd9ea0913c61d57529c73a3e2fcb8a13024/docs/getting-started.md#L62C183-L62C287
It's for sure not used in DOKS, but maybe out in the wild.

cloud-controller-manager/do/loadbalancers.go
```diff
@@ -570,13 +576,23 @@ func (ns *nodeState) isAllDualStack() bool {
...
 // singleStackV4, singleStackV6, or dualStack.
 //
-// For INTERNAL load balancers, public IPs not required, private assumed.
-// For EXTERNAL load balancers, have at least one external IP,
+// Private-only nodes (no ExternalIP) are allowed when allowsPrivateOnlyBackends returns true.
+// For EXTERNAL REGIONAL_NETWORK load balancers, nodes must have at least one external IP,
 // and singleStackV6 nodes are filtered out (not supported for external connectivity).
-func filterAndClassifyNodes(nodes []*v1.Node, isInternalLB bool) *nodeState {
+func filterAndClassifyNodes(nodes []*v1.Node, lbType, lbNetwork string) *nodeState {
+	allowPrivateOnly := allowsPrivateOnlyBackends(lbType, lbNetwork)
```
varshavaradarajan, line 593: nit: rename, this could be interpreted as only allow private droplets.

varshavaradarajan (approved):

cloud-controller-manager/do/loadbalancers.go
varshavaradarajan: can't think of a better name.. :)

PR #937, opened 2026-07-28
added subnet uuid and lb byoip support

rbhatia-code:


cloud-controller-manager/do/loadbalancers.go
```diff
@@ -990,6 +993,7 @@ func (l *loadBalancers) buildLoadBalancerRequest(ctx context.Context, service *v
 
 	req.Region = l.region
 	req.VPCUUID = l.resources.clusterVPCID
+	req.VPCSubnetUUID = getSubnetUUID(service)
```
varshavaradarajan: This should be in buildLoadBalancerRequest helper function where others that are to be got from annotations are: https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/cloud-controller-manager/do/loadbalancers.go#L878

varshavaradarajan (changes requested):

varshavaradarajan (approved):

